PR #1, opened 2019-04-03
Add fip-0001 and initialize the repo

whyrusleeping:
This is largely just ethereums process. It's a good one that has been adapted from bitcoins, which was adapted from pythons, and has served the community pretty well over the years.

I don't want to bikeshed things *too* much here, but there are a few points that we should figure out before proceeding (and several TODOs i wrote into the docs that need to be addressed).

Note that this process will not be actually used until the protocol and initial implementations are complete.

PR #4, opened 2020-09-11
Security consideration

momack2:


momack2:
Added based on @jsoares comments here: https://github.com/filecoin-project/FIPs/pull/1

PR #8, opened 2020-09-24
FIP Template Update

zixuanzh:
- Rename `Motivation` to `Problem Motivation` and `Rationale` to `Design Rationale` to reflect the intent of the section.
- Add `Incentive Considerations` and `Product Considerations` as they are core to Filecoin's success as a network of reliable and useful storage.


templates/template_FTP.md
```diff
@@ -26,15 +26,15 @@ The title should be 44 characters or less.
 <!--A short (~200 word) description of the technical issue being addressed.-->
 A short (~200 word) description of the technical issue being addressed.
 
-## Motivation
+## Problem Motivation
```
jsoares, line 29: Semantic point: one motivates an action/solution, not a problem. You could call it "Problem **and** motivation" but seems rather redundant.

momack2: Mitigated by renaming to "Change Motivation"

momack2:
Discussed in the 2020-10-08 Core Devs Weekly - no comments/concerns.

PR #20, opened 2020-10-23
No repay debt requirement for DeclareFaultsRecovered

irenegia:
Remove the requirement to repay debt from DeclareFaultsRecovered.


FIPS/fip-000x.md
```diff
@@ -0,0 +1,60 @@
...
+## Test Cases
+N/A
+
+## Implementation
+Remove the following line of code in the miner actor
+https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1322.
+
+
+## Security Considerations
+After this change, a miner in fee debt continues to not be able to produce block and withdraw tokens from its available balance, so the possible actions and strategies for malicious miners are the same as in the current version of the protocol. In particular, a miner that does not recover from debt (ie, a miner that does not pay its fees) can make profit from the power registered in the power table because it is not eligible to mine blocks.
```
momack2, line 49:
```suggestion
After this change, a miner in fee debt continues to not be able to produce block and withdraw tokens from its available balance, so the possible actions and strategies for malicious miners are the same as in the current version of the protocol. In particular, a miner that does not recover from debt (ie, a miner that does not pay its fees) cannot make profit from the power registered in the power table because it is not eligible to mine blocks. 
```

zixuanzh (approved):

zixuanzh:
There is a potential product concern here which is that the protocol does want miners to recover as soon as possible. However, I still think this FIP is fine because the penalty stops when the miner has recovered the data and hence there is still guarantee to the storage client even though the miner may not be able to produce blocks until debt is paid. cc @pooja

momack2:
I don't think your logic lines up here, @zixuanzh. The protocol wants the miner to recover their sectors as soon as possible - however blocking the miner on declaring the recovery due to a debt actually makes that much harder - increasing the chances the miner ends up in a downward spiral that causes them to lose all deals/sectors. Landing this FIP makes it easier for miners to recover their sectors and repay a more manageable debt (which they are still highly incentivized to do) - increasing storage performance overall.

FIPS/fip-000x.md
```diff
@@ -0,0 +1,60 @@
...
+In other words, no check about the FeeDebt variable is required in the method “DeclareFaultsRecovered”.
+Diff from current version: remove the following line of code in the miner actor
+https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1322.
+
+
+## Design Rationale
+The main motivation for this change is introducing the minimal change that allows miners in debt to recover storage faults in order to stop paying for penalties.
+
+## Backwards Compatibility
+ This change requires a actor upgrade.
```
anorth, line 38: This change requires a network version increment to distinguish the old behaviour from the new. It does not change any state schemas or exported method signatures.

anorth (approved):

FIPS/fip-000x.md
momack2:
```suggestion
This change requires a network version increment to distinguish the old behaviour from the new. It does not change any state schemas or exported method signatures.
```

FIPS/fip-000x.md
```diff
@@ -0,0 +1,60 @@
...
+N/A
+
+## Implementation
+Remove the following line of code in the miner actor
+https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/miner/miner_actor.go#L1322.
+
+
+## Security Considerations
+After this change, a miner in fee debt continues to not be able to produce block and withdraw tokens from its available balance, so the possible actions and strategies for malicious miners are the same as in the current version of the protocol. In particular, a miner that does not recover from debt (ie, a miner that does not pay its fees) can make profit from the power registered in the power table because it is not eligible to mine blocks. 
+Miners that recover their faults are still not eligible until they recover from fee debt, this means that their power will still count towards the total power, and the average WinCount is reduced. Although this is not a new problem, this change could worsen the potential power inflation. We do not believe this is a risk for the protocol and this problem should be addressed separately with a new FIP (e.g. allowing miners in debt to miner or remove the power from the power table).
```
momack2, line 50: @irenegia - can you clarify the "e.g." statement?

> allowing miners in debt to **miner or** remove the power from the power table

irenegia: @momack2
the "e.g." statement contains two examples of possible FIP that we may want to consider to resolve the power table inflation problem.  
if you think the mentioning theme here is confusing, it it totally fine to remove the statement.

momack2: do you mean the two options are - 
1. allowing miners in debt to mine
2. remove the power from the power table?

The sentence as currently written doesn't parse (I can't understand what it means)

irenegia: let me know if this version is more clear
https://github.com/filecoin-project/FIPs/pull/20/commits/63e596d216f41a5f506838001a9dea84dc5ab2ff

momack2: yep! thanks!

Issue #25, opened 2020-11-02
Add a batched PreCommitSectors method


anorth:
The miner `PreCommitSector` method only supports committing a single sector at a time. It's one of the two highest frequency methods observed on the chain at present (the other being ProveCommitSector). High-growth miners commit sectors at rates exceeding 1 per epoch. It's also a relatively expensive method, with multiple internal sends and loading and storing state including:
- the Storage Power actor's power total (read)
- the Storage Market actor's deal proposal AMT (read)
- the Reward actor's totals (read)
- the `AllocatedSectors` bitfield (read and modify)
- the `PrecommittedSectors` HAMT (read and modify)
- the `Sectors` AMT (read)
- the `PreCommittedSectorsExpiry` AMT (read and modify)
- the Storage Power actor's pledge total (read and modify)

A `PreCommitSectorsBatch` method has potential to amortize some of these costs across multiple sectors. If miner operators implemented a relatively short batch aggregation period (a few epochs), the number of invocations could be reduced significantly, and some of the state manipulations above reduced in proportion.

anorth:
`ProveCommitSector` could also be batched, but that's a separate discussion. There is already a complex set-up around batching the proof verifications until the end of each epoch, and that discussion should start there.

nicola:
I approve the spirit of this change.

One thing to note is that in the future (may not belong to this FIP), a PreCommitBatched could contain PreCommits from other miners and the batched method would be verifying the signature on those PreCommits (similarly to what we do for deals). 

In this way, smaller miners could get together to jointly submit PreCommits.

zixuanzh:
I mentioned during all dev that batched precommits may give miners with high sealing throughput an advantage over miners who are smaller in scale. It is unclear if the cost saving is worth the tradeoff. On second thought, the number of sectors that we are batching is probably small enough that it wouldn't make much of a difference. Can we quantify the difference? Of course allowing different miners' PreCommits to batch will be great but that's also extra work and hurdle that presents challenges to smaller miners.

nicola:
This FIP ensures that a miner can organize their PreCommits so that they can amortize their gas costs. The gas saving is pretty substantial. For example if you submit 32 sectors instead of 1, you may end up paying ~32x less gas (likely less than that).

This is a good concern, but since this is a "one-off" cost, I dont think this will have major impact in the total cost of mining of a small miner vs larger miner.

Extending this FIP in the future to support PreCommit from multiple miners will only benefit the miners.

[...]

PR #27, opened 2020-11-03
First draft for batch sector pre-commit

anorth:
See #25.

I'm open to discussion from implementers and operators about the relative value of supporting per-sector success/failure, rather than the proposed whole-batch success/failure.

FYI @ZenGround0 @momack2 @arajasek @magik6k @nicola

### DIFF
--- a/FIPS/fip-xxxx-batchprecommit.md
+++ b/FIPS/fip-xxxx-batchprecommit.md
@@ -1,0 +1,122 @@
+---
+fip: <to be assigned>
+title: Add miner batched sector pre-commit method
+author: Alex North (@anorth)
+discussions-to: https://github.com/filecoin-project/FIPs/issues/25
+status: Draft
+type: Technical
+category: Core
+created: 2020-11-04
+spec-sections: 
+  - section-systems.filecoin_mining.sector.lifecycle
+---
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+Add a method for a miner to submit sector pre-commitments in batches.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+The miner `PreCommitSector` method only supports committing a single sector at a time.
+It is both frequently executed and relatively expensive.
+This proposal adds a `PreCommitSectorBatch` method to amortize some of the costs across multiple sectors,
+and removes some redundant but costly checks.
+
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+The miner `PreCommitSector` method only supports committing a single sector at a time. 
+It's one of the two highest frequency methods observed on the chain at present (the other being `ProveCommitSector`). 
+High-growth miners commit sectors at rates exceeding 1 per epoch. 
+It's also a relatively expensive method, with multiple internal sends and loading and storing state including:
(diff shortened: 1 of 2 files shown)


FIPS/fip-xxxx-batchprecommit.md
```diff
@@ -0,0 +1,122 @@
...
+- The `PreCommittedSectorsExpiry` AMT queue is loaded and stored just once for the batch.
+
+The redundant state checks are removed from the non-batch `PreCommitSector` method too.
+
+### Failure handling
+- If any predicate on the parameters fails, the call aborts (no change is persisted).
+- If the miner has insufficient balance for all pre-commit deposits, the call aborts.
+
+### Scale and limits
+The number of sectors that may be pre-committed in a single batch is limited to 32.
```
nicola, line 72: Why 32? It could be much more than this.

E.g. 1000

FIPS/fip-xxxx-batchprecommit.md
```diff
@@ -0,0 +1,122 @@
...
+- If the miner has insufficient balance for all pre-commit deposits, the call aborts.
+
+### Scale and limits
+The number of sectors that may be pre-committed in a single batch is limited to 32. 
+
+
+## Design Rationale
+The existing `PreCommitSector` method will become redundant, since a batch of one will not be significantly less efficient.
+The method is left intact to support smooth operation through the upgrade period.
+Support for the `PreCommitSector` method may be dropped in a future FIP.
```
nicola, line 78: should this FIP set a time for which that method becomes invalid?

FIPS/fip-xxxx-batchprecommit.md
ZenGround0: Explained [below](https://github.com/filecoin-project/FIPs/pull/27/files#diff-9ebbdcedea827e8122d1f8037e9b3f3bc463e42c7e15ed091c65fb50cc19a2abR89)

FIPS/fip-xxxx-batchprecommit.md
ZenGround0: I consider this out of scope for this FIP since deprecation is not a key part of enabling batching.  However I am interested to hear arguments for doing it in this FIP.

FIPS/fip-xxxx-batchprecommit.md
nicola: I understand the explanation, but I don't understand why this limit is in place. I still think that this should enable much higher precommits e.g. 1000.

anorth: As a matter of policy, we must specify a limit (even though there will be an implicit limit from the chain message size limit too). A bias to conservatism suggests erring low (and this limit is not very low). I think the burden of argument lies with your proposal for a higher one. What benefit would that bring?

If F is the fixed cost that can be amortized across a batch, a batch size of 32 has a maximum benefit of reducing the ongoing to 3.1% of F. So we get ~97% of the possible benefit already. The marginal value of increasing the batch size is very small and vanishes quickly.

nicola: We can surely start with 32.

I understand that it will get most of the gains in respect to today's situation - however, in the future since precommitting is getting cheaper, the throughput of precommit might increase and then we may need to raise the number again.

Since we can go higher, at least as I understand it right now, I don't see why we should not.

momack2:
@anorth is this draft ready to be merged or does it need additional iteration?

anorth:
I need two review approvals. It's ready to be merged IMO, while the FIP status remains draft pending discussion at an implementers call.

momack2 (approved):

jennijuju (approved):

jennijuju:
Approved and assigned FIP number.

PR #44, opened 2020-12-11
Update FIP-0005 to Final

arajasek:
This is now live on the Filecoin network.


momack2 (approved):

jennijuju:
Is this `Final` b/c it is implemented, merged and running in the network? If so, the [implementation](https://github.com/filecoin-project/FIPs/blob/master/templates/template_FTP.md#implementation) section should be added with the link to the pr imo.

jennijuju (approved):

PR #46, opened 2020-12-11
Update FIP-0006 to Deferred

arajasek:
As discussed in https://github.com/filecoin-project/tpm/issues/15

### DIFF
--- a/FIPS/fip-0006.md
+++ b/FIPS/fip-0006.md
@@ -3,7 +3,7 @@ fip: 0006
 title: No repay debt requirement for DeclareFaultsRecovered
 author:  Nicola (@nicola), Irene (@irenegia)
 discussions-to: https://github.com/filecoin-project/FIPs/issues/21
-status: Draft
+status: Deferred
 type: Technical
 created: 2020-10-23
 ---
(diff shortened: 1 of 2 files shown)


momack2:
This would be our first "deferred". Is the aim of those that we ARE going to do it but just in a future upgrade, or just "not now"? IIRC, there were some research questions to land in the FIP that needed additional work - so Draft may be a better designation to indicate there's open work before approval...?

jennijuju:
Could we please update the status in the [FIP table](https://github.com/filecoin-project/FIPs#fips) with status changing prs too?

arajasek:
@momack2 Based on what we've said, "Deferred" means "a FIP that is not being considered for immediate adoption. May be reconsidered in the future for a subsequent chain upgrade." 

That sounds like what we agreed to, except that we likely will not be reconsidering in a future upgrade (and instead be opening a new, better FIP that achieves a similar purpose). Perhaps it's worth adding a more explicit "Rejected" state?

momack2:
Sg, let’s move to “deferred” for now and we can separately consider the
recent changes to the EIP process we may want to adopt re “withdrawn”
proposals:
https://eips.ethereum.org/EIPS/eip-1#eip-process

jennijuju (approved):

momack2 (approved):

Issue #42, opened 2020-12-09
Off-chain Window PoSt verification


anorth:
## Problem
Window PoSt messages are necessary for ongoing maintenance of storage power. Verifying the submitted proofs is expensive and when the gas base fee rises (due to congestion) these messages become expensive. In bad cases, for small miners with mostly-empty partitions, this cost can exceed their expected reward from maintaining power. We need to ensure that these messages are cheap for miners, even when specifying a very high gas fee cap.

Note the framing of the problem here _assumes_ congestion. There is a separate and just-as-pressing problem of reducing congestion. But even if we make great improvements there, congestion will happen sometimes in the future for different reasons (E.g. massive popularity! Deals! DeFi!) and we'll need Window PoSt to remain cheap and effective.

This is an alternative to the [Fast Track for Window PoSt](https://github.com/filecoin-project/FIPs/issues/24) and related proposals.

## Proposed solution
Don't verify most Window PoSt proofs "on chain". Assume they are valid, and implement a challenge mechanism for external parties to dispute an invalid proof, verifying on-chain only at that point.

### Outline
- Change the SubmitWindowPoSt method to store the proof bytes (or a hash) in per-deadline chain state instead of verifying the proof. The method still checks that the proof is expected, records skipped faults, etc. Optimistically update state assuming the proof is valid, including marking recovered sectors and gaining power for new/recovered sectors.
- During end-of-deadline cron, snapshot the partition state relevant for proof verification (the active/faulty sector sets and proofs themselves)
- The proofs and snapshots remains in chain state until the next occurrence of the deadline (~24h)
- At any point in that challenge window, a challenger can force the on-chain validation of a proof. If the proof fails validation, all of the sectors in the partitions included in that proof are marked faulty, penalised, and the miner loses power for them. 

With this mechanism, the vast majority of Window PoSt proofs would never be validated on chain. The network cost of maintaining storage would be effectively constant, rather than linear in storage as it is today (at sufficient scale it would revert to linear for the work of submitting the proofs, but over that timeframe expect to develop other aggregation techniques).

## Discussion
This issue is intended for discussion. There are a number of details to work out before drafting a FIP.

### Impact
[...]

Issue #47, opened 2020-12-11
Sealing Transaction Fee Rate


jbenet:
## Problem

Chain congestion due to sector sealing messages -- `PreCommit` and `ProveCommit` -- is driving up the BaseFee and making WindowPosts, Deals, and other messages very expensive. Other proposals (#24, [#24-b](https://github.com/filecoin-project/FIPs/issues/24#issuecomment-718339203), #42) propose ways to make WindowPost cheaper or get a dedicated control gas plane, but most of those solutions will take significant time to implement.

It seems that Miners for now have settled on a BaseFee expense between 1-5 nFIL (gas expenditure dominated by `PreCommit` and `ProveCommit`.

## Temporary Solution

The other solutions discussed are the right thing to do. But in the meantime:

We might alleviate the pain by introducing a gas fee _rate increase_ for `PreCommit` and `ProveCommit` messages. This means multiply the gas fee by a `SealingFeeRate` -- a factor, say `10x` or even `20x`. This will cause parties that are sealing massive amounts currently (in this difficult congestion period) to arrive at a BaseFee about 10x cheaper. 

**Scales with BaseFee**. because this is a multiplication on top of the gas costs of those messages, it will scale with the BaseFee, and will cause the network to find the miners' price point much faster.

**Burn fees pass to miners sealing**. This is unlikely to _reduce_ burn fees, only move the proportion paid by most users (sp WindowPosts) to be paid by miners growing their storage significantly. 

**Downside: harms sealing new storage**. This harms _all_ sealing, not just CC sectors. This will hurt new sectors with storage, as they will be more expensive to seal. This may be an acceptable price to pay until we get a better solution in.

**Network Policy Lever**. This creates a network policy lever that trades off between cheaper capacity onboarding or cheaper everything else.

## Implementation Notes

This should be very easy to implement. this is why I'm proposing it. This should be a relatively straight forward change to introduce in gas accounting.

## Other Solutions

#24, [#24-b](https://github.com/filecoin-project/FIPs/issues/24#issuecomment-718339203), #42 and more

PatrickDeuse:
> It seems that Miners for now have settled on a BaseFee expense between 1-5 nFIL (gas expenditure dominated by PreCommit and ProveCommit.

this does not mean that miners are paying that because they want to or can afford it. most miners are paying the 3-6 (that's what we are actually at...) nFIL because they have to. if they don't - they will end up with a power share that equals basically 0 rewards. 

[...]

Issue #24, opened 2020-10-28
Discussion about "Fast-track for WindowPoSt"


irenegia:
### Problem
A WindowPoSt message is a “unique” kind of message because it has an expiration (i.e., it has to go on chain before a given epoch in order to be valid and effective). Because of this, we need to guarantee that enough windowPoSt messages are selected and included by a block producer, even when there is message congestion.

### Proposed Solution
Allocate dedicated block space for WindowPoSt messages.

#### Specification
We introduce a message classification system:
- Type A messages: WindowPoSt submission; 
- Type B messages: any message that is not in type A;

Each message type now has its own gas pricing mechanism. Each mechanism is the same as in the Dynamic Network Fee mechanism, but they are independent. More precisely: 
We introduce variables: BlockGasLimit_A, BlockGasTarget_A, BlockGasLimit_B, BlockGasTarget_B
- BlockGasLimit_A = TypeAShare*BlockGasLimit, 
- BlockGasTarget_A = TypeAShare*BlockGasTarget,
  - TypeAShare is a number between 0 and 1 , to be decided
- BlockGasLimit_B = TypeBShare*BlockGasLimit, 
- BlockGasTarget_B = TypeBShare*BlockGasTarget,
  - TypeBShare = 1 -TypeAShare

Each block now has two base fees: baseFee_A and baseFee_B. Both base fees are computed using the formula “NewBaseFee = OldBaseFee + delta*OldBaseFee”, where delta depends on the difference between the actual gas limit of the block (i.e.,  the sum of each included message’s gas limit) and the BlockGasTarget. The differences among the two are
In the computation of the delta for baseFee_B (resp. baseFee_A) we use BlockGasTarget_B/2  (resp.  BlockGasTarget_A/2) and consider msg.GasLimit taken from messages of type B (resp. Type A) only;
Currently the delta is a value in the range [-⅛, ⅛], we can keep this range for both or use two different ranges. 

###  Discussion
One problem @Kubuxu pointed out for this proposal is that if we reserve too little for WindowPoSt then the window post partition will run up the corresponding base fee very high, on the other side reserving too much decreases the chain throughput and creates problem for the other messages (especially precommit/proveCommit).

###  Other Solutions
- Discounted base fee for windowPost message (to be analyzed/discussed)

anorth:
[...]

PR #59, opened 2021-01-08
Off-Chain Window PoSt

Stebalien:
See #42. WIP.

### DIFF
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -270,6 +270,7 @@ However:
 1. If these posts are successfully disputed, the attacker will lose 5.51 times the expected block reward per-day for the incorrectly proven sectors.
 2. Restoring faulty sectors requires an on-chain proof. This prevents attackers from extending faulty sectors indefinitely by repeatedly submitting faulty proofs.
 
+5.51 was chosen as it gives negative profit for miners as long as invalid proofs are caught within 12 hours with probability 1/3.
 
 ### Vesting Fund Extraction
 
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -1,7 +1,7 @@
 ---
 fip: 0010
 title: Off-Chain Window PoSt Verification
-author: Alex (@anorth), Steven (@stebalien), et al.
+author: Steven (@stebalien), Alex (@anorth), et al.
 discussions-to: https://github.com/filecoin-project/FIPs/issues/42
 status: Draft
 type: Technical
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -213,6 +213,11 @@ Of course, there needs to be at least one party disputing proofs for this to wor
 Currently, there is one primary reason to dispute an invalid proof: to ensure the healthy operation of the network.
 Organizations invested in the continued operation of the network and the reliability of it's storage have a vested interest in ensuring that all data purported to be stored on the network is being correctly proven.
 
+Specifically:
+
+1. Checking proofs is relatively cheap compared with running a Filecoin miner. At the moment, _all_ Filecoin nodes are verifying _all_ proofs.
+2. If a significant number of invalid proofs go undisputed, the confidence in the Filecoin network would be effected, impacting the value of the token. Given that all large miners have large amounts of vesting Filecoin rewards, they have a strong incentive to ensure this doesn't happen.
+
 Disputing a proof also yields a minimal reward. Unfortunately, given the expected rarity of invalid proofs, no reward will fully cover the cost of actually checking all these proofs. However, the reward will at least cover the cost of submitting the dispute, and provide a little extra as an incentive.
 
 ### Including Proof Dispute Messages Is Rational
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -164,7 +164,7 @@ Proofs may not be disputed after this time has passed.
    disproven). Otherwise, we continue.
 3. All sectors that were active in the partition snapshot are declared as faulty (sectors terminated
    since the last challenge window are skipped).
-4. The target miner looses power based on the sectors marked faulty in step 3.
+4. The target miner loses power based on the sectors marked faulty in step 3.
 5. The disputed proof is removed from `Deadline.OptimisticPoStSubmissionsSnapshot` so it can't be
    disputed twice.
 6. The target miner is fined IPF+DisputeReward based on the power was incorrectly proven (taken from
@@ -288,7 +288,7 @@ See:
 
 ## Backwards Compatibility
 
-As this is a consensus breaking change, all Filecoin implementer must implement the changes to the
+As this is a consensus breaking change, all Filecoin implementers must implement the changes to the
 Filecoin actors. Specifically, the Miner actor has changed as described in the "specification"
 section.
 
@@ -409,7 +409,7 @@ Of course, there needs to be at least one party disputing proofs for this to wor
 
 Currently, there is one primary reason to dispute an invalid proof: to ensure the healthy operation
 of the network.  Organizations invested in the continued operation of the network and the
-reliability of it's storage have a vested interest in ensuring that all data purported to be stored
+reliability of its storage have a vested interest in ensuring that all data purported to be stored
 on the network is being correctly proven.
 
 Specifically:
@@ -417,7 +417,7 @@ Specifically:
 1. Checking proofs is relatively cheap compared with running a Filecoin miner. At the moment, _all_
    Filecoin nodes are verifying _all_ proofs.
 2. If a significant number of invalid proofs go undisputed, the confidence in the Filecoin network
-   would be effected, impacting the value of the token. Given that all large miners have large
+   would be affected, impacting the value of the token. Given that all large miners have large
    amounts of vesting Filecoin rewards, they have a strong incentive to ensure this doesn't happen.
 
 Disputing a proof also yields a minimal reward. Unfortunately, given the expected rarity of invalid
@@ -428,7 +428,7 @@ incentive.
 ### Including Proof Dispute Messages Is Rational
 
 Proof dispute messages directly reduce the power of other miners on the network, proportionally
-increasing the relative power of all other miners in the network. Therefore, baring collusion, it is
+increasing the relative power of all other miners in the network. Therefore, barring collusion, it is
 always rational to include dispute messages.
 
 Furthermore, proof dispute messages give a reward. Assuming the base fee isn't so high that the
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -204,6 +204,9 @@ In this section, we discuss the three actions being incentivised:
 
 ### Submitting Invalid Proofs Is Irrational
 
+As long as one participant in the network is reliably checking and disputing invalid proofs, submitting an invalid proof will incur a higher penalty than simply not submitting a proof, or declaring the sector faulty.
+
+Of course, there needs to be at least one party disputing proofs for this to work.
 
 ### Disputing Proofs Is Rational
 
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -32,12 +32,13 @@ The proposed change will, remove nearly all of the burden of checking `SubmitWin
 We propose two key changes:
 
 1. New Window PoSts, except those that restore faulty sectors, are optimistically accepted, assumed correct, and recorded in the state-tree for one proving period.
-2. Optimistically accepted window posts can be disputed until `WPoStProofDisputeWindow` epochs after the challenge window in which they were submitted closes. When a dispute successfully refutes an optimistically accepted Window PoSt, the miner is fined one IPF per active sector in the partition (at the moment when the proof was submitted) and all incorrectly proved sectors are marked faulty.
+2. Optimistically accepted window posts can be disputed until `WPoStProofDisputeWindow` epochs after the challenge window in which they were submitted closes. When a dispute successfully refutes an optimistically accepted Window PoSt, the miner is fined one IPF per active sector in the partition (at the moment when the proof was submitted),  all incorrectly proved sectors are marked faulty, and the disputer (address that submitted the dispute) is rewarded a fixed `DipsuteReward`.
 
 Parameters:
 
 * `WPoStProofDisputeWindow`: 1800 (2x finality).
 * `IPF`: 5.51BR (IPF = Invalid Proof Fee, BR = Expected Block Reward per sector in 24h).
+* `DisputeReward`: To be decided but it's likely to be in the range of 1-2FIL. The reward must at least cover the expected gas.
 
 In the following sections, we describe the changes to the Miner actor necessary to implement this FIP.
 
@@ -128,7 +129,9 @@ Proofs may not be disputed after this time has passed.
 3. All sectors that were active in the partition snapshot are declared as faulty (sectors terminated since the last challenge window are skipped).
 4. The target miner looses power based on the sectors marked faulty in step 3.
 5. The disputed proof is removed from `Deadline.OptimisticPoStSubmissionsSnapshot` so it can't be disputed twice.
-6. The target miner is fined IPF based on the power was incorrectly proven (taken from the partitions snapshot). This fine will also cover sectors that have since been terminated.
+6. The target miner is fined IPF+DisputeReward based on the power was incorrectly proven (taken from the partitions snapshot). This fine will also cover sectors that have since been terminated. This fine is taken from the miner's "vesting rewards" in the following order.
+    1. The IPF portion of the fine (variable depending on the amount of incorrectly proven power) is burnt. If insufficient funds are available in the vesting rewards, the remainder is added to the miner's fee debt to be burnt if and when the fee debt is repaid.
+	2. The `DisputeReward` portion of the fine (fixed) is sent to the address that submitted the dispute. If insufficient funds are available in the vesting rewards, the remainder of the reward is added to the miner's fee debt _and not_ sent to the address that submitted the dispute.
 
 ## Design Rationale
 <!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
@@ -204,18 +207,10 @@ In this section, we discuss the three actions being incentivised:
 
 ### Disputing Proofs Is Rational
 
-Currently, there are two reasons to dispute an invalid proof:
+Currently, there is one primary reason to dispute an invalid proof: to ensure the healthy operation of the network.
+Organizations invested in the continued operation of the network and the reliability of it's storage have a vested interest in ensuring that all data purported to be stored on the network is being correctly proven.
 
-1. To maximize profits as a miner by minimizing the power of other miners.
-2. To ensure the healthy operation of the network. Organizations invested in the continued operation of the network and the reliability of it's storage have a vested interest in ensuring that all data purported to be stored on the network is being correctly proven.
-
-Disputing a proof does not currently give a reward. There are three potential sources of funds:
-
-1. Unlocked/spendable funds.
-2. Collateral.
-3. Vesting rewards.
-
-Any rational miner submitting invalid proofs will drain their unlocked funds down to 0. That means funds would need to come from either collateral or vesting rewards. Unfortunately, taking from either would affect circulating supply and could be used to extract "locked" funds.
+Disputing a proof also yields a minimal reward. Unfortunately, given the expected rarity of invalid proofs, no reward will fully cover the cost of actually checking all these proofs. However, the reward will at least cover the cost of submitting the dispute, and provide a little extra as an incentive.
 
 ### Including Proof Dispute Messages Is Rational
 
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -431,6 +431,10 @@ Proof dispute messages directly reduce the power of other miners on the network,
 increasing the relative power of all other miners in the network. Therefore, baring collusion, it is
 always rational to include dispute messages.
 
+Furthermore, proof dispute messages give a reward. Assuming the base fee isn't so high that the
+reward doesn't cover the gas fees, a rational miner will front-run proof dispute messages to collect
+the reward.
+
 ## Product Considerations
 
 This FIP will increase chain bandwidth by moving Window PoSt processing off-chain. This will reduce
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -14,31 +14,50 @@ spec-sections:
 
 ## Simple Summary
 
-Optimistically accept Window PoSt proofs on-chain without verification, allowing them to be disputed later by off-chain verifiers.
+Optimistically accept Window PoSt proofs on-chain without verification, allowing them to be disputed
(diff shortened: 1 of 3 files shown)


TippyFlitsUK:
This would be extremely welcome.

cwhiggins:
I would say that I am not for this proposal.  I prefer what @jbenet proposed with a control plane and data plane,https://github.com/filecoin-project/FIPs/issues/24#issuecomment-743137490, I feel that is more in line with the type of community we should promote.  It was compared to making sure emergency service vehicles are able to get where they are going, people move out of the way not because of what they can gain. 
 In this FIP, the POST mechanism is reduced to this possiblity,   ### Disputing Proofs Is Rational
Currently, there are two reasons to dispute an invalid proof:
1. To maximize profits as a miner by minimizing the power of other miners. 

 and I don't think that this possiblity is appropriate for our communities goals.

Stebalien:
@cwhiggins this proposal is complementary to the "fast lane" proposals, it doesn't conflict with them. I completely agree that "necessary" messages should be prioritized.

I'm fielding this proposal because:

1. Unlike the fast-lane proposals, this proposal actually increases chain throughput. Even _with_ a fast-lane for window posts, the network will eventually grow to a size where window posts take _all_ chain bandwidth. That would break the network.
2. All the fast lane proposals I've seen (there are several) are significantly more difficult to implement than this proposal. This proposal requires some changes to the actors logic and state. Those proposals will require complex incentive analysis, changes to message selection, and likely changes to the block header format itself.

That's why I'm proposing _this_ solution _now_.

>> To maximize profits as a miner by minimizing the power of other miners.

> and I don't think that this possiblity is appropriate for our communities goals.

The _goal_ is to ensure that miners are correctly storing data, that's it. I'm providing both an altruistic and a selfish reason:

1. Altruistically, some parties will dispute incorrect window posts to ensure that data storage on Filecoin is reliable. By reliably disputing these proofs, miners are _highly_ incentivized to correctly store and prove their storage or risk penalties.
2. Selfishly, some parties will dispute incorrect window posts to maximize their own relative power. Power in a blockchain is a (mostly) zero sum game. The less power you have, the more power I have.

NSC-FIL:
I'm for this enhancement if the it doesn't slow down the sealing process.

cwhiggins:
Will this have to happen two days in a row before a miner is penalized? How often in a day can it be requested a miner prove their storage? Will it scale with size of miner, say only one request per day for 10Tib miner, 1000 for 10Pib?

cwhiggins:
Also, the end points of your response 1,2 you use the term "incorrect window posts", the requestor does not know the state of the Post. It is a fishing expedition.  The more resources you have, the more you can go fish.  Sorry, I don't like this implementation.

Stebalien:
> Will this have to happen two days in a row before a miner is penalized?

No. A miner will still be able to "miss" a single window post without paying penalties, but they won't be able to submit a _bad_ proof.

> How often in a day can it be requested a miner prove their storage? Will it scale with size of miner, say only one request per day for 10Tib miner, 1000 for 10Pib?

Miners will still need to prove their storage as normal using the normal proof mechanisms. The only change here is that the network won't validate these proofs unless some other miner submits a _new_ message disputing the proof.

Stebalien:
> Also, the end points of your response 1,2 you use the term "incorrect window posts", the requestor does not know the state of the Post. It is a fishing expedition. The more resources you have, the more you can go fish. Sorry, I don't like this implementation.

I think you're misunderstanding the proposal. This proposal requires _no_ additional work from miners proving their storage.

cwhiggins:
Could please clarify what will happen when some one submits a dispute to a proof?Will the miner at that point just be required to submit a saved message from the Post?

Stebalien:
> Could please clarify what will happen when some one submits a dispute to a proof?Will the miner at that point just be required to submit a saved message from the Post?

This process is explained in detail in the "Change: SubmitWindowedPoSt" and "New: DisputeWindowedPoSt" sections.

Simple summary (but you should still actually read the proposal):

1. When a proof is "optimistically" accepted (but not verified immediately) it is recorded in the state tree.
2. Later, another miner can "dispute" this proof by submitting the index in the state tree where the proof is recorded on chain.

cwhiggins:
so only other miners can dispute proofs, as they are then doing the work to verify against chain? It seems to me there still needs to be some sort of engagement with the miner in question at the time of the dispute?  I have read the proposal, obvious by my first post, really, had to go there?

Stebalien:
>  It seems to me there still needs to be some sort of engagement with the miner in question at the time of the dispute?

No, the proof is recorded in the on-chain state when the miner first submits it. The disputer just tells the chain to check the proof.

Steps:

1. The miner submits a proof. <-- only miner interaction
2. **New:** The chain stores it temporarily in the miner's state.
3. **New:** Someone disputes it (calls `DisputeWindowedPoSt`).
4. The network verifies the proof.

The only difference is the addition of steps 2 and 3 to make step 4 (the expensive part) optional unless someone notices that the proof is incorrect.

funktafide:
I think it looks like a god idea, but the possibility of large miners spamming bad proofs is somewhat concerning.  Obviously you don't want to penalize someone for 1 bad proof, but I think once the number reaches a specific point, the penalty should be stiff.

BrunoZell:
Does the network power only decrease when the dispute lands on chain? What happens in the intermediate time the miner had unlawful power to his name? Is it just the slashing penalty or are possible block rewards he got from that power deducted as well? (It seems difficult to get right though as it's usually just a fraction of his power that got lost). Also, will the process of continuously validating proofs and filing disputes be automated when running a node?

In general I am all for doing as much work off chain as possible. However, I feel it is very weird that miners need to pay each other in order to get paid. Especially considering that WindowPoST is very much expected and also required. As soon as a sealed sector is known the system should make room for the expected PoSTs, but not for more. Maybe a variable block size or a fixed fee set at sector commit?

Stebalien:
@funktafide

> Obviously you don't want to penalize someone for 1 bad proof...

That's actually not obvious to me and I'm curious what you're thinking here. At the moment, there are pretty stiff penalties (~5.1 days worth of block rewards) for submitting _any_ bad proofs (note, these aren't just missed proofs, they're actually incorrect). The idea is to check these proofs thoroughly locally before submitting them.

@BrunoZell

Great questions! I'll try to work my answers into the proposal.

> Does the network power only decrease when the dispute lands on chain?

Unfortunately, yes. Actually, a bit longer because Filecoin determines the winning miner by looking Finality (900) epochs back in the power table.

> What happens in the intermediate time the miner had unlawful power to his name?

They will continue to win blocks as normal.

> Is it just the slashing penalty or are possible block rewards he got from that power deducted as well?

The penalty scales with the amount of incorrectly proven power which is also proportional to the expected number of blocks won.

Specifically, this proposal currently sets the penalty to 5.1 "BR". That's the _expected_ block reward earned by the incorrectly proven sectors over 5.1 days.

So in theory, a miner could "get lucky" and win more than it loses. But a miner is significantly more likely to lose more than it wins.

> Also, will the process of continuously validating proofs and filing disputes be automated when running a node?

The network is secure as long as _someone_ checks the proofs. I expect the Filecoin Foundation to sponsor one, at least, given that the cost of running one such node is reasonably low.

For now, any "disputer" process will likely be a special node but I'd personally love to see a world where all lotus nodes randomly sample and check proofs at some user configurable frequency.

> In general I am all for doing as much work off chain as possible. However, I feel it is very weird that miners need to pay each other in order to get paid. As soon as a sealed sector is known the system should make room for the expected PoSTs, but not for more. Maybe a variable block size or a fixed fee set at sector commit?

I think I'm missing some context here but I'll answer as best I can. Specifically, could you expand on "miners need to pay each other in order to get paid" and "maybe a variable block size or a fixed fee set at sector commit"?

There are a number of reasons to verify posts off-chain, but I they all boil down to: it frees up chain bandwidth. Basically, we don't _need_ to synchronously and deterministically verify all proofs in lockstep, so why do it? We can let nodes verify proofs in whatever manner works for them, and report issues to the chain, This enables:

* Buffering. It's ok to fall behind a bit.
* Parallelism. It's possible to perform some operations in parallel on-chain, but it's tricky to get right and optimize.
* Caching. Caching in a deterministic environment is tricky and error prone so Lotus does _some_ caching but not nearly as much as it could. However, an off-chain proof verification system can cache the inputs to the proof verification logic and avoid re-loading them from the state tree.

funktafide:
@Stebalien I think I confused a couple different sections when reading.  I was looking at the Risks section.

A large miner could spam bad proofs to try to overwhelm the chain with the subsequent verification cost.
      We'd have to pick penalties/limits to avoid this.
      Each partition can only be used once in this way before a valid proof is needed to restore its power.
      A the moment, we have chain bandwidth to handle challenging every proof, and if we greatly reduce congestion, 
      headroom here will increase by >= 10x
      Future proof aggregation techniques may support submitting a challenge for multiple proofs with sub-linear verification 
      cost.

I was referring to a large miner spamming bad proofs from that section.  So if someone disputes a proof 1 time and its actually a valid proof, does the miner that disputes it get penalized? I was thinking the first time you wouldn't, but if you are spamming to overwhelm the chain, the penalty would get increasingly higher.  It looks like you are already looking at this, so my post was a bit redundant.  This was the part that kind of stuck out to me as a bit concerning since there are some pretty large miners out there.

jennijuju:
Can one proof being challenged multiple times within 1800 epochs by the same challenger or different challengers?

BrunoZell:
@funktafide The poential of spam is not limited to this - all transactions can be used to spam the network. The attacker would be paying the fees according to the computational load that is required to process the transaction in form of gas usage and simultaneously would drive up the base fee as intended by EIP1559.

@Stebalien 

> miners need to pay each other in order to get paid

...in a senes that they need to pay transaction fees for their sector commits and window PoSTs in order to mine blocks, and when they get elected to mine a block they are rewarded with the block reward plus the fees payed by all other miners. Specifically I dislike that gaining power in the network is coupled to the transaction fee market. The tx fee market is mainly dictated by the capacity and energy consumption of the blockchain.

IMO it is essential to open blockchain systems that everyone can gain power in the network regardless of congestion and fee market. In fact, I belive congestion should be an additional incentive for brand new miners to get these sweet transaction fees. Having new miners first pay those high fees to get power results in the opposite effect or even a death spral where existing miners will stop adding new storage due to falling profits.

> Maybe a variable block size or a fixed fee set at sector commit?

So how could we solve that? Two ideas that come to my mind.

1) Whenever a new sector is sealed and committed, the network knows for how long the sector is expected to stay alive. This publicly verifiable information can be used to make room in the chain for all expected window PoSTs, possibly by having miners always include those transactions regardless of block size. This would decouple those messages from the gas market and miners would essentially give each other network capacity. I see two problems with this approach: What's the incentive for miners to follow this protocol instead of leaving out messages and claim they didn't knew about them? Especially considering they would gain power in not including such PoSTs. And could such a mechanic block future developments of making file modifications possible or adding other new features that do not make the amount of window PoSTs predicateble?

2) Somehow fix the fee for all window PoST at commit time. After thinking more about this it really doesn't make any sense so I won't elaborate any furhter here 😄

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+In this section, we discuss the three actions being incentivised:
+
+1. Submitting invalid proofs must not be rational.
+2. Disputing invalid proofs must be rational.
+3. Including proofs dispute messages must be rational.
+
+### Submitting Invalid Proofs Is Irrational
+
+
+### Disputing Proofs Is Rational
```
irenegia, line 203: I am not fully convinced that "disputing proofs is rational":  the arguments used here to support that statement are true but what about miners try do be "free riders"?
That is, a miner can choose to let the others do the verification and disputing work (zero work for him, same "rewards" as before).
cc @sa8

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+* Three fields are being added to the end of the `Deadline` object (described in the specification section). State inspection tools will need to accommodate this.
+* The parameters for `SubmitWindowedPoSt` have not changed and the logic for generating and submitting proofs remains the same.
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
+Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.
+
+TODO
+
+## Security Considerations
```
irenegia, line 157: The security considerations section misses this points:
- Short fork attack may be possible and some invalid proof may not be disputed.
If the challenge for an invalid proof arrives close (ie, < 200 epochs) to the end of the challenge window, the miner (the one who submitted the invalid proof) could fork using selfish mining and create a chain where there is no challenges for his invalid proof.
- About "Power & sector extension attacks": Note that after this FIP the power for invalid proofs is removed only when the invalid proof is detected and therefor, from the deadline to this moment, there is some small PT inflation we did not have before.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+The chain will not accept proofs that cannot be disputed.
+
+1. Dispute messages are always a constant size, regardless of the size of the proof being disputed.
+2. Proofs are still limited to proving 4 partitions at a time. The amount of work required by any given dispute is bounded by the maximum number of sectors/partitions that could have been proven.
+
+### Power & Sector Extension
+
+Given that proofs are accepted optimistically, an attacker can keep a faulty sector "active" by submitting invalid window PoSts. However:
+
+1. If these posts are successfully disputed, the attacker will lose 5.1 times the expected block reward per-day for the incorrectly proven sectors.
```
irenegia, line 185: 5.1? 
I think it is 5.51 (FF + 2*BR), and for completeness I would add the this value is chosen to give a negative profit for miner who sends invalid proofs when we have that an invalid proof is detected in 12h (after deadline) with probability ≥ 1/3.

PatrickDeuse:
> The miner submits a proof. <-- only miner interaction
> New: The chain stores it temporarily in the miner's state.
> New: Someone disputes it (calls DisputeWindowedPoSt).
> The network verifies the proof.

who is "paying" for the last step?

anorth:
> So if someone disputes a proof 1 time and its actually a valid proof, does the miner that disputes it get penalized

@funktafide the party that disputes it pays the gas cost of checking the proof validity, but earns no reward that might cover that cost if the dispute fails. This is not any more prone to spam than any other method. @Stebalien was worried about the case of a miner submitting many bad proofs, thus kind of obligating the network to process valid disputes for them all. It's a very expensive attack. (I expect future proof batch/aggregation techniques to dramatically reduce the cost of disputing en-masse.)

> Can one proof being disputed multiple times within 1800 epochs by the same challenger or different challengers

@jennijuju an invalid proof can be disputed only once. A valid proof can be disputed repeatedly (at cost to the disputer, and will always fail to penalize/reward).

> everyone can gain power in the network regardless of congestion and fee market

@BrunoZell thanks this is a great perspective. I think there's a lot of support for that vision long term, we'd love your help figuring it out. The control/data plane model might be a good fit, or others. But let's take that discussion elsewhere (#24, #47, or a new issue) – it's complementary to this technique of moving most proof verification off-chain, which we will want anyway.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
+---
+fip: 0010
+title: Off-Chain Window PoSt Verification
+author: Alex (@anorth), Steven (@stebalien), et al.
```
anorth, line 4: I would prefer you to be first author here: you figured out all the gnarly details.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+  - <section-id>
+---
+
+## Simple Summary
+
+Optimistically accept Window PoSt proofs on-chain without verification, allowing them to be disputed later by off-chain verifiers.
+
+## Abstract
+
+When a miner proves continued storage of data to the chain (`SubmitWindowedPoSt`), optimistically accept and record the proof on-chain instead of verifying it. After the chain has accepted the proof, a third party may dispute it by invoking `DisputeWindowedPoSt`. A successfully dispute marks the incorrectly proven sectors as faulty, removes the associated power (until proven again in a subsequent Window PoSt), and fines the miner proportional to the expected block reward received from the incorrectly proved sectors.
```
anorth, line 21:
```suggestion
When a miner proves continued storage of data to the chain (`SubmitWindowedPoSt`), optimistically accept and record the proof on-chain instead of verifying it. After the chain has accepted the proof, a third party may dispute it by invoking `DisputeWindowedPoSt`. A successful dispute marks the incorrectly proven sectors as faulty, removes the associated power (until proven again in a subsequent Window PoSt), and fines the miner proportional to the expected block reward received from the incorrectly proved sectors.
```

BrunoZell:
I was suspecting this would be out of scope for this issue, still finding my way around FIL development. Thanks for the pointers!

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+This section addresses:
+
+* Chain state size attacks.
+* Proof disputability attacks.
+* Power & sector extension attacks.
+* Vesting funds extraction.
+
+### Chain State Size
+
+This change should not appreciably increase the size of chain state. The stored proofs are small (190 bytes) only one proof may be submitted per partition with live sectors.
```
anorth, line 172: Do the arithmetic. 1EiB of densely packed sectors is 13164 partitions -> 2.5MB per EiB.

Note that the snapshots are ~free  assuming nodes store full state history for at least one proving period anyway.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+### Power & Sector Extension
+
+Given that proofs are accepted optimistically, an attacker can keep a faulty sector "active" by submitting invalid window PoSts. However:
+
+1. If these posts are successfully disputed, the attacker will lose 5.1 times the expected block reward per-day for the incorrectly proven sectors.
+2. Restoring faulty sectors requires an on-chain proof. This prevents attackers from extending faulty sectors indefinitely by repeatedly submitting faulty proofs.
+
+### Vesting Fund Extraction
+
+`DisputeWindowedPost` takes feeds from "vesting funds", but burns any fees taken. Therefor, it cannot be used to "take an advance" on vesting funds.
```
anorth, line 190: We need to describe the reward above, and then outline its potential for extraction here.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+
+1. Unlocked/spendable funds.
+2. Collateral.
+3. Vesting rewards.
+
+Any rational miner submitting invalid proofs will drain their unlocked funds down to 0. That means funds would need to come from either collateral or vesting rewards. Unfortunately, taking from either would affect circulating supply and could be used to extract "locked" funds.
+
+### Including Proof Dispute Messages Is Rational
+
+Proof dispute messages directly reduce the power of other miners on the network, proportionally increasing the relative power of all other miners in the network. Therefore, baring collusion, it is always rational to include dispute messages.
```
anorth, line 220: Expand this to note that in fact plagiarising disputes made by other parties is the rational behaviour if reward exceeds base fee.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Product Considerations
+
+This FIP will increase chain bandwidth by moving Window PoSt processing off-chain. This will reduce the per-gigabyte network-wide maintenance cost, removing a scalability constraint on the total amount of storage that can be supported by the network.
```
anorth, line 224: Address benefits specifically to small miners

anorth (commented):
A suggestion you are free to ignore: I find that formatting markdown files with a line break after every sentence makes targeting review comments much easier. It's also more sympathetic to view in a non-wrapping editor (like many people's code editors).

anorth:
This change increases the length of chain history that must be accessible at low cost by the VM in order to fetch a challenge seed from `dispute period + challenge lookback` in the past. This period is less that the period required for pre-commit randomness, so likely no implementation change, but FYI @magik6k @arajasek to ensure that both are taken as bounds.

FIPS/fip-0010.md
Stebalien: In terms of "minimizing power", I agree that the argument is very weak and I'm happy to remove it. From my perspective, the only argument that really works is the "altruistic" (ish) one. That is, large enough players will verify proofs because:

1. It should be relatively cheap. At the moment, everyone is verifying all proofs. Even if we grow by 10x or 100x, verifying proofs should still be relatively inexpensive compared to the cost of actually mining on Filecoin.
2. If a significant fraction of proofs on Filecoin are incorrect and go undisputed, people will lose trust in the Filecoin network devaluing the token.

Stebalien: Updated. How does the new text read?

FIPS/fip-0010.md
Stebalien: done.

FIPS/fip-0010.md
Stebalien: Done.

Stebalien:
> A suggestion you are free to ignore: I find that formatting markdown files with a line break after every sentence makes targeting review comments much easier. It's also more sympathetic to view in a non-wrapping editor (like many people's code editors).

Yeah, I'll switch it. I've stopped inserting line breaks because I've optimized my editor work around the lack there of... but I agree.

FIPS/fip-0010.md
Stebalien: I'm happy to take the credit and the blame.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,440 @@
...
+2. Restoring faulty sectors requires an on-chain proof. This prevents attackers from extending
+   faulty sectors indefinitely by repeatedly submitting faulty proofs.
+
+5.51 was chosen as it gives negative profit for miners as long as invalid proofs are caught within
+12 hours with probability 1/3.
+
+### Vesting Fund Extraction
+
+`DisputeWindowedPost` takes feeds from "vesting funds", but burns any fees taken. Therefor, it
+cannot be used to "take an advance" on vesting funds.
```
anorth, line 380: I think this is untrue if there is a non-zero reward. We should note here the maximum fund extraction rate for a self-destructing miner (and that it is lower than that already possible from declaring consensus faults). cc @Kubuxu for details.

anorth (approved):

arajasek (approved):

FIPS/fip-0010.md
Stebalien: Yep. Updating now.

Stebalien: The rate is a bit tricky to calculate. The rate for slashing is one reward per day but the rate here is one reward per partition per 2 days. On top of that, the fine to reward ratio depends on how "full" each partition is.

I'd like to consider setting a minimum fine. That would make it easy to reason about this in terms of a tax. E.g., if the minimum fine is 3x the reward, the "advance" would have a 75% tax.

FIPS/fip-0010.md
Stebalien: Mostly done. We still need to pin down the exact reward.

FIPS/fip-0010.md
```diff
@@ -0,0 +1,450 @@
...
+
+In the future, additional state could be tracked in the `Deadline` object to avoid loading
+partitions when "optimistically" accepting a proof.
+
+See:
+[@filecoin-project/specs-actors#1352](https://github.com/filecoin-project/specs-actors/issues/1352)
+
+## Backwards Compatibility
+
+As this is a consensus breaking change, all Filecoin implementer must implement the changes to the
```
arajasek, line 291:
```suggestion
As this is a consensus breaking change, all Filecoin implementers must implement the changes to the
```

PR #66, opened 2021-01-15
FIP 0010: describe proof spam attack

Stebalien:
This patch describes a "spam" attack with faulty proofs and explains why it's unlikely to be rational.

### DIFF
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -386,9 +386,9 @@ proofs, and disputing their own proofs. However:
 
 1. This attack is rate-limited. It can only be used to extract one reward per active partition every
    two days.
-3. Another miner may dispute the proof first and steal the reward. This is especially true because
+2. Another miner may dispute the proof first and steal the reward. This is especially true because
    the attack is slow so other miners will have a chance to notice it happening.
-4. The attack is expensive. The minimum fine is 20FIL, and the reward is 4FIL. This means an
+3. The attack is expensive. The minimum fine is 20FIL, and the reward is 4FIL. This means an
    attacker would lose at least 5/6th of their vesting FIL.
 
 ### Proof Spam Attack
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -407,7 +407,7 @@ Most rational miners will simply ignore these proofs as the target miner cannot 
 reward. However, for the good of the network, some parties may wish to dispute any/all invalid
 proofs.
 
-Luckily, even though the miner in question will no _pay out_ rewards to the party that submitted the
+Luckily, even though the miner in question will offer no _pay out_ rewards to the party that submitted the
 dispute, they still need to pay the penalty and rewards eventually or abandon the miner. That is, unpaid penalties and rewards are
 added to the miner's debt and the miner will be unable to declare their sectors "recovered" until
 they pay their debts.
--- a/FIPS/fip-0010.md
+++ b/FIPS/fip-0010.md
@@ -325,6 +325,7 @@ This section addresses:
 * Proof disputability attacks.
 * Power & sector extension attacks.
 * Vesting funds extraction.
+* Proof spam attacks.
 
 ### Chain State Size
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0010.md
```diff
@@ -385,9 +386,35 @@ proofs, and disputing their own proofs. However:
 
 1. This attack is rate-limited. It can only be used to extract one reward per active partition every
    two days.
-2. Another miner may dispute the proof first and steal the reward. This is especially true because
+3. Another miner may dispute the proof first and steal the reward. This is especially true because
```
anorth, line 389: This numbering looks odd

FIPS/fip-0010.md
```diff
@@ -385,9 +386,35 @@ proofs, and disputing their own proofs. However:
...
+
+Given that this miner will have less power than necessary to mine blocks, the "vesting rewards" will
+be empty. Therefore, disputes will pay out no rewards.
+
+Most rational miners will simply ignore these proofs as the target miner cannot pay out a
+reward. However, for the good of the network, some parties may wish to dispute any/all invalid
+proofs.
+
+Luckily, even though the miner in question will no _pay out_ rewards to the party that submitted the
+dispute, they still need to pay them eventually or abandon the miner. That is, unpaid rewards are
```
anorth, line 411:
```suggestion
dispute, they still need to pay the penalty and rewards eventually or abandon the miner. That is, unpaid penalties and rewards are
```

anorth:
Another possible path to a proof spam attack is a miner who has lots of power but has lost access to the sectors for some time and cannot submit a valid Window PoSt. 

If the sectors are gone for good, the miner choses between terminating the sectors immediately, or attempting vesting funds extraction before doing so. So this case resembles the previous one, but the angle here is a large volume of proofs to be verified. The miner will race other parties to submit the disputes, and hope to win some races to extract some funds. But I guess at equilibrium the miner should expect to lose most of those races except where they mine a block (because the disputes should be stolen by block producers).

If the sectors are not gone for good, the miner choses between missing the Window PoSts or submitting fakes until they can recover. So long as the penalty associated with a fraudulent proof exceeds that associated with a missed PoSt, and the miner expects to recover to vest their funds, there seems to be no incentive to pay the higher fraud penalty (which they can only choose to do once per partition anyway). Maybe the worst case here is that a large miner submits all fraudulent PoSts, consuming ~all the chain bandwidth to verify them, pushing the base fee up to (or past) the reward breakeven rate until they can all be processed.

I think it would be helpful for us to calculate the fraud proof rate that would saturate chain bandwidth, and hence the storage power associated with a deadline which, if fully fraudulent, would raise the base fee to this level for 30min. *This is an argument for keeping the dispute reward modest, rather than generous.* And maybe an argument for limiting maximum miner actor size.

Stebalien:
> If the sectors are gone for good, the miner choses between terminating the sectors immediately, or attempting vesting funds extraction before doing so. So this case resembles the previous one, but the angle here is a large volume of proofs to be verified. The miner will race other parties to submit the disputes, and hope to win some races to extract some funds. But I guess at equilibrium the miner should expect to lose most of those races except where they mine a block (because the disputes should be stolen by block producers).

This should be very rate limited but I'll write it up. Basically, a miner can recover 4FIL per partition, losing 20FIL plus IPF in the process. Given _current_ network rewards (0.128FIL/TiB), 4FIL is less than half a days worth of block rewards for a full partition. 

1. They cannot repeat the same attack against the same partition because they'd need to have the sectors to actually recover the partition.
2. They must perform the attack _immediately_ on losing the sectors, because the attack becomes impossible once the chain marks them as faulty.

Stebalien:
> They cannot repeat the same attack against the same partition because they'd need to have the sectors to actually recover the partition.

Well, they might have lost _most_ of their storage, but not all.

jennijuju:
@Stebalien does this pr describes the final FIP and ready to merge?

Stebalien:
Yes.

jennijuju (approved):

FIPS/fip-0010.md
```diff
@@ -387,7 +388,33 @@ proofs, and disputing their own proofs. However:
...
+3. Start proving the sector with bad proofs.
+
+Given that this miner will have less power than necessary to mine blocks, the "vesting rewards" will
+be empty. Therefore, disputes will pay out no rewards.
+
+Most rational miners will simply ignore these proofs as the target miner cannot pay out a
+reward. However, for the good of the network, some parties may wish to dispute any/all invalid
+proofs.
+
+Luckily, even though the miner in question will no _pay out_ rewards to the party that submitted the
```
arajasek, line 410:
```suggestion
Luckily, even though the miner in question will offer no _pay out_ rewards to the party that submitted the
```

arajasek (approved):
one nit

PR #82, opened 2021-03-16
IPP spec supplement

nikkolasg:


jennijuju:
@nikkolasg Is the proposal now ready to be moved to "Last Call" or there will be future edits?

nikkolasg:
I don't believe there is going to be majors edits. Minors edits will probably come, like the final URL of the paper, and details (minimum number of proofs allowed to be aggregated for example)

nicola:
We should merge it and as updates come put them in.
I don't think we are ready to call it "last call", we are still in "draft" mode until @ZenGround0 resolves on-going minor problems (I may be out of date)

jennijuju (approved):

Issue #57, opened 2020-12-30
Support non-deal data in sectors (off-chain deals part 1)


magik6k:
Problem
--

Currently Filecoin sectors can only store deal data referenced by on-chain deals. CC sectors and non-deal areas of sectors must store null (`\0`) bytes to be verifiable on-chain. This limitation is caused by how the UnsealedCID (CommD) is computed in the miner actor.

Unfortunately publishing deals is a very expensive on-chain operation, especially for smaller pieces where this cost it can be the main part of the storage fee. Off-chain deal support / alternative markets (e.g. more scalable storage market on ETH2.0 with data stored on Filecoin) seem to be good solutions to this problem, at least for non-FIL+ deals.

This proposal is only the first, but major, part of what needs to be done to support off-chain deals. It's likely that some off-chain deal protocols will require additional actor methods to be implemented (for example a method to check if a sector wasn't terminated early for settling payment channels)

Current state
--

Currently, when precommitting sectors, a number of DealIDs can be specified with `SectorPreCommitInfo`:

```go
// Information provided by a miner when pre-committing a sector.
type SectorPreCommitInfo struct {
	SealProof       abi.RegisteredSealProof
	SectorNumber    abi.SectorNumber
	SealedCID       cid.Cid `checked:"true"`
	SealRandEpoch   abi.ChainEpoch
	DealIDs         []abi.DealID    // ****
	Expiration      abi.ChainEpoch
	ReplaceCapacity bool
	ReplaceSectorDeadline  uint64
	ReplaceSectorPartition uint64
	ReplaceSectorNumber    abi.SectorNumber
}
```

This information is then stored in miner actor state, and used to compute UnsealedCID when verifying PoRep by turning specified DealIDs into `[]abi.PieceInfo`s, and calling the `ComputeUnsealedSectorCID` syscall through the storage market actor.

```go
type abi.PieceInfo struct {
	Size     abi.PaddedPieceSize
	PieceCID cid.Cid
}
```

```go
ComputeUnsealedSectorCID(reg abi.RegisteredSealProof, pieces []abi.PieceInfo) (cid.Cid, error)
```

Proposed solution
--

We can change or create a new version of the `SectorPreCommitInfo` miner actor struct, changing the `DealIDs` array to a new array which allows specifying non-deal PieceCIDs:
```diff
+type SectorPieceInfo struct{
+	PieceCID  *cid.Cid `checked:"true"` // CommP
+	PieceSize abi.PaddedPieceSize
[...]

PR #91, opened 2021-03-31
Pack arbitrary data in CC sectors and be able to upgrade them to verified data later.

i-life-i:


mko-io:
until the unsealing time can decrease to a 5min - 15min range, storage any real data is not economical.

FIPS/fip-0015.md
```diff
@@ -0,0 +1,70 @@
+---
+fip: "0015"
+title: Pack arbitrary data in CC sectors 
+author: donghengzhao (@1475)
+discussions-to: https://github.com/filecoin-project/FIPs/issues/89
```
jennijuju, line 5:
```suggestion
discussions-to: https://github.com/filecoin-project/FIPs/issues/57
```

Stebalien: These aren't quite the same. This FIP is about packing arbitrary data in CC sectors (which can be upgraded later and are therefore unsuitable for off-chain deals).

FIPS/fip-0015.md
```diff
@@ -0,0 +1,70 @@
...
+
+Currently in seal process, miners pack empty pieces (all zeros) into CC sectors, after porep finished, create precommit
+messages with empty dealID information. In verification of commit messages, verifier will get the DealID from
+corresponding precommit info, and if DealID is empty, verifier recover the unsealed id by all zeros, which limit the CC
+sector to seal only zeroed data.
+
+## Change Motivation
+
+We propose design to allow miners to pack real data for CC sectors, add unsealed id in precommit message and keep it on
+chain. It introduce the possibility for CC upgrade without re-seal, and turn the piece into verified data afterwards (
```
Stebalien, line 28: Additional motivation: This could be used to store _replicas_ of deal data so it can be recovered if lost.

Stebalien: Nvm. I think I'm misreading the proposal (and this is already possible, apparently)

FIPS/fip-0015.md
```diff
@@ -0,0 +1,70 @@
+---
+fip: "0015"
```
arajasek, line 2: This number will change

FIPS/fip-0015.md
ZenGround0, line 28: I think the "CC upgrade without reseal" described is of the form 1) commit sector without deals and with piece cids (called unsealed IDs here)  2) After sector is sealed create a (verified) deal for this data 3) Upgrade already sealed sector to fulfill terms of a deal in the case its non-deal piece cids cover the deal.

This makes sense.  However note the more general "CC Upgrade without reseal" process allowing a cc sector to be upgraded to *any* data not declared at commit actually depends on all CC sectors storing the same data which this FIP would change.

FIPS/fip-0015.md
```diff
@@ -0,0 +1,70 @@
...
+
+Fully backwards compatibility is expected
+
+## Security Considerations
+
+N/A
+
+## Incentive Considerations
+
+No impact on existing incentives, but provides extra flexibility to miners to pack real data in CC sectors. As the
```
arajasek, line 60: I think this section needs more detail. There _are_ significant impacts to incentives here, since people can now store their data in "implicit deals" with miners.

arajasek (commented):
Thanks a lot for the PR! I've flagged some sections for improvement. Some other topics of discussion:

- @jbenet notes that the FIP would be more motivated if we contrasted this with the status quo, which is for a miner to create one big deal (a single piece that's the size of the entire sector)
- We should discuss whether we want to expand the scope of this FIP to also allow non-zero padding pieces in deals that DO have sectors (see https://github.com/filecoin-project/FIPs/issues/57)

arajasek:
Once some of the requested additions are made, I'd be happy (and excited) to land this PR!

jennijuju (approved):

jennijuju:
The FIP format is good as a `Draft`.

PR #94, opened 2021-04-22
FIP 00013 Proposing Last Call

ZenGround0:
Editing FIP 0013, adding some last mile details, proposing last call

### DIFF
--- a/FIPS/fip-0013.md
+++ b/FIPS/fip-0013.md
@@ -1,7 +1,7 @@
 ---
 fip: "0013"
 title: Add ProveCommitSectorAggregated method to reduce on-chain congestion
-author: Nicola (@nicola), Zenground0, nemo (@cryptonemo), nikkolasg
+author: Anca (@ninitrava), Nicola (@nicola), Zenground0, nemo (@cryptonemo), nikkolasg
 discussions-to: https://github.com/filecoin-project/FIPs/issues/50
 status: Last Call
 type: Technical
@@ -92,7 +92,7 @@ type ProveCommitSectorAggregatedParams {
 Semantics will be similar to those of `ProveCommitSector` with the following proposed changes:
 
 - Use a SectorNumber bitfield in place of a single abi.SectorNumber in parameters
-- Include a Aggregate proof in place of single porep proof in parameters
+- Include an Aggregate proof in place of single porep proof in parameters
 - MaxProveCommitSize parameter will change
 - Minimum and maximum number of sectors proven will be enforced
 - PreCommitInfos read in batch
@@ -269,14 +269,10 @@ smooth operation through the upgrade period.
 ### Failure handling
 
 Aborting on any precondition failure is chosen for simplicity. 
-Submitting an invalid prove commitment should
-never happen for correctly-functioning miners.  Aborting on failure will provide
-a clear indication that something is wrong, which might be overlooked by an
-operator otherwise.
+Submitting an invalid prove commitment should never happen for correctly-functioning miners.  Aborting on failure will provide
+a clear indication that something is wrong, which might be overlooked by an operator otherwise.
 
-An alternative could be to allow sectors in the aggregation to succeed or fail independently. 
-In this case, the method should return a value indicating which sectors succeeded and which failed.
-This would complicate both the actor and node implementations somewhat, though not unduly.
+Submitting an aggregate including an already proven sector is a failure.
 
 ### Scale and limits
 
--- a/FIPS/fip-0013.md
+++ b/FIPS/fip-0013.md
@@ -3,7 +3,7 @@ fip: "0013"
 title: Add ProveCommitSectorAggregated method to reduce on-chain congestion
 author: Nicola (@nicola), Zenground0, nemo (@cryptonemo), nikkolasg
 discussions-to: https://github.com/filecoin-project/FIPs/issues/50
-status: Draft
+status: Last Call
 type: Technical
 category: Core
 created: 2021-17-02
--- a/FIPS/fip-0013.md
+++ b/FIPS/fip-0013.md
@@ -27,11 +27,11 @@ is currently limiting network growth.
 The miner `ProveCommitSector` method only supports committing a single sector at
 a time.  It is both frequently executed and expensive.  This proposal adds a
 way for miners to post multiple `ProveCommit`s at once in an aggregated fashion
-using a `ProveCommitSectorAggregated`. This method amortize some of the costs
+using a `ProveCommitSectorAggregated`. This method amortizes some of the costs
 across multiple sectors, removes some redundant but costly checks and
 drastically reduces per-sector proof size and verification times taking
 advantage of a novel cryptography result. This proposal also increases the delay
-between pre-commit an prove-commit to allow miners of all sizes to enjoy the
+between pre-commit and prove-commit to allow miners of all sizes to enjoy the
 highest gas saving factor.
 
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0013.md
```diff
@@ -109,11 +110,25 @@ Semantics will be similar to those of `ProveCommitSector` with the following pro
...
+The number of sectors that may be proven in a single aggregation
+is a minimum of 12 and a maximum of 203
+
+MaxProveCommitDuration, the enforced delay between pre-commit and prove-commit is also increased from 1 day + PreCommitChallengeDelay to 6 days + PreCommitChallengeDelay. This does not impact security of the system it only increases the possible time pre-commits remain on chain before expiry.  See the section about incentives for motivation.
+
+#### Gas calculations
+
+Similar to existing PoRep gas charges, gas values will be determined from empirical measurements of aggregate proof validation on representative hardware. Each PoRep count corresponding to a power of two snark count will be assigned to a gas table. See the "Proof scheme changes" section for discussion on padding to the next power of two for motivation. The number of sectors proven in a call to `ProveCommitSectorAggregated` will be rounded up to the nearest number in the table to calculate the gas value. Currently the table looks like this (subject to change)
+
+- 12 PoReps (128 snarks): 300M
```
ZenGround0, line 122: @nikkolasg note that there are some slight differences between the points cryptolab has measured performance at in the past (13, 26, 51, 102, 205) and the numbers here that I've computed as being the largest number of poreps that can be padded to the specified number of snarks. Please correct this if I am missing something.

ZenGround0: @cryptonemo because of padding I believe every aggregate in the the given PoRep batch size ranges validates exactly the same number of snarks. Please correct me if I am oversimplifying.

FIPS/fip-0013.md
```diff
@@ -133,13 +148,18 @@ A more detailed technical report on the new constructions can be found here (TOD
 
 #### Proofs API
 
+##### RegisteredAggregationProof
+
+The API changes introduce a new type RegisteredAggregationProof for future proofing the aggregation scheme. As of this change there is only one valid member of the RegisteredAggregationProof type corresponding to the new aggregation scheme.
```
ZenGround0, line 153: @cryptonemo this is a proposal we can iterate on before merging this to Last Call

FIPS/fip-0013.md
```diff
@@ -310,17 +333,18 @@ miner that can aggregate 800 proofs in a day will be able to enjoy a 20x gas
...
-`ProveCommitSectorAggregated` do not yield a benefit when using them in batch.
+`ProveCommitSectorAggregated` do not scale logarithmically with batch size.
 
-In order to support all size of miners on the Filecoin network, this FIP
+In order to support similar gas improvements for all sizes of miners on the Filecoin network, this FIP
 proposes to:
+
 1. Increase the maximum delay between pre-commit and prove-commit to 6 days 
-2. Limit the number of aggregated proofs to 203 (`=floor(2048/10)`)
+2. Limit the number of aggregated proofs to 204 (`=floor(2048/10)`)
```
ZenGround0, line 342: @nikkolasg I updated this to match the expression and my understanding that we want to allow the max number of PoReps with the 2048 snark validation gas cost / time.  Let me know if this overlooks something important..

FIPS/fip-0013.md
```diff
@@ -109,11 +110,25 @@ Semantics will be similar to those of `ProveCommitSector` with the following pro
 
 #### Scale and limits
 
-The number of sectors that may be pre-committed in a single aggregation starts
-from 50 (TODO) and a maximum of 203 (TODO)
+The number of sectors that may be proven in a single aggregation
+is a minimum of 12 and a maximum of 204
```
ZenGround0, line 114: @nikkolasg from what I've heard we might be able to set this lower limit as low as 4 now.

FIPS/fip-0013.md
```diff
@@ -3,7 +3,7 @@ fip: "0013"
 title: Add ProveCommitSectorAggregated method to reduce on-chain congestion
 author: Nicola (@nicola), Zenground0, nemo (@cryptonemo), nikkolasg
```
nicola, line 4: add Anca to this list

FIPS/fip-0013.md
```diff
@@ -90,17 +91,17 @@ type ProveCommitSectorAggregatedParams {
 
 Semantics will be similar to those of `ProveCommitSector` with the following proposed changes:
 
-- Sectornumber bitfield in place of a single abi.SectorNumber in parameters
-- Aggregate proof in place of single porep proof in parameters
+- Use a SectorNumber bitfield in place of a single abi.SectorNumber in parameters
+- Include a Aggregate proof in place of single porep proof in parameters
```
nicola, line 95: an*

FIPS/fip-0013.md
```diff
@@ -248,7 +269,7 @@ smooth operation through the upgrade period.
 ### Failure handling
 
 Aborting on any precondition failure is chosen for simplicity. 
-There is no good reason for submitting an invalid pre-commitment, so this should
+Submitting an invalid prove commitment should
 never happen for correctly-functioning miners.  Aborting on failure will provide
 a clear indication that something is wrong, which might be overlooked by an
 operator otherwise.
```
nicola, line 275: line 277 talks about an alternative - if we move to last call we should remove "alternatives".

nicola, line 275: Can we clarify what happens what happens if a sector has been already prove committed?

FIPS/fip-0013.md
nikkolasg: That's correct but to be confirmed with a full e2e benchmark since the benchmark I run doesn't take into account the IPLD calls.

FIPS/fip-0013.md
nikkolasg: No that looks correct !

FIPS/fip-0013.md
nikkolasg: The correct answer I believe is that the call should return an error.

FIPS/fip-0013.md
cryptonemo: Yes, that should work -- will just have to update the code.

FIPS/fip-0013.md
cryptonemo: I find this confusing, @nikkolasg are you following it correctly?  I think of each PoRep (to be aggregated) as already being a completed snark, so not sure how the number is mapping at the moment.

FIPS/fip-0013.md
cryptonemo: Do we have a suitably named instance of the RegisteredAggregationProof type (or do you have one in mind)?  Otherwise, something very drab like `IppV1` or something may be a default candidate.

jayr2trill56 (approved):

jayr2trill56 (commented):
FR2MVVUR47FLQ6X4

jayr2trill56 (approved):

FIPS/fip-0013.md
ZenGround0: @cryptonemo I've been under the assumption that every PoRep contributes 10 snarks to the aggregation.  So 12 PoReps is 120 snarks and is the biggest number that fits into the 2^7 sized snark aggregation. 13 PoReps would have 2 too many so must be packed into the 2^8 aggregation with padding.

Let me know if this gets something wrong.

cryptonemo: Yes, agreed, sorry was thinking about it incorrectly at the time.  Thanks!

FIPS/fip-0013.md
```diff
@@ -109,11 +110,25 @@ Semantics will be similar to those of `ProveCommitSector` with the following pro
...
+
+#### Gas calculations
+
+Similar to existing PoRep gas charges, gas values will be determined from empirical measurements of aggregate proof validation on representative hardware. Each PoRep count corresponding to a power of two snark count will be assigned to a gas table. See the "Proof scheme changes" section for discussion on padding to the next power of two for motivation. The number of sectors proven in a call to `ProveCommitSectorAggregated` will be rounded up to the nearest number in the table to calculate the gas value. Currently the table looks like this (subject to change)
+
+- 12 PoReps (128 snarks): 300M
+- 13-25 PoReps (256 snarks): 300M
+- 26-51 PoReps (512 snarks): 350M
+- 52-102 PoReps (1024 snarks): 350M
+- 103-204 PoReps (2048 snarks): 350M
```
ZenGround0, line 126: These gas values will be finalized after lotus integration using the existing lotus-bench methodology for gas estimation.

jennijuju (approved):

PR #109, opened 2021-05-31
Add audit section to FIP0009

nikkolasg:
Add the reports of the two audits on the cryptographic part of FIP0009


arajasek:
@nikkolasg i don't think this was opened against the right FIP (you want FIP-0013 i'm pretty sure)

PR #118, opened 2021-07-03
add impl description for fip16

lyswifter:
Add some impl description of fip-0016

### DIFF
--- a/FIPS/fip-0016.md
+++ b/FIPS/fip-0016.md
@@ -62,7 +62,122 @@ This proposal introduces the possibilities for further improvement (which could 
 
 ## Implementation
 
-WIP
+As the design descripted above, we need a new package version of v6 and network migration to v14.
+
+1. add PieceInfos field to SectorPrecommitInfo and SectorOnChainInfo struct to store piece data cid and piece size for forward verify.
+```
+type SectorPreCommitInfo struct {
+	SealProof       abi.RegisteredSealProof
+	SectorNumber    abi.SectorNumber
+	SealedCID       cid.Cid `checked:"true"` // CommR
+	SealRandEpoch   abi.ChainEpoch
+	DealIDs         []abi.DealID
+	Expiration      abi.ChainEpoch
+	PieceInfos      []abi.PieceInfo
+	ReplaceCapacity bool // Whether to replace a "committed capacity" no-deal sector (requires non-empty DealIDs)
+	// The committed capacity sector to replace, and it's deadline/partition location
+	ReplaceSectorDeadline  uint64
+	ReplaceSectorPartition uint64
+	ReplaceSectorNumber    abi.SectorNumber
+}
+
+type SectorOnChainInfo struct {
+	SectorNumber          abi.SectorNumber
+	SealProof             abi.RegisteredSealProof // The seal proof type implies the PoSt proof/s
+	SealedCID             cid.Cid                 // CommR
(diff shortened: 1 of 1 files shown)


FIPS/fip-0016.md
```diff
@@ -62,7 +62,121 @@ This proposal introduces the possibilities for further improvement (which could
...
+	VerifiedDealWeight    abi.DealWeight  // Integral of active verified deals over sector lifetime
+	PieceInfos            []abi.PieceInfo //PieceInfos
+	...
+}
+```
+
+Rightnow, the `PieceInfos  []abi.PieceInfo` just present meaning of this field, for real environmnet it will be a cid.Cid to save onchain space. 
+```
+PieceInfos     cid.Cid 
+```
```
anorth, line 99: The size is a necessary part of `PieceInfo` for the CommD calculations.

FIPS/fip-0016.md
```diff
@@ -62,7 +62,121 @@ This proposal introduces the possibilities for further improvement (which could
...
+
+2. change getVerifyInfo for diffrent way to get unsealedCIDs of cc sectors with piece data inside, if in single sector proven verify
+```
+builtin.RequireState(rt, len(params.DealIDs) > 0 && len(params.PieceInfos) > 0, "deals and raw pieces cannot exist at the same time in one sector")
+
+commDs := requestUnsealedSectorCIDs(rt, &market.SectorDataSpec{
+	SectorType: params.RegisteredSealProof,
+	DealIDs:    params.DealIDs,
+	Pieces:     params.PieceInfos,
+})
```
anorth, line 109: Calling the market actor for something that is explicitly trying to sidestep the on-chain market seems wrong. We should instead either refactor that method to return the the PieceInfos, or otherwise invoke `rt.ComputeUnsealedSectorCID` here by duplicating the relevant logic. This comment applies to all the code in (3) too.

anorth (commented):
This detail adds some clarity to the proposal, thanks.

I will note, however, that I'm opposed to this FIP. I'll take my discussion to #57.

FIPS/fip-0016.md
lyswifter: image

I do store pieceCid and pieceSize info in miner's state. but current I think we need more thinking and disscussion as your above #57.

FIPS/fip-0016.md
lyswifter: the whole code is here, maybe you can review it and give me some suggestions. [feat/v6](https://github.com/lyswifter/specs-actors/tree/feat/v6)

anorth:
@lyswifter let's close out this PR one way or the other. We can close the PR independent of whether the FIP is complete and/or implementation teams like it.

We can either close the PR without merging, or update and merge it. I don't want to discard the work you've done here, so I'm happy to approve and merge if you review and make any changes before then.

kaitlin-beegle:
> @lyswifter let's close out this PR one way or the other. We can close the PR independent of whether the FIP is complete and/or implementation teams like it.
> 
> We can either close the PR without merging, or update and merge it. I don't want to discard the work you've done here, so I'm happy to approve and merge if you review and make any changes before then.

Given how old this PR is, but given that there's already a FIP number, I'm in favor of merging as-is and considering this FIP deferred.  Seems like the path of least resistance.  @anorth do you agree?

anorth (approved):

PR #130, opened 2021-07-27
FIP: Lightweight Sector updates

nicola:
Since 90+% of sectors in the Filecoin Network are CC sectors, having a protocol that allows for updating CC sectors to store real data without incurring in a full re-sealing would massively improve our network in terms of the amount of real data stored through it. 

1. It would allow **decoupling sealing latency from deal-making speed** - offering storage clients an improved experience for how quickly their data can land in on-chain deals
2. It would **unlock the 8EiB of storage already committed to the Filecoin network** to be quickly used for deals - enabling a 100PiB+ client to make deals for their entire dataset with a single miner like [f0127595](https://filfox.info/en/address/f0127595) which already has 120PiB of committed capacity.
3. It makes **utilizing existing committed capacity much cheaper for miners** (since they’ve already invested the sealing cost), increasing the chances they pay clients to add FIL+ data to these sectors.

### DIFF
--- a/FIPS/fip-0017.md
+++ b/FIPS/fip-0017.md
@@ -1,5 +1,5 @@
 ---
-fip: 0017
+fip: <to be assigned>
 title: Three-messages Lightweight Sector updates
 author: @nicola, @lucaniz, @irenegia
 discussions-to: <URL>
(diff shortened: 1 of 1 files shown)


FIPS/fip-0017.md
```diff
@@ -0,0 +1,175 @@
+---
+fip: <to be assigned>
+title: Three-messages Lightweight Sector updates
+author: @nicola, @lucaniz, @irenegia
+discussions-to: <URL>
```
jennijuju, line 5: Please create an issue for further discussion

jennijuju (changes requested):

jennijuju (approved):

PR #150, opened 2021-08-28
fix broken link to FIPs in README

marten-seemann:


bmann:
@marten-seemann good catch. Ideally, a permanent URL of `fips.filecoin.io` is going to get applied, and that link will get used instead.

momack2 (approved):

Issue #152, opened 2021-09-03
FIP Discussion: In ProveCommit[Sector | Aggregate] a single bad deal doesn't ruin the sector


ZenGround0:
# Summary

In the current protocol a single expired deal id in the dealIDs field of an on chain precommit info will cause market deal activation to fail and cause the sector to be dropped from the batch or aggregate of sectors being committed.  Instead this FIP proposes that the provider of a sector with such expired deals provide enough information for the commitment verification even when deals have expired.

# Motivation

Occasionally deals expire between PreCommit and ProveCommit and in these cases sectors containing these deals will fail to commit.  This is bad for miners and bad for the non-expired deals that would become active if the sector were committed.

# Design

The ProveCommit and method takes an array of pieceCids and two associated arrays of PieceSizes and dealIDs as arguments.  The ProveCommitAggregate method takes an array of such arrays.  In the common case these fields are empty.  In the case a miner has an expired deal they can specify the missing pieceCids their size and their DealIDs.

ComputeDataCommittment is now extended to take in associated arrays of dealIDs, pieceCids and PieceSizes so that the market actor has all the information needed to call the syscall for computing the sector's commD.  This method will also include a validation that all provided overwrite pieceCids belong to a dealID with a proposal no longer on chain and that this dealID is below the current max dealID.

The market ActivateDeals method will now return a bitfield of the indexes of dealIDs that are valid and will skip over expired deals instead of raising an error (similar to the approach proposed in #142). 

In the confirmSectorProofsValid subroutine the return to the market ActivateDeals method will be validated to match with the input expired dealIDs.  Expired dealIDs will then be removed from their sector's PrecommitInfo and the DealWeight and VerifiedDealWeight fields will be recomputed with a call to the market actor.  The created SectorInfo will be given these modified fields before being written to actor state.

# Considerations 

This proposal will increase the input parameter size for the common case of ProveCommit and ProveCommitAggregate messages with no expired dealIDs.  We should investigate wrapping ProveCommit params in a cbor array or pointer so that we can represent empty expired dealIDs as compactly as possible.

There is no significant risk of the pieceCids of expired deals being set to different data than that in the expired proposal because deal existence is checked in precommit and so there is no cost savings between expiring a deal and writing an arbitrary cid to the sector vs making a deal with the correct pieceCid in the first place.

We could optionally use this opportunity of changing the ActivateDeals signature to make this method work with batches of sector-deals.

[...]

PR #155, opened 2021-09-06
Commit with expired deals

ZenGround0:
Writing needs to be polished but the main ideas are here.  Discussion issue: https://github.com/filecoin-project/FIPs/issues/152

@whyrusleeping any fundamental reason cbor-gen [can't do integer keyed maps](https://github.com/whyrusleeping/cbor-gen/blob/master/gen.go#L382)?  I want this datatype for this proposal and I'm hoping to be able to add cbor-gen support in 1-2 engineer days but let me know if there's a reason that's unrealistic.

### DIFF
--- a/_fips/fip-0025.md
+++ b/_fips/fip-0025.md
@@ -1,9 +1,9 @@
 ---
-fip: <to be assigned>
+fip: 0025
 title: Handle Expired Deals in ProveCommit
 author: ZenGround0 (@ZenGround0)
 discussions-to: https://github.com/filecoin-project/FIPs/issues/152
-status: Draft
+status: Deferred
 type: Technical 
 category: Core
 created: 2021-09-06
@@ -25,6 +25,8 @@ Today storage providers are unable to commit sectors with expired deals because 
 ## Change Motivation
 <!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
 Occasionally deals expire between PreCommit and ProveCommit of a sector and in these cases sectors containing these deals will fail to commit. This is bad for storage providers committing sectors with deals and bad for the non-expired deals that would become active if the sector were committed. While offchain logic can help in many cases there is always a small risk of deals expiring even with optimal offchain logic as long as there is a delay between precommit and provecommit.
+
+This change is deferred (1) because the design adds a lot of complexity for a small useability gain (2) there is not broad agreement on what the storage provider SLA should be exactly and it is possible this change takes the network slightly in the wrong direction.
 
 
 ## Specification
--- a/_fips/fip-commit-expired-deals.md
+++ b/_fips/fip-commit-expired-deals.md
@@ -43,7 +43,7 @@ type ExpiredDealInfo struct {
 
 1. `ProveCommitSector` takes in an array of `ExpiredDealInfo`s as a new parameter.
 2. If all deals are expired the method fails and this sector cannot be committed.
-3. Market actor `ComputeDataCommitment` is extended to take in an array of `ExpiredDealInfo's. It validates that all expired deals are actually expired by asserting that each dealID is less than the global next dealID and the deal is not found on chain. If any expired deal fails validation the method will error, failing the caller. After validating `ComputeDataCommitment` then uses expired deal information to correctly compute the data commitment for `ProveCommitSector`.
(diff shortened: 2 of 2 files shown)


whyrusleeping:
@ZenGround0 that is a can of worms.

IPLD cannot do integer keyed maps, so as a result, cbor-gen does not support them.

ZenGround0:
@zixuanzh I have some cryptoecon TODOs in this FIP draft PR.  I think they are straightforwardly not problems but I don't want to move forward without your sign off

momack2:
@nicola - correct me if I'm wrong, but I believe if a miner never packs a deal into a sector, even after calling declare publish storage deals, they never lose a sector collateral (because all sectors they pack, not including that deal, are valid). Therefore I think we are already in the first case and should apply this change because it will have the beneficial impact of causing FEWER deals to need retries / fail to complete without actually changing the mental model. @arajasek @ZenGround0 - is that right?

arajasek:
@momack2 Yes, that's right.

More generally, I think the current state is that a clear SLA doesn't necessarily exist today -- the strongest statement that can be made is that a Storage Provider _will_ lose their deal collateral if they publish a deal and then something* goes wrong (and this FIP doesn't change that). I don't think this FIP worsens the situation. Going awry between Pre and Prove Commit is something of an edge-case regardless. 

What is uncontroversial, IMO, is that nothing is gained by "trapping" a PreCommitted sector in an unCommittable state, forcing all the other deals in the sector to also necessarily fail to be activated. Seems like a pretty clear net bad to me, so I'd strongly advocate for this FIP.


```
*something can be:
- never precommitting a sector with the deal
- precommitting but not provecommitting the deal
- provecommitting (activating) the deal, but it then goes faulty
- (probably more)
```

arajasek:
Thanks for the review, @nicola! 

> Questions:
> 
> * What happens today if the miner does not ever publish a sector for a published deal?
>   
>   * If the miner is does not lose sector collateral - then the system is already complex, so we may as well apply the change

Yeah, no sector collateral is lost, since none was ever locked up.

arajasek:
@ZenGround0 Can we please slightly rename / reword this FIP, though? `Expired` has a certain definition w.r.t. deals that isn't quite what we're talking about here. 

Idk what the right term is, though..."Past-activation-deadline" deals? :P

nicola:
> if a miner never packs a deal into a sector, even after calling declare publish storage deals, they never lose a sector collateral (because all sectors they pack, not including that deal, are valid).

I think that is true, which means that we already have strange edge cases (so maybe it's fine to add one more).

This is why I am trying to say that instead of adding edge cases we should try to simplify the protocol (or having a clearer Filecoin SLA)

ZenGround0:
> This is why I am trying to say that instead of adding edge cases we should try to simplify the protocol (or having a clearer Filecoin SLA)

While I wrote this FIP I think this is exactly right.  With a couple weeks of reflection I am not enthusiastic about the complexity and technical debt this approach adds to the system.  While we can't handle this offchain with *absolute* certainty, we should be able to get very high guarantees against deals activating before commitment and avoid adding on chain complexity with scheduling in node software.  

To the extent that this is a problem storage providers are dealing with and want a fix for now I think the first step should be looking at what went wrong during node scheduling before we move to consider on chain solutions like this FIP.

I'm in support of including this as a *Deferred* FIP since the design will be useful in informing more significant refactors of the market / miner interface and filecoin SLA.  I'm also fine with moving this to *Rejected* and shutting down the PR since the closed PR documents the design well enough.

jennijuju:
> This is why I am trying to say that instead of adding edge cases we should try to simplify the protocol (or having a clearer Filecoin SLA)

@nicola Why do you consider this an edge case? Also, a question, should the deals in the same sector, that may come from different clients, have an impact on other deals from a protocol perspective?

jennijuju:
I disagree this FIP should be `Deferred` based on the conversation happening here, id suggests us to merge the draft and have a discussion in the issue.

kaitlin-beegle:
Agreed; merging and moving to deferred.  Will merge and update to include FIP number and status.

anorth:
@ZenGround0 @kaitlin-beegle let's close this PR somehow. Kaitlin and I have chatted and I believe she's now on board with merging FIP PRs sooner than when the FIP is completed/accepted. I would support merging this as-is, as `Deferred`. Kaitlin, if you concur, would you lift your requested change?

I think this underlying issue will be resolved by a more significant refactor of the miner<->market interactions that I am planning as part of enhancing programability for the implementation of alternative deal markets on FVM. This is helpful as a motivating problem, and we can reject if after it's actually fixed by another means.

anorth (approved):

PR #162, opened 2021-09-14
create fip draft to require peerid when multiaddr is set

jennijuju:


_fips/fip-xxx.md
```diff
@@ -0,0 +1,71 @@
+---
+fip: to be assigned
+title: Miner onchain PeerId and Multiaddrs must co-exists
```
arajasek, line 3:
```suggestion
title: Miner onchain PeerId and Multiaddrs must co-exist
```

arajasek (commented):
Thanks for the PR!

My primary question would be whether we also want to enforce the converse (that we want to see a valid multiaddr if we try to set the PeerID).

If so, we can probably collapse the 2 setter methods.

@ZenGround0 opinions?

Kubuxu:
There is a use case (not implemented today) for just multiaddr. It would be a dnsaddr that resolves to multiaddr+PeerID pair.

It can be useful as it allows the miner to dynamically change the IP, port, protocol and PeerID of deal nodes. This would for example allow load balancing.

arajasek:
Thanks @Kubuxu! i think this can be closed.

jennijuju:
> There is a use case (not implemented today) for just multiaddr. It would be a dnsaddr that resolves to multiaddr+PeerID pair.
> 
> It can be useful as it allows the miner to dynamically change the IP, port, protocol and PeerID of deal nodes. This would for example allow load balancing.
> 
> Just PeerID can be resolved via DHT into multiaddr, this would allow above benefit in fully decentralised manner.
> 
> My suggestion for right now would be to add checks making “just multiaddr” invalid configuration when setting multiaddr via Lotus.

Thanks for the suggestion  @Kubuxu

PR #167, opened 2021-09-20
Correct quality calculation on expiration

ZenGround0:
Proposing a FIP to fixup an outstanding issue related to the sector quality of sectors with an extended expiration.


arajasek (approved):

arajasek:
A review from @zixuanzh would be good here

PR #171, opened 2021-09-21
move fip-0021 to the right folder

jennijuju:


kaitlin-beegle:
Oops, sorry about that @jennijuju.  I didn't realize you had already done this. In the PR I just submitted, I reorganize and rename all of the FIP files so they're in the right place.  Closing this request for now, since the PR I just submitted encompasses it.

Issue #187, opened 2021-09-29
Change type of Label in DealProposal to be []byte, rather than String


laudiacay:
Background
Also see https://github.com/filecoin-project/specs-actors/issues/1248 for important relevant discussion.

The DealProposal Label is not guaranteed to be UTF-8, but is assumed to be in some client implementations. This is causing trouble for said implementations. People suggested either enforcing UTF-8 (this is more difficult) or switching the type to be a byte array that can be interpreted as needed. Discussion settled on the latter.

Proposal
Change the type of the Label field to be a []byte, rather than a string. Draft PR with these changes inside is here: https://github.com/filecoin-project/specs-actors/pull/1496.

kaitlin-beegle:
Hi @laudiacay! It looks like work remains ongoing for this proposal, which is great. 

Once these details are set, a FIP draft can be written according to the guidance in FIP0001 and pushed to the repo.  I will then work through the process of facilitating community acceptance of the FIP draft. 

If you have any questions about this process, please feel free to ask!

laudiacay:
https://github.com/filecoin-project/specs-actors/issues/1248 Work done!

BlocksOnAChain:
@kaitlin-beegle can we move/do we have all the needed elements to move this for V15 upgrade since it’s  in "code complete " state from our side?
 CC: @jennijuju @laudiacay @arajasek

kaitlin-beegle:
@BlocksOnAChain you bet we do! Thanks for flagging that this is technically complete.  We can move this item to last call status this week.  I expect it will be approved without issue, but diligence requires that the last call period last for at least two weeks. 

I would expect that this item will be approved for v15 incorporation by Friday, January 21, but that implementation teams should prepare to add the change in their calibnet upgrades.

kaitlin-beegle:
After connecting with @laudiacay and @arajasek, it was confirmed that this FIP is no longer ready for 'last call' status. This FIP proposal will instead be moved back to the discussion forum, where it can be reconsidered for future upgrades. 

@laudiacay, it's my understanding that there are two technical paths that we can take to move this forward: pursuing a more technically complex implementation within the existing market modules, or waiting until markets are decoupled from implementations.  Any idea yet which is a better path?

PR #188, opened 2021-09-29
DealProposal Label type switch

laudiacay:
For issue #187

### DIFF
--- a/FIPS/fip-dealprop-label-type.md
+++ b/FIPS/fip-dealprop-label-type.md
@@ -1,5 +1,5 @@
 ---
-fip: <to-be-assigned>
+fip: "0027"
 title: Change type of DealProposal Label field from String to byte array
 author: Laudiacay (@laudiacay)
 discussions-to: https://github.com/filecoin-project/FIPs/issues/187
--- a/FIPS/fip-dealprop-label-type.md
+++ b/FIPS/fip-dealprop-label-type.md
@@ -16,11 +16,11 @@ spec-sections:
 
 ## Simple Summary
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
-DealProposal's Label field is currently a String, but is not enforced to be UTF-8. This is source of bugs and difficulties in client implementations because of variations in String libraries. Rather than enforce it as UTF-8 everywhere, this FIP changes the type to be raw bytes for ease of implementation and bug avoidance.
+DealProposal's Label field is currently a String, but is not enforced to be UTF-8. This is not only outside the CBOR string specification, but can also be a source of bugs and difficulties in client implementations due to variations in String libraries. Rather than enforce it as UTF-8 everywhere, this FIP changes the type to be raw bytes for ease of implementation, bug avoidance, and CBOR compliance.
 
 ## Abstract
 <!--A short (~200 word) description of the technical issue being addressed.-->
-String libraries differ in implementation of UTF-8. For maximum standardization and ease of bug-free client implementations, the DealProposal label should be raw bytes.
+String libraries differ in implementation of UTF-8. For maximum standardization and ease of bug-free client implementations, the DealProposal label should be raw bytes. Further, non-UTF-8 CBOR strings are technically outside of the CBOR specification, so the current implementation is outside of CBOR spec.
 
 ## Change Motivation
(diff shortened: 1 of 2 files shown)


laudiacay:
Hi- this is my first FIP, and I think it sounds very silly, but I wasn't sure what to put in some of the fields because this is such a tiny change. Please let me know what you think.

FIPS/fip-dealproposal-label-string-to-bytes.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Design Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+
+Design decisions are detailed extensively in above sections.
+
+## Backwards Compatibility
+<!--All FIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The FIP must explain how the author proposes to deal with these incompatibilities. FIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+
+This changes specs-actors. It should be backwards-compatible, in that the new version will accept any sort of String from the previous version. However, out-of-date clients that validate UTF-8 on this field may reject newly-valid non-UTF-8 labels.
```
ZenGround0, line 45: To reduce complexity all previous label fields should be moved from cbor strings to cbor bytes.  This accomplishes accepting all previous strings without introducing versioning of proposal data.

This change requires a migration which we'll need to describe here.  Specifically all deal proposals on chain will need to change the encoding of the label field.

FIPS/fip-dealproposal-label-string-to-bytes.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
+
+Checking that a non-UTF-8 string is accepted as a DealProposal Label will be sufficient to test this FIP.
```
ZenGround0, line 50: I think the problem is that non-UTF-8 strings are accepted, and this breaks the cbor spec.  

To test this FIP I think we just need to check that v0-v6 proposals labels become cbor bytes with the same raw byte payload after migration, and that new proposals have labels encoded as cbor bytes.

FIPS/fip-dealproposal-label-string-to-bytes.md
```diff
@@ -0,0 +1,73 @@
...
+created: 2021-09-29
+spec-sections: 
+  - specs-actors
+---
+
+<!--You can leave these HTML comments in your merged FIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new FIPs. Note that a FIP number will be assigned by an editor. When opening a pull request to submit your FIP, please use an abbreviated title in the filename, `fip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
```
ZenGround0, line 18: It's worth mentioning here or in the abstract that the proposal struct is technically invalid cbor with the acceptance of non-UTF-8 cbor strings.

FIPS/fip-dealproposal-label-string-to-bytes.md
laudiacay: is the current change correct? do I write the migration in specs-actors?

laudiacay:
TODO want to also add a bound on the length of the label field- this is not ready to go.

laudiacay:
> TODO want to also add a bound on the length of the label field- this is not ready to go.

never mind this was already done lol. just need to write the migration and this will be ready :)

laudiacay:
Migration done. I think this is ready to go? https://github.com/filecoin-project/specs-actors/pull/1496

ZenGround0 (approved):

FIPS/fip-dealprop-label-type.md
```diff
@@ -0,0 +1,73 @@
+---
+fip: <to-be-assigned>
```
kaitlin-beegle, line 2: FIP0027! Let's add it in, and we'll be ready for 'Last Call' status! :)

anorth:
@kaitlin-beegle and FIP editors, can we please merge this? Leave it in Draft if you want to align Last Call status with imminent implementation (though I would just go to Last Call). But leaving the PR open is no-one's ideal solution.

kaitlin-beegle (commented):
Just updated FIP status and FIP number; ready to merge.

PR #206, opened 2021-11-16
FIP-0019: Clarify that only deal-less sectors can be updated

arajasek:


FIPS/fip-0019.md
```diff
@@ -140,8 +140,8 @@ type SnapDealsUpdate struct {
 Unless stated assume that each step operates on all inputs updates in a loop
 
 1. **Param validation**
-   1. Assert that `SectorID` was originally CC sector
-   2. Assert that `SectorID` has no deals active
+   1. Assert that `SectorID` has no deals active
```
ZenGround0, line 143: This is a good spec for the future state of SnapDeals but I think that falls under a different fip. 

For this FIP the right check is "Assert that `SectorID` is CC, i.e. has no associated deal IDs"

ZenGround0 (approved):
LGTM

PR #208, opened 2021-11-16
Add activation update

ZenGround0:
Discovered this while reviewing implementation


arajasek (commented):
Hmm, why should we do that?

ZenGround0:
Short summary is that power attributed to the sector by the update should match the power calculation used by faults and penalties. There's more in the actors review once I have completed enough to send out more comments.

arajasek:
Gotcha, thanks!

arajasek (approved):

PR #221, opened 2021-12-10
Beneficiary address for miners

steven004:
This is to propose a beneficiary address to take finance control of a miner from owner. 
See the discussions from #253


FIPS/fip-0028.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+The following channges apply to specs-actors module:
+
+- Add **Beneficiary** address into **MinerInfo** structure
+- The **Beneficiary** is set to the same address of **Owner** when initiating a miner without specifying a beneficiary address (for back-compatibility)
+- The **Beneficiary** address could be a signaable address, or multisig address, or actor address (for smart contract)
+- The **WithdrawBalance** method of a miner can only be called by the miner's **Beneficiary**
```
anorth, line 43: We may wish to support the owner _calling_ WithdrawBalance, but the balance always being sent to the beneficiary. This permits the beneficiary to receive funds as a cold wallet (unfortunately prevented by the current implementation, where the owner must sign the withdrawal request). In the motivating scenario of lending FIL against future rewards, this permits the owner to make calls on behalf of the beneficiary, who may be far less technically sophisticated.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,121 @@
...
+    - The owner/beneficiary confirm the expiration height
+
+- *(TBC)* Add **SetBeneficiaryQuota** method to miner actor to propose or connfirm the expiration of a beneficiary. A beneficiary can only withdraw balance up to the quota. When the quota is met, the owner could take it back directly. There are 2 steps for this setting. 0 setting means no quota setting.
+    - The benneficiary/owner propose a beneficial quota
+    - The owner/beneficiary confirm the beneficial quota
+
+- Add **ChangeBeneficiaryAddress** method to miner actor to propose or confirm a change of beneficiary address. The owner can change the beneficiary if the current beneficiary is expired or the quota is used up, otherwise, there are 3 steps to go: 
+    - The current beneficiary address propose a change using the new beneficiary address as the parameter
+    - The owner address confirm the change by sending the same message
+    - The new beneficiary address accept the change by sending the same message
```
anorth, line 56: I suggest combining the above functionality into a single `ChangeBeneficiary` method which accepts:
- a beneficiary address
- an expiry
- a withdrawal quota

This permits the common case of an initial set-up as a single call, with no transition through a state where the beneficiary is set but has no expiry/quota. It permits mutual change to the expiry or to the quota.

Always requires confirmation by the _new_ beneficiary address, unless it is the owner, by submitting the same message (like ChangeOwnerAddress).
If there is a current beneficiary, also requires confirmation by the _existing_ beneficiary.

The code will probably be simpler if we mandate an order in which these calls/confirmations must be made. But UX would be slightly better if we can allow either the owner or current beneficiary to start the proposal.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,121 @@
...
+- The **Beneficiary** address could be a signaable address, or multisig address, or actor address (for smart contract)
+- The **WithdrawBalance** method of a miner can only be called by the miner's **Beneficiary**
+
+- *(TBD)* Add **SetBeneficiaryExpiration** method to miner actor to propose or connfirm the expiration of a beneficiary. When a beneficiary expired, the owner could take it back directly. There are 2 steps for this setting. 0 setting means no expiration setting
+    - The benneficiary/owner propose an expiration height
+    - The owner/beneficiary confirm the expiration height
+
+- *(TBC)* Add **SetBeneficiaryQuota** method to miner actor to propose or connfirm the expiration of a beneficiary. A beneficiary can only withdraw balance up to the quota. When the quota is met, the owner could take it back directly. There are 2 steps for this setting. 0 setting means no quota setting.
+    - The benneficiary/owner propose a beneficial quota
+    - The owner/beneficiary confirm the beneficial quota
```
anorth, line 51: A quota will require additional bookkeeping in miner state about how much has been sent to the beneficiary. We'd also need to determine how to treat changes to quota. Do we reset the amount sent when the quota is changed, or leave it in place (so the new quota is a new high-water mark). Do we clear the amount sent when the beneficiary address changes?

FIPS/fip-0028.md
```diff
@@ -0,0 +1,121 @@
...
+
+This FIP does not touch consensus, block generation.
+
+There is a concern that the beneficiary address might be controlled by people who are not so familiar with Filecoin, which may lead to some mistake of private key loss. Basically this can not be really handled well. In addition, the expiration date and quota setting is for kind of pretectioins.
+
+
+## Incentive Considerations
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
+
+There is no change to incentives.
```
anorth, line 107: I think we need a more thorough consideration of incentives to node operators who have no or little financial stake in their node.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,121 @@
...
+* The expiration and quota criteria set and work well
+* Beneficiary change logic works fine
+* Beneficiary support signable, multisig and actor address (this might be test later when FVM is in place)
+
+
+## Security Considerations
+
+This FIP does not touch consensus, block generation.
+
+There is a concern that the beneficiary address might be controlled by people who are not so familiar with Filecoin, which may lead to some mistake of private key loss. Basically this can not be really handled well. In addition, the expiration date and quota setting is for kind of pretectioins.
```
anorth, line 101: Using a multisig can handle this pretty well. Perhaps worth expanding on how different policies can be implemented with that.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,121 @@
...
+The following channges apply to specs-actors module:
+
+- Add **Beneficiary** address into **MinerInfo** structure
+- The **Beneficiary** is set to the same address of **Owner** when initiating a miner without specifying a beneficiary address (for back-compatibility)
+- The **Beneficiary** address could be a signaable address, or multisig address, or actor address (for smart contract)
+- The **WithdrawBalance** method of a miner can only be called by the miner's **Beneficiary**
+
+- *(TBD)* Add **SetBeneficiaryExpiration** method to miner actor to propose or connfirm the expiration of a beneficiary. When a beneficiary expired, the owner could take it back directly. There are 2 steps for this setting. 0 setting means no expiration setting
+    - The benneficiary/owner propose an expiration height
+    - The owner/beneficiary confirm the expiration height
```
anorth, line 47: I think an expiration date should be mandatory. Far future dates are possible. We can force at least some consideration about the time horizon of incentives. It can always be extended by mutual agreement (possibly contractual, etc).

FIPS/fip-0028.md
steven004: This is included

FIPS/fip-0028.md
steven004: Yes. Put this into the implementation section.

FIPS/fip-0028.md
steven004: Basically I think the staking is another layer above the Filecoin basic design. 
Do you have some specific ideas about this?

FIPS/fip-0028.md
steven004: Yes, agreed. 
I just rephrased a little. Please feel free to directly propose changes in the doc. thanks.

FIPS/fip-0028.md
steven004: Just removed all the TBD. Welcome more comments.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,125 @@
...
+spec-sections: v6.0.x
+---
+
+
+<!--You can leave these HTML comments in your merged FIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new FIPs. Note that a FIP number will be assigned by an editor. When opening a pull request to submit your FIP, please use an abbreviated title in the filename, `fip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+To provide more flexibilities for Filecoin finance market, a new role *beneficiary* is proposed for a storage provider, which takes over the financial control from the owner.
```
jennijuju, line 18:
```suggestion
Add a new *beneficiary* keypair to associate with the miner actor, it can be add or removed by the owner keypair. The beneficiary address can **only** be used for withdrawing miner available balance.
```

FIPS/fip-0028.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+
+There are different roles for a storage provide node for differennt specific purpose, including a worker and optional multiple controllers for mining/market and PoRep/PoSt message handling. At the same time, a node is owned by the role owner which has full control of the node, including finance.
```
jennijuju, line 23:
```suggestion
There are several keypairs associated with a given storage provider address for different operational purpose, including a worker and zero or more controllers block producing, making deals, onboarding sectors and proving storage. There is also the owner keyapir that is used for a SP administration and fund withdraws, to be more specific, currently **only the owner address can withdraw the funds from the miner actor.
```

FIPS/fip-0028.md
```diff
@@ -0,0 +1,125 @@
...
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+To provide more flexibilities for Filecoin finance market, a new role *beneficiary* is proposed for a storage provider, which takes over the financial control from the owner.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+
+There are different roles for a storage provide node for differennt specific purpose, including a worker and optional multiple controllers for mining/market and PoRep/PoSt message handling. At the same time, a node is owned by the role owner which has full control of the node, including finance. 
+
+The beneficiary role is to take financial control from owner, i.e. only financiary role could withdraw filecoin from a storage provider node, and the owner keep other controls.
```
jennijuju, line 25:
```suggestion
A SP may use the owner key pair to add one beneficiary accounts, in which those accounts can solely be used to withdraw the funds from the miner actor. A beneficiary account can be a secp256k1, BLS or a multisig account.
```

FIPS/fip-0028.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+
+Filecoin's pledge collateral mechanism makes it a perfect lending market. We already see some projects launched to solve the problem, e.g. Coinlist is running lending business to Filecoin storage providers while asking excess pledge paid in other digital currency. However, the filecoin node itself has great value including pledge collateral and locked rewards, which, ideally, can be used as collateral for filecoin lending. DARMA project is trying to do this but asking the ownership transferred, which is also practical in some Filecoin service providers.
```
jennijuju, line 30: >DARMA project is trying to do this 

Does this has public source?

FIPS/fip-0028.md
```diff
@@ -0,0 +1,125 @@
+fip: "0028"
+title: Beneficiary address for storage providers
+author: Steven Li(@steven004), Jiaying Wang(@jennijuju)
```
jennijuju, line 3:
```suggestion
author: Steven Li(@steven004)
```

FIPS/fip-0028.md
jennijuju, line 30: This reads to me as "Filecoin is a perfect lending market", which is a bit weird imho, wondering what other editors thinks. @arajasek @kaitlin-beegle maybe?

jennijuju: Should we just keep it simple? just to say its for unlocking more flexibility / opportunity in SP financial assets management?

jennijuju:
(haven't done a full pass yet - will 👁️ more later this week.

FIPS/fip-0028.md
anorth: We must acknowledge that a provider who assigns all future earnings to some other party, possibly by accident, possibly forever, has a dramatically reduced incentive to maintain storage power and deals. It's not a blocker, but deserves thinking about and possibly mitigating. cc @zixuanzh @nicola

FIPS/fip-0028.md
anorth: I think concrete example use cases are very helpful as motivation here. Otherwise a reasonable default stance would be to reject risky change.

FIPS/fip-0028.md
steven004: My original concept is really to add the beneficiary role, which could share the same address with owner, but they are different roles. That is, there will be always beneficiary address, just as the miner always has worker address, even worker can share the same address with owner too. 

The implementation will reflect the concept. With this in mind, there is no way to remove a beneficiary.

FIPS/fip-0028.md
steven004: See my comments above.

FIPS/fip-0028.md
kaitlin-beegle: I agree, but I think this might just be a language/cultural discrepancy :) 

I think "creates opportunities for a lending market" is a bit more appropriate, yes?

FIPS/fip-0028.md
```diff
@@ -1,4 +1,4 @@
-fip: "0028"
+fip: "to be assigned"
```
kaitlin-beegle, line 1: +1 on this...if you need to set a file name in order to open the PR, just add the name of the FIP.  We assign FIP numbers only when FIPs enter last call status, so that they are numbered in accordance with which they are accepted and implemented on-chain.  

Since FIPs are worked on and reviewed for varying lengths of time, adding a FIP number before 'last call' status adds confusion.  I've added this to my 'documentation needs' list, since it could be clearer!

FIPS/fip-0028.md
steven004: Mark this as solved due to no more comments on this.

FIPS/fip-0028.md
steven004: Thanks, changed.

FIPS/fip-0028.md
steven004: Added some considerations in this section. please review. @anorth

FIPS/fip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+- The **Beneficiary** address could be a signable address, or multisig address, or actor address (for smart contract)
+- The **WithdrawBalance** method of a miner can be called by the miner's **Beneficiary** or **Owner**, but the balance always being sent to the beneficiary. The total balance withdrawed to a beneficiary address could not exceed the quota if it is set.
+
+- Add **ChangeBeneficiary** method to miner actor to propose or confirm a change of beneficiary address and/or beneficiaryInfo. When changing beneficiaryInfo, the exprireDate needs to be the future, and the quota should be larger than the usedQuota. The beneficiaryChange is following a proposal/confirmation manner. There are different scenarios under which the proposal and confirmation steps are different, as below:   
+    1) The current beneficiary address is the owner, the ChangeBeneficiary just can be sent by the Owner 
+        - ChangeBeneficiary command needs to propose a new beneficiary address, at the same time, the beneficiaryInfo can be set at the same time
+        - The ChangeBeneficiary just need to be sent once and immediately take effect
```
anorth, line 52: This is a bit risky. It's possible for the owner to make a mistake and set an invalid address and then they have no recourse (until expiration, which may be ~indefinite).

I think this should be confirmed by the proposed beneficiary address, following the same pattern as ChangeOwnerAddress.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+There are a few parts to be implemented to achieve the beneficiary separation and related security matters, including: 1) miner-state change to accommodate benefiary and its state; 2) methods added to miner-actor to support beneficiary change; 3) Date and/or quota experition support for beneficiary. 
+
+The following channges apply to specs-actors module:
+
+#### 
+- Add **Beneficiary** address into **MinerInfo** structure
+- Add **BeneficiaryInfo** structure into **MinerInfo** structure to include `{beneficialQuota (attoFIL), beneficialExpireDate(Height), usedQuota(attoFIL)}`.
```
anorth, line 44: I think that `beneficialQuota` and `usedQuota` can be replaced by a single number, `remainingQuota`.

- When installing a new beneficiary, set `remainingQuota` to the same value you would use for `beneficialQuota`
- During withdraw subtract from `remainingQuota` instead of adding to `usedQuota`
- To change quota, set `remainingQuota` to the desired value for `beneficialQuota - usedQuota`

I might be missing something that prompted you to the total-plus-used representation. It does have the advantage that the total quota for the beneficiary is persistent on chain, so easily discoverable. If so, let's spell the rationale out in the proposal. You can easily convince me, I just wanted to point out this simpler representation.

Note that a potential "race" exists in either case where, if a withdrawal is processed between the sending of and execution of a message increasing the quota, the miner will effectively get a larger increase than what might have been intended based on the `beneficialQuota - usedQuota` when the increase was sent. An alternative design would be to submit a quota _delta_ when changing quota for an existing beneficiary. We could probably fit this in with the same message structure, just carefully name the quota field as `QuotaDelta`, with the interpretation that it's a delta from zero if the beneficiary address is new.

I could be convinced that either approach to the "race" is reasonable, too. But let's spell it out in the design rationale.

FIPS/fip-0028.md
```diff
@@ -0,0 +1,130 @@
...
+        - When there is a new beneficiary (not the owner itself), after the owner sent the ChangeBeneficiary command to change the beneficiary and beneficiaryInfo, the new beneficiary should send the same command to confirm, and then it can take effect. In addition, the usedQuota is set to 0 since it's new. 
+    3) The beneficiary (not the owner) is not expired and quota is not used up, the ChangeBeneficiary command is also firstly sent by the owner, and next confirmed by the beneficiary(ies) to take effect.
+        - When the beneficiary address does not change, i.e. only change beneficiaryInfo, thus, only need the beneficiary to send the same command to confirm (like what we do for owner change)
+        - When there is a new beneficiary address is proposed. It requires the current beneficiary to send the same command to confirm, and followed the new beneficiary confirmation command the the same parameters. In this case, the usedQuota is set to 0, and the newly set quota is counted from 0.
```
anorth, line 59: This is eliding some details about how this three-phase commit state is represented. E.g. I think we'll need at least a `ProposedBeneficiaryInfo` field and something to record who has approved it.

However that may be fine to note just briefly and leave to implementation. I think the behaviour you've specified constrains it enough.

anorth (commented):
Thanks. I like this proposal, my comments are relatively minor. The only high importance one is confirming the beneficiary address to prevent the accidental loss of rewards.

I'll be pleased to approve this after that's addressed.

FIPS/fip-0028.md
anorth: Thanks 👍

FIPS/fip-0028.md
steven004: This is a good catch, I'm adding `ProposedBeneficiaryInfo` into the implementation section. 
The detail of the structure could be referred to the implementation reference.

FIPS/fip-0028.md
steven004: Very good point!
I would think it is better to keep `beneficialQuota` and `usedQuota`. There are two reasons: 
1. As you mentioned, it provided a clear view of how much of total quota for a beneficiary address, and how much used, very easy to track;
2. It's about the *race* you mentioned. I would think the *race* condition is harder to handle if we only have `remainingQuota` since the beneficiary could really withdraw before confirmation, which makes the beneficiary has more quota than expected perhaps. However, when we keep both, the quota change applies to the `beneficialQuota`, and the withdraw value is added to usedQuota, so the withdraw could not impact the totalQuota setting for the beneficiary. 

There might be a question when decreasing the beneficialQuota, the beneficiary can withdraw before the confirmation, in this case, the `ChangeBeneficiary` will fail, which is expected. But, this `ChangeBeneficiary` only happens under the owner and beneficiary's agreement. This scenario only means the beneficiary does not agree the change. 

@anorth , have more thoughts? If you agree, I will put this into the implementation rationale section.

steven004: The QuotaDelta is a way to solve the *race condition*, but it will still have user to calculate by themselves. The two parameters design looks more user-friendly, while costing a little more storage.

FIPS/fip-0028.md
steven004: Yes, I missed this, I think. 
The design should be as you mentioned.

FIPS/fip-0028.md
anorth: Sounds good.

Yes, I see the race disappears if you think about the operation as "increase the total quota to X". I was thinking "increase the quota from what they have already used by X", which semantics have a race with the two-variable representation. But I agree your way is better, and "increase total quota" is probably the way to think about it anyway.

anorth (approved):
Great! I think this is good to land as draft status.

hunjixin:
https://github.com/filecoin-project/specs-actors/pull/1571

anorth:
@kaitlin-beegle and FIP editors, can we please review and merge this PR?

@arajasek and @magik6k this proposal would really benefit from your review too (doesn't have to block this PR, you can make comments in #253)

anorth:
I realised something else we should have here: a `GetBeneficiary` method to read the beneficiary address, quota, and expiration from the miner. I'm thinking forward to FVM smart contracts here: a contract beneficiary will need to be able to on-chain confirm that it is indeed the beneficiary and for how long.

steven004:
I think I have a question here is whether a contract needs to call an
actor's method to get its status, or states can be gotten via system calls?
e.g. we have a state structure standard (like JSON format as an example),
any actor's state can be read via a syscall if it follows the standard.

anorth:
No, actors cannot read each other's state. They must call methods (so an actor's internal state is abstracted). It's not very explicit in Go because we control all the actors, but will be enforced in the FVM.

anorth:
We should add `GetProposedBeneficiary` too so that a contract or UI can read the proposal in order to submit matching configuration to confirm it.

PR #267, opened 2022-01-13
Owner key change by random PoSt challenge proposal

lionsoul2014:
The Filecoin network will be more secure and consistent with the original intention of data storage if we provide a mechanism that the owner key could be changed by random PoSt challenge, also this change will bring better engagement and investment experience.


lionsoul2014:
the CC sectors could be regenerated once the ticket for PC1 is get or extracted and that is a problem.

lionsoul2014:
move to https://github.com/filecoin-project/FIPs/discussions/269

PR #272, opened 2022-01-25
Update README.md

kaitlin-beegle:


anorth:
I love the spirit here, but could we please merge the linked PRs and then link to the document created, rather than link to the PRs? 

Links to the document will remain current if the document changes through subsequent review, but links to PRs will have to be constantly updated. I also like that wanting to reference something is a good trigger to force the FIP PR process to happen. The present situation signals a problem to me that we should resolve at its root, rather than work around.

PR #280, opened 2022-01-31
First draft of fungible token standard (following ERC-20).

anorth:
This proposal provides a standard API for the implementation of fungible tokens within smart contracts. 
This proposal is an adaptation of Ethereum's [ERC-20](https://eips.ethereum.org/EIPS/eip-20) for the Filecoin VM.

Discussion topics:
- #277 for discussion of the _content_ of this proposal
- #279 for meta discussion about the _form_ of this proposal: how we should propose token standards in general

cc @raulk @Stebalien

### DIFF
--- a/FIPS/fip-tokenstandard.md
+++ b/FIPS/fip-tokenstandard.md
@@ -1,0 +1,237 @@
+---
+fip: <to be assigned>
+title: Fungible token standard
+author: Alex North (@anorth)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/277
+status: Draft
+type: Technical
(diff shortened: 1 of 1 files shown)


anorth:
After sitting with this idea for a while and learning more about similar standards both in Ethereum and other ecosystems, I no longer think that following ERC-20 this closely is a good idea. We have an opportunity to take the lessons from years of experience subsequent to the adoption of the early Ethereum standards and apply them to a relatively clean slate.

I'm going to withdraw this draft, but continue discussion toward better standards. FYI @nicola @Kubuxu @raulk.

PR #281, opened 2022-01-31
Create resources directory and add SnapDeals theory report for FIP-0019.

anorth:
Implementing my proposal in #275: A directory tree `/resources/FIP-xxx/...`. FIP documents may then link to resources with relative URLs.

Note: the link as rendered in the "rich diff" view does not work, but I believe it will after this PR is merged.


Kubuxu (approved):

kaitlin-beegle (changes requested):
Can we update the folder name to just "Resources" rather than resources/fip0019? 

I'm currently working with cryptoeconlab to create better audits/auditing standards for FIPs, and this would be a natural home for those documents.  I think we ought to be able to add subfolders for each FIP that has supporting materials.

anorth:
@kaitlin-beegle yes my change does add a hierarchy of a `resources` folder and a `fip-xxxx` for each FIP (that has any resources) under that. I think that's what you asked for, but might be confusing in the GitHub UI.

If you care about the capitalisation of `Resources`, I can change that.

kaitlin-beegle:
> @kaitlin-beegle yes my change does add a hierarchy of a `resources` folder and a `fip-xxxx` for each FIP (that has any resources) under that. I think that's what you asked for, but might be confusing in the GitHub UI.
> 
> If you care about the capitalisation of `Resources`, I can change that.

Yep, I'm guessing you used a CLI to push the folder to the repo? Not a problem; let's merge as is and we can update it if it's not what we need. 

And thanks but the capitalization doesn't matter to me :)

kaitlin-beegle (approved):

PR #294, opened 2022-02-08
FIP-0031: Atomic switch to non-programmable FVM

raulk:
A few days ago, we submitted [FIP-0030 (Introducing the Filecoin Virtual Machine)](https://github.com/filecoin-project/FIPs/pull/288). That FIP served as a baseline specification for the FVM, and did not propose concrete network actions.

This FIP that proposes the activation of a non-programmable version of the FVM in the Filecoin network.

In terms of deployment, this FIP should be coupled with an upcoming FIP that proposes a rearchitecture of the gas model to prepare for arbitrary code.


FIPS/fip-atomic_fvm_switch.md
```diff
@@ -0,0 +1,265 @@
+---
+fip: <to be assigned>
```
kaitlin-beegle, line 2: FIP0031

FIPS/fip-atomic_fvm_switch.md
```diff
@@ -0,0 +1,265 @@
...
+
+However, the code and deployment form of built-in actors will now be Wasm
+bytecode. Actor code will begin to be identified through real
+content-addressing. The current _synthetic_ actor CodeCIDs will migrate to CIDs
+that hash over their actual Wasm bytecode.
+
+## Change Motivation
+
+Swapping out core components in the execution layer of a blockchain is a complex
+operation. It's analogous to replacing the engine of an airplane while airborne.
```
anorth, line 81: The overly dramatic analogy is not helpful here.

FIPS/fip-atomic_fvm_switch.md
```diff
@@ -0,0 +1,265 @@
...
+
+For this reason, its advisable to manage risk carefully. Monolithic, big-bang
+deployments are risky. It's appropriate to break up changes in smaller units
+that can be deployed in an orderly, incremental fashion, eventually leading up
+to the final goal: on-chain user-programmability.
```
anorth, line 86: Style nit: prefer first-person. "We advise...."

FIPS/fip-atomic_fvm_switch.md
```diff
@@ -0,0 +1,265 @@
...
+| `fil/7/storagepower`              | TBD                                                   |
+| `fil/7/storageminer`              | TBD                                                   |
+| `fil/7/storagemarket`             | TBD                                                   |
+| `fil/7/paymentchannel`            | TBD                                                   |
+| `fil/7/multisig`                  | TBD                                                   |
+| `fil/7/reward`                    | TBD                                                   |
+| `fil/7/verifiedregistry`          | TBD                                                   |
+
+> Notes:
+> What would a premigration look like here?
```
anorth, line 165: It's not that big, probably unnecessary.

FIPS/fip-atomic_fvm_switch.md
```diff
@@ -0,0 +1,265 @@
...
+The `CreateMiner` method (method number 2) calls the Init actor to construct a
+new Miner actor. It will need to pass the new content-addressed CodeCID for the
+latter.
+
+## Design Rationale
+
+WIP.
+
+## Backwards Compatibility
+
```
anorth, line 207: Note whether or not a node running the intrinsic actors is expected to be able to maintain sync after this epoch, after simple changes like changing the code CIDs.

anorth (commented):
Looking good so far.

anorth:
A couple of questions to answer:
- What will be the impact on block validation times? What levers to we have to pull if that impact is non-zero?
- How well prepared will we be to quickly fix a critical error in the built-in actors, should one be discovered after the switch? How would the process and timelines change? (There is a subtext here about the thoroughness of testing that would enable us to be confident deploying a small change direct to the network with very little testnet time).

Stebalien:
> What will be the impact on block validation times? What levers to we have to pull if that impact is non-zero?

We're not sure yet. Unptimized, I'm seeing my lotus/FVM node go from 1.5-3s block validation times, to 3-6s. Luckily:

1. We can be benchmark before we actually upgrade.
2. We have room for optimization.

- IPLD access can be significantly optimized: https://github.com/filecoin-project/ref-fvm/issues/332
- We can explore alternative wasm VMs like wasmer (one of our goals is to be wasm VM agnostic).
- We can explore actor and syscall optimizations (https://github.com/filecoin-project/ref-fvm/issues/265).

> How well prepared will we be to quickly fix a critical error in the built-in actors, should one be discovered after the switch? How would the process and timelines change? (There is a subtext here about the thoroughness of testing that would enable us to be confident deploying a small change direct to the network with very little testnet time).

I'm planning on running a mock upgrade on a test network to force us to get everything in-place. We have 3 upgrade points:

1. We can replace some or all of the actor implementations. The FVM itself won't actually care about this as it'll just execute the actors as-specified by the chain state.
2. We can swap out the FVM Kernel, Machine, etc. for different network versions (or add conditionals). Although I'd prefer to avoid when possible.
3. We can import multiple versions of the entire FVM (like we do with specs-actors) and construct the correct version for the current epoch.

I'd like to test 1 and 3 (an upgrade that swaps out both the actors and the FVM itself) in a test network before calling the upgrade "ready".

> There is a subtext here about the thoroughness of testing that would enable us to be confident deploying a small change direct to the network with very little testnet time

Our current best source of test coverage is our test vectors. Once we get some test coverage information (see https://github.com/filecoin-project/ref-fvm/issues/314) we can determine how dire the situation really is.

I'd also like to "foreward-port" our test vectors to newer network versions by re-running them while ignoring anything related to gas.

But I'm actually less concerned about test _coverage_ than I am about test _frameworks_. We need to be able to write tests _quickly_ if we want to be able to fix bugs quickly.

anorth:
Thanks @Stebalien. By asking the question I really meant that the FIP text should answer it. I get that the validation cost question may be "we don't know" at present, but, if you want to rush the FIP, we could still outline the policy options.

FIPS/fip-0031.md
```diff
@@ -0,0 +1,396 @@
+---
+fip: "0031"
+title: Atomic switch to non-programmable FVM
+authors: Raúl Kripalani (@raulk), Steven Allen (@stebalien), Jiaying Wang (@jennijuju)
```
jennijuju, line 4:
```suggestion
authors: Raúl Kripalani (@raulk), Steven Allen (@stebalien)
```

jennijuju:
Hi https://github.com/orgs/filecoin-project/teams/fips-editors - from a FIP DRAFT completion perspective, i think its good to merge? authors may address the comments in follow up PRs and work towards last call?

anorth:
Fine to merge as a draft. I don't think it can move to last call with the `WIP`s in place for some important sections.

raulk (commented):
That works for me. In that case, I'd appreciate an approval so I can merge this PR as a draft FIP.

jennijuju (approved):

PR #302, opened 2022-02-17
update fip number

jennijuju:
without the "" FIP id is taken as octal value , which suggests the fip number is 24 instead of 30


anorth:
Lolwat? How much more antagonistic to humans does Markdown wanna be?

anorth (approved):

PR #305, opened 2022-02-18
Supporting contracts as deal clients in the built-in storage market actor - first draft

anorth:
This is a formal proposal for the idea proposed in https://github.com/filecoin-project/FIPs/discussions/271

I would recommend we strive to include this before the FVM enables user-programmable contracts, even if other built-in actor changes supporting that are pushed back.

FYI @nicola @zixuanzh @Kubuxu @Stebalien @ZenGround0

### DIFF
--- a/FIPS/fip-xxxx-contract-deal-clients.md
+++ b/FIPS/fip-xxxx-contract-deal-clients.md
@@ -1,0 +1,225 @@
+---
+fip: <to be assigned>
+title: Support actors as storage market clients
+author: Alex North (@anorth)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/271
+status: Draft
+type: Technical
+category: Core
+created: 2022-02-17
+---
+
+## Simple Summary
+Extend the storage market actor to support deal proposal by clients without the need to submit a signed message.
+This allows actors (which have no signing key) to act as deal clients.
+
+## Abstract
+The built-in storage market actor does not support other actors as deal clients.
+When the FVM enables user-programmable actors, this would be a big limitation on their functionality.
+This limitation arises from the market actor's requirement for a client signature on a deal proposal,
+but only externally-owned accounts (EOAs) can make signatures.
+
+This proposal adds new methods to the storage market actor for a client to propose a deal on-chain,
+and then a provider to confirm that deal. This complements the existing one-step publishing of storage
+deals that is suitable for EOAs.
+
+## Change Motivation
+Interacting with storage markets is likely to be a significant use case for FVM actors.
+Examples of contracts that might act as deal clients include actors representing a group (multisig), 
+more complex DAOs, decentralised replication/repair services, 
+or brokering actors such as auction or bounty contracts built on top of the built-in market.
+
+Without making a change like this, storage deals will remain restricted to external account owners, 
+which is a significant barrier to automation and trust-minimised systems.
+
+## Specification
+Add new state and methods to the built-in storage market actor.
+
+The existing code already uses the term "proposals" to describe the collection of deals
+that both parties have already committed to,
+and "pending" to describe any deal that has not yet completed.
+To avoid confusion, the on-chain proposals are termed "offers".
+
+### State
+```go
+type State struct {
+    // Existing state remains as-is.
+    
+    // Offers by clients for specific deal terms, not yet accepted by the provider.
+    // TODO: consider structuring as HAMT[Address]AMT[abi.DealID]DealProposal,
+    // indexing first by client address and then ID.
+    // This could provide cheaper access if/when the collection becomes very large.
+    Offers cid.Cid // AMT[abi.DealID]DealProposal
+}
+```
+
+Offers are identified by integers in the same namespace as other deals, so that the identifier
+remains constant across the phases of offer and acceptance by the storage provider.
+
+### Methods
+
+```go
+ type OfferStorageDealsParams struct {
+ 	Deals []DealProposal
+ }
+type OfferStorageDealsReturn struct {
+	IDs []abi.DealID // Parallel array of IDs for the proposals.
+}
+
+// Offers a set of storage deals from a client to specific providers.
+// Each deal proposal must specify the client as the calling actor, but may specify different providers.
+// Client collateral and payment are locked immediately; the call fails if 
+// insufficient client funds are available for all offers.
+// Fails if any of the proposals are internally invalid or expired.
+// Fails if an identical proposal already exists as a pending proposal (submitted here or via PublishStorageDeals).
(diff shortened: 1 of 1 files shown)


FIPS/fip-xxxx-contract-deal-clients.md
```diff
@@ -0,0 +1,225 @@
...
+### State
+```go
+type State struct {
+    // Existing state remains as-is.
+    
+    // Offers by clients for specific deal terms, not yet accepted by the provider.
+    // TODO: consider structuring as HAMT[Address]AMT[abi.DealID]DealProposal,
+    // indexing first by client address and then ID.
+    // This could provide cheaper access if/when the collection becomes very large.
+    Offers cid.Cid // AMT[abi.DealID]DealProposal
```
jennijuju, line 52: The Lotus team wanna strongly propose that the "provider" can be an optional field in the offer as  a part of this proposal rather than future improvement/consideration.

FIPS/fip-xxxx-contract-deal-clients.md
```diff
@@ -0,0 +1,225 @@
...
+}
+type AcceptStorageDealsReturn struct {
+    Accepted []abi.DealID
+    Failed []abi.DealID
+}
+
+// Accepts a set of storage deals by a provider, committing both parties.
+// Each ID must reference a proposal currently offered and specifying as the provider a 
+// miner actor for which the caller is the owner, worker, or control address.
+// All offers must specify the same provider.
```
jennijuju, line 96: Does this have to be a must?  if a SP runs multiple SPs with the same owner/worker/control keys, they might want to accept storage deals on via different providers!

FIPS/fip-xxxx-contract-deal-clients.md
```diff
@@ -0,0 +1,225 @@
...
+We will likely see this kind of pattern repeated in other actors until Filecoin offers
+a first-class account abstraction primitive
+(see discussion in [Ethereum](https://docs.ethhub.io/ethereum-roadmap/ethereum-2.0/account-abstraction/)).
+
+### Partial failure
+The client's call to offer deals always succeeds completely or fails. 
+It never succeeds at making only some offers.
+This route is chosen to provide the simplest UX to client contracts,
+which can rely on the atomic operation for their own logic.
+It is unfortunate that there is no mechanism to provide data to the caller about which offers failed, or why,
```
jennijuju, line 129: why cant we just return the successful ones in `OfferStorageDealsReturn minus failure ones' IDs?

FIPS/fip-xxxx-contract-deal-clients.md
```diff
@@ -0,0 +1,225 @@
...
+
+Non-contract deal clients can use these new methods too, if they wish. 
+As compared with `PublishStorageDeals`, the `OfferStorageDeals` flow:
+- involves the client to posting a message on chain, rather than sending it to the provider out of band;
+- uses slightly more total gas, split between the client and the provider;
+- is cheaper in gas for providers;
+- uses significantly _less_ total gas if the client is a BLS address.
+
+In both flows, an offer cannot be retracted by a client, and thus may be relied upon by the provider.
+
```
jennijuju, line 218: in this session we should add that it adds a cost (as in gas) for client

FIPS/fip-xxxx-contract-deal-clients.md
anorth: We'll have to think about it a bit. The `DealProposal` type is used off-chain for negotiation, so changing it is high friction (it will need to change later anyway to accomodate #298). We could possibly hack in a sentinel "empty" address rather than make it actually nullable. Or maybe we can use a slightly different type here, and migrate other deals to match it later (tricky because we used CID to check uniqueness).

Please make the suggestion in the discussion issue.

FIPS/fip-xxxx-contract-deal-clients.md
anorth: I'd prefer simplicity. This follows the way PublishStorageDeals works. I can see how your idea would work, but I'm not trying here to make the built-in market as ergonomic as possible, just have _basic_ functionality. I hope that future non-built-in market contracts will add all the nice features people want.

FIPS/fip-xxxx-contract-deal-clients.md
anorth: It's a design decision not to support partial success. This sentence is about the fact that when failing (via abort), there is no way to return information.

If you mean to question why don't we choose to support partial success, I've forgotten the details, but please ask in the discussion issue. It's about client (contract) ergonomics, so that they don't have to deal with partial success. We might be able to change it, but we need to understand how client contracts will want to invoke this.

FIPS/fip-xxxx-contract-deal-clients.md
jennijuju: this is fair and we agree its good to have the simplicity- closing!

FIPS/fip-xxxx-contract-deal-clients.md
anorth: Moved to https://github.com/filecoin-project/FIPs/discussions/271#discussioncomment-2851880

anorth:
I plan to propagate some suggestions from the discussion #271 here and then I think this'll be ready to merge as a FIP in draft stage.

FIPS/fip-xxxx-contract-deal-clients.md
anorth: I've changed the proposal now to offer atomic operation or partial success, at client's request.

FIPS/fip-xxxx-contract-deal-clients.md
```diff
@@ -0,0 +1,201 @@
+---
+fip: <to be assigned>
```
kaitlin-beegle, line 2: FIP0035

FIPS/fip-xxxx-contract-deal-clients.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Abstract
+The built-in storage market actor does not support other actors as deal clients.
+When the FVM enables user-programmable actors, this would be a big limitation on their functionality.
+This limitation arises from the market actor's requirement for a client signature on a deal proposal,
+but only externally-owned accounts (EOAs) can make signatures.
+
+This proposal adds new methods to the storage market actor for a client to propose a deal on-chain,
+and then a provider to confirm that deal. This complements the existing one-step publishing of storage
+deals that is suitable for EOAs.
```
kaitlin-beegle, line 25: Small nit, could we please spell out what 'EOA' stands for?

kaitlin-beegle (approved):

PR #307, opened 2022-02-20
Set pre-commit deposit independent of sector content - first draft

anorth:
The sector pre-commit deposit (PCD) is currently set equal to an estimate of the _storage pledge_ for a sector.
The storage pledge is a 20-day projection of the expected reward to be earned by the sector.
This reward depends on power, which depends on the presence of verified deals hosted by the sector.
This dependency means the miner actor must consult the market actor, which intermediates verified data cap,
to determine the pre-commit deposit (and then again to compute initial pledge when the sector is proven).

This proposal sets the pre-commit deposit to a value independent of sector content:
the estimated storage pledge of a sector with space-time that is exactly filled by verified deals.

Discussion at https://github.com/filecoin-project/FIPs/discussions/290

FYI @zixuanzh @nicola

### DIFF
--- a/FIPS/fip-xxxx-precommit-deposit.md
+++ b/FIPS/fip-xxxx-precommit-deposit.md
@@ -1,7 +1,7 @@
 ---
 fip: <to be assigned>
 title: Fix pre-commit deposit independent of sector content
-author: Alex North (@anorth)
+author: Alex North (@anorth), Jakub Sztandera (@Kubuxu)
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/290
 status: Draft
 type: Technical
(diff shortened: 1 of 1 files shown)


zixuanzh:
Thanks @anorth this proposal generally looks good and considerations have been described in discussions. We might want to double-check how PCD is used/refunded when a sector is ProveCommitted during implementation in case of bugs.

jennijuju:
Notes from lotus team 

from implementation consideration:
- atm deals are validated in ProveCommit stage, in which we should consider to move to precommit stage to prevent sufficient PCD lost.
- We may revisit https://github.com/filecoin-project/specs-actors/issues/1470 for prove commit as now there will be more PCD to lose.

anorth:
Updated:
- Changed SectorPreCommitOnChainInfo schema to record CommD, matching the schema eventually required when SPs specify their own CommD rather than compute from a privileged market actor
- Restore the deal validity checks during pre-commit
- Removed now-redundant loading of deal info for CommD calculation at ProveCommit, resulting in net gas saving
- Noted the restoration of security buffers in the PCD calculation in light of SnapDeals

nicola:
Proposal looks good to me.

Increasing PCD shouldn't be an issue, since the capital is locked only for ~140 blocks.

ZenGround0 (approved):

jennijuju (approved):

FIPS/fip-xxxx-precommit-deposit.md
```diff
@@ -0,0 +1,166 @@
+---
+fip: <to be assigned>
```
kaitlin-beegle, line 2: FIP0034! I'll merge into repo, and update number for you, @anorth

kaitlin-beegle (approved):

PR #315, opened 2022-03-08
FIP-0032: Gas model adjustment for non-programmable FVM

raulk:
This FIP is the first in a two-FIP series that adjust the gas model to prepare it for the FVM.
- `FIP-0032` (this FIP): Gas model adjustment for non-programmable FVM. Introduces execution gas, rearchitects IPLD state management fees, adjusts the actor call fee.
	- FIP-0032 should activate with nv16 (Skyr upgrade -- non-programmable FVM).
- `FIP-nnnn` (upcoming FIP): Gas model adjustment for user-programmability. Introduces the actor installation fee, introduces the actor loading fee, introduces memory expansion gas, prices unpriced syscalls, and more.
    - FIP-nnnn should activate with n18 (enabling user-programmability).

### DIFF
--- a/FIPS/fip-0032.md
+++ b/FIPS/fip-0032.md
@@ -1,0 +1,260 @@
+---
+fip: "0032"
+title: Gas model adjustment for non-programmable FVM
+author: Raúl Kripalani (@raulkr), Steven Allen (@stebalien)
+discussions-to: https://github.com/filecoin-project/FIPs/issues/204
+status: Draft
+type: Technical
+category: Core
+created: 2022-03-08
+spec-sections:
+  - TBD
+requires:
+  - FIP-0030
+  - FIP-0031
+replaces: N/A
+---
+
+# FIP-0032: Gas model adjustment for non-programmable FVM
+
+## Simple Summary
+
+This FIP is part of a two-FIP series that adjust the gas model of the Filecoin
+blockchain in preparation of user-programmable actors. This FIP focuses on
+attaining high-fidelity gas accounting under the Wasm runtime that built-in
+system actors will migrate to as part of FIP-0030 and FIP-0031. It does not
+cover upcoming system features like actor deployment, actor loading, memory
+management, and more. Those will be addressed in a subsequent FIP that focuses
+on gas model changes to support user-programmability.
+
+## Abstract
+
+With the arrival of the Filecoin Virtual Machine, the Filecoin network will gain
+the capability to run untrusted code in the form of user-defined actors.
+However, the current gas model is designed strictly for predefined, trusted
+code. In order to maintain network security guarantees, the gas model needs to
+evolve to accurately charge for computational instructions and resource usage of
+actor logic.
+
+This FIP expands the current gas model by introducing a new gas category to
+account for the real compute cost of the logic performed by every actor in the
+system. It also rearchitects IPLD state management gas, and adjusts the actor
+call fee to account for Wasm invocation container mechanics.
+
+This is the first FIP out of a two-FIP series to revise the gas model as a
+result of the introduction of the Filecoin Virtual Machine in FIP-0030 and
+FIP-0031. This FIP should accompany those FIPs under the same network upgrade.
+
+## Change Motivation
+
+Gas is an essential mechanism to (a) compensate the network for the computation
+performed during transaction execution, and to (b) restrict the compute capacity
+of blocks to achieve a designated epoch time (30 seconds).
+
+Today, Filecoin determines the amount of gas a transaction consumes by
+intercepting specific operations and actions, and applying gas explicit gas
+charges on them. Concretely:
+
+- On including a chain message, variable on the length of the message.
+- On returning a value, variable on the length of returned payload.
+- On performing a call, dependent on the type of call (simple value transfer, or
+  other).
+- On performing an IPLD store get, fixed amount.
+- On performing an IPLD store put, variable on block payload size.
+- On actor creation, fixed amount.
+- On actor deletion, fixed amount.
+- On cryptographic operations:
+    - On signature verification, variable on signature type and plaintext
+      length.
+    - On hashing, fixed price.
+    - On computing the CID of an unsealed sector, variable on proof type and
+      piece characteristics.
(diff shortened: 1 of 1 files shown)


kaitlin-beegle:
:tada: :100: :heart: 

@raulk, do you intend for the second FIP (FIPnnnn) to be published and implemented with FIPs 30-32, or is is it intended to support the forthcoming M2 work? In other words, are we awaiting a fourth FIP for acceptance prior to the v16 calibnet sync?

raulk:
@kaitlin-beegle an early draft of the second FIP is coming later today! But that FIP will stay open for longer and is entirely independent from the nv16 timeline.

Fun fact: I had drafted both FIPs as one, but later realised that we had too many undefined items in the M2-related items that would hinder the approval of the M1-related items (nv16). So I decided to split that FIP up into these two. This structure aligns much more nicely with how we're scheduling FVM work.

kaitlin-beegle:
> @kaitlin-beegle an early draft of the second FIP is coming later today! But that FIP will stay open for longer and is entirely independent from the nv16 timeline. Also, its parameters are dependent on benchmarking code that hasn't been written yet.
> 
> Fun fact: I had drafted both FIPs as one, but later realised that we had too many undefined items in the M2-related items that would hinder the approval of the M1-related items (nv16). So I decided to split that FIP up into these two. This structure aligns much more nicely with how we're scheduling FVM work.

Appreciate the quick reply, and fast work! It was a great call to separate these two documents so that we can move more quickly on the most time-sensitive part. 

We'll call for reviews at this week's Core Devs, so we can see how many outstanding questions, etc., exist after an initial review. Assuming feedback is mostly minimal, I'm hoping we can schedule FIPs 30 - 32 for Last Call beginning on March 21.  However, we have some flexibility, should anyone need a longer review time. 

@raulk do you plan to attend this week's Core Devs and talk through the specifics of some of the gas modeling? Not a requirement, but want to make sure you have enough time if you do intend to join.

FIPS/fip-0032.md
```diff
@@ -0,0 +1,260 @@
+---
+fip: "0032"
+title: Gas model adjustment for non-programmable FVM
+author: Raúl Kripalani (@raulkr), Steven Allen (@stebalien)
+discussions-to: https://github.com/filecoin-project/FIPs/issues/204
```
anorth, line 5: This is not the right link. Could you please open a discussion thread?

FIPS/fip-0032.md
```diff
@@ -0,0 +1,260 @@
...
+
+To simplify community discussion and align with the FVM deployment roadmap, all
+changes related to new system features related to user-programmable actors are
+kept out of scope of this FIP. This set of changes covers only the technological
+transition to a Wasm runtime for built-in actors. This FIP will be followed by
+another FIP introducing the former.
+
+Concerning the design of each individual change, the specification section
+elaborates on overhead that is being accounted for and the variables that make
+up every pricing formula.
```
anorth, line 203: Is your plan that the pricing formulae be constructed so that the existing syscall charges remain fixed, and the other items have proportional values so as to maintain constant relative cost per call? Otherwise, it may be necessary to reprice all the existing syscalls.

FIPS/fip-0032.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Incentive Considerations
+
+With the introduction of execution gas to account for real execution cost, actor
+compute logic will no longer be exempted from gas charges, thus resulting in a
+net increase in gas expenditure in the network.
+
+## Product Considerations
+
+No product considerations apply.
```
anorth, line 234: The relative costs of various actor methods are going to change, which is product-relevant whenever the base fee is non-trivial. Once you have the data, we should spell those out here.

anorth (commented):
Looking good. Let's leave it open a few days for other editors, but I think we could merge this draft. 

Needs security considerations, concrete numbers, and references to the analysis backing up those numbers before last-call.

raulk:
@kaitlin-beegle 

> @raulk do you plan to attend this week's Core Devs and talk through the specifics of some of the gas modeling? Not a requirement, but want to make sure you have enough time if you do intend to join.

Thanks for the invitation. Yes, I will join and present this FIP in the call.

FIPS/fip-0032.md
raulk: Good point.

FIPS/fip-0032.md
raulk: This FIP plans to keep the current syscall prices unchanged for simplicity. As stated in the Change Motivation section, _execution gas_ is additive to the current syscall prices because the current system does not charge for the cost of actor logic because it doesn't have a way to do so. In a future FIP (likely #317), we can adjust the syscall price parameters. But the overarching goal is always _execution cost fidelity_ (i.e. what each syscall _actually_ costs), not the maintenance of legacy prices.

FIPS/fip-0032.md
raulk: Actually, I'm not sure I follow your comment about the base fee being non-trivial. There is nothing changing here about how the base fee itself is calculated, which basically depends on total gas used in tipsets AFAIR.

anorth: There are two bits here:
1. People will only care about relative gas changes when the base fee is not zero, because otherwise everything costs $0.
2. This proposal will increase total gas usage, which runs a good chance of increasing base fee if we don't also increase the block gas limit (but we probably want to do the opposite to give ourselves buffer in validation times if we don't get the gas costs right first go)

Kubuxu: Agreed with Anoth here, this has a possibility of significantly affecting the ratios of GasUsed of different operations.
We also have the gas balancer in place for the aggregated prove commits, if gas changes significantly it will need to be changed as well.

kaitlin-beegle:
Merging FIP draft, per @anorth's comment above.  Current FIP draft will be circulated in this week's community governance update.

FIPS/fip-0032.md
zixuanzh: Generally agreed but the principle for calculating GasUsed is based on the resources consumed by the messages/operations. As long as we are applying the same principle it should probably be fine.

PR #329, opened 2022-03-23
define manifest data type for FIP 0031

vyzo:


FIPS/fip-0031.md
```diff
@@ -142,8 +142,8 @@ struct Manifest {
 }
 ```
 
-The _Manifest_ data for version 1 (the only version currently defined) is a Cid linking
-to an IPLD `Vector<(String, Cid)>`.
+The _Manifest_ data for version 1 (the only version currently defined) is a CID linking
+to an IPLD `Vector<(String, CID)>`.
```
raulk, line 146: We should formally define the order of this vector.

raulk: Here and, most importantly, in the system actor's state.

vyzo: So whats the order? By type? Lexicographic?
I dont think it really matters all that much, but i am not opposed ti specifying it either.

vyzo: So the bundler uses a BTreeMap, which orders by actor type -- this is perfectly reasonably I guess, should we make this the standard order?

vyzo: Updated to define the order by enum position.

PR #335, opened 2022-04-01
Update fip-0032.md

kaitlin-beegle:
Proposed updated language (in the 'Abstract') to clarify the relationship between various FVM FIPs over time.


raulk (approved):
LGTM; just noting that this FIP is formatted with 80-char width, but these lines aren't.

kaitlin-beegle:
> 

Apologies, probably because I proposed changes through the Github UI rather than a console text editor :) will fix and merge.

PR #375, opened 2022-05-19
Create fip-0035.md

peter198:


peter198:
Optimize parallel PreCommit1 tasks according to hardware configuration
* Quantity can improve overall task completion speed.

anorth:
This is not a Filecoin protocol change, but a feature request for an implementation. Please re-submit it to https://github.com/filecoin-project/lotus

PR #399, opened 2022-07-07
Initial draft of FRC for calling convention

anorth:
An actor calling convention that derives method numbers by hashing method names.

This proposal describes a convention for the labelling of methods across actors which will support the development of standardised APIs, e.g. a token API. The convention maps human-friendly method names to method numbers by hashing the method name. This supports the definition of multiple standardized APIs with minimal potential for collision on method numbers.

See https://github.com/filecoin-project/FIPs/discussions/382 for discussion.
PTAL @Kubuxu @raulk @Stebalien

### DIFF
--- a/FRCs/frc-xxxx-calling-convention.md
+++ b/FRCs/frc-xxxx-calling-convention.md
@@ -1,0 +1,139 @@
+---
+fip: <to be assigned>
+title: Calling convention with hashed method name
+author: @anorth, @kubuxu
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/382
+status: Draft
+type: FRC
+created: 2022-07-07
+---
+
+## Simple Summary
+An actor calling convention that derives method numbers by hashing method names.
+
+## Abstract
+This proposal describes a convention for the labelling of methods across actors which will 
+support the development of standardised APIs, e.g. a token API.
+The convention maps human-friendly method names to method numbers by hashing the method name.
+This supports the definition of multiple standardized APIs with minimal potential for collision on method numbers.
+
+## Change Motivation
+Filecoin messages include a method number and parameter block as the essential payload for method invocation.
+The FVM implements internal dispatch: there is a single entry point to each actor for receiving messages.
+The FVM does not further constrain dispatch, and the development team wish for the FVM to remain neutral with respect to such conventions.
+
+The built-in actors' methods are numbered sequentially, partly in order to enjoy a dense representation for top-level chain messages.
+This scheme is inappropriate for development of standardized APIs to be implemented by many actors, 
+as different APIs would need to coordinate to avoid colliding on different meanings for the same method number.
+Method _names_ are a more developer-friendly handle on a method, and much more suited to avoiding collisions.
+
+This proposal continues to use method numbers, but provides a deterministic mapping from method names to those numbers.
+
+Goals:
+- A familiar dispatch-by-name abstraction for programmers
+- Compact, uniform-size on-chain encoding
+- No change to blockchain message schema
+- Permit specification of standard actor interfaces, including retroactively in recognition of widely-adopted patterns
+- Permit extending the interface of a deployed actor to implement methods defined in a new standard
+- Independent of programming language
+
+This proposal assumes that method names are defined statically, 
+i.e. that actors will have the interface definitions of any other actors they call available at compile-time.
+However, it can support dynamic resolution for future reflection capabilities.
+
+## Specification
+### Method number computation
+The method number for a symbolic method name is calculated as the first four bytes of `hash(method-name)`,
+interpreted as an unsigned, big-endian 32-bit integer.
+
+The `hash` function is given as:
+- if `method-name == "Constructor` then `1`, else
+- if `blake2b(method-name) == 0` then `4294967295` (`0xffffffff`), else
+- `blake2b(method-name)`
+
+### Method name conventions
+A method is *exported* when a method number is computed for it and the actor will recognise this number for internal dispatch. 
+Conventions on exported method names are independent of any programming language conventions. 
+Note that these conventions only apply to methods exported to the VM for inter-actor dispatch. 
+Actor developers can continue to use relevant programming language conventions for simple internal function calls.
+
+Exported method names *should*:
+- Use only the ASCII characters in `[a-zA-Z0-9_]` (the same set as the C programming language). Other characters, including unicode beyond this set, are excluded in order to reduce the opportunity for misleading spelling of names in user interfaces.
+- Have an initial capital letter, and use CamelCase to identify word boundaries.
+- Capitalize all letters in acronyms.
+
+Examples: `Approve`, `SubmitWindowPoSt`, `MintNFT`, `FRC1234AuthorizeOperator`
+
+## Design Rationale
+### Method number computation
+The blake2b hash function is already implemented in the VM and available as a syscall.
+This permits easy dynamic method number calculation, 
+though build tools are expected to compute most method numbers statically at compile time.
+
+Method number `1` is expected as the constructor function by the built-in `Init` actor, 
+and user-deployed actors must implement it. This proposal assigns a conventional name.
+
+Method number `0` is treated by the FVM as a special-case bare send of the native Filecoin token, 
+and will not invoke an actor's code. 
+Rather than have some subset of method names be inaccessible, this proposal maps them to a different method number.
+
+Both the blockchain `Message` structure and FVM support up to 64-bits for a method number.
+This proposal restricts values to 32-bits in order to compress the representation for top-level message method numbers
+to four bytes (a minor gas saving on every message).
+
+### Collisions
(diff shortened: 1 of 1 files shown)


Tom-OriginStorage:
Coming as an end user (someone who is planning to write tons of contracts on FVM with Rust), I am against this FRC. FVM is WASM based, not EVM based. Therefore, it is more conventional to follow the standard of other WASM-based VMs rather than EVM.

This is a version that we have adopted internally which closely follows the convention of Cosmwasm. Over here, the method number simply represents the type of function (constructor, write, read-only), we can also easily generate a JSON schema using the `JsonSchema` derive macro which is the equivalent of an ABI in Ethereum.

```
#[no_mangle]
pub fn invoke(params: u32) -> u32 {
    // Conduct method dispatch. Handle input parameters and return data.
    let ret: Option<RawBytes> = match sdk::message::method_number() {
        1 => constructor(),
        2 => execute(params),
        3 => query(params),
        _ => abort!(USR_UNHANDLED_MESSAGE, "unrecognized method"),
    };

    // Insert the return data block if necessary, and return the correct
    // block ID.
    match ret {
        None => sdk::message::NO_DATA_BLOCK_ID,
        Some(v) => match sdk::ipld::put_block(DAG_CBOR, v.bytes()) {
            Ok(id) => id,
            Err(err) => abort!(USR_SERIALIZATION, "failed to store return value: {}", err),
        },
    }
}

pub fn query(params: u32) -> Option<RawBytes> {
    let params = sdk::message::params_raw(params).unwrap().1;
    let params = RawBytes::new(params);
    match params.deserialize().unwrap() {
        QueryMsg::BalanceOf { owner } => balance_of(owner),
        QueryMsg::OwnerOf { token_id } => owner_of(token_id),
        QueryMsg::GetApproved { token_id } => get_approved(token_id),
        QueryMsg::IsApprovalForAll { owner, operator } => is_approved_for_all(owner, operator),
    }
}
```

Example on how the calling convention looks like when interacting with Lotus:
`lotus chain invoke t01001 2 oWpiYWxhbmNlX29moWVvd25lcmVhbGljZQ==`
whereby `oWpiYWxhbmNlX29moWVvd25lcmVhbGljZQ==` is the cbor-base64 encoding of `{"balance_of": {"owner": "alice"}}`. If lotus automatically does cbor-base64 encoding for us in the future, we can directly pass in the json.

We believe this addresses all of the goals:
✔️ Provide a familiar dispatch-by-name abstraction to programmers
✔️ Compact, uniform-size on-chain encoding
✔️ No change to blockchain message schema
✔️ Permit specification of standard actor interfaces, including retro-actively in recognition of widely-adopted patterns
✔️ Permit extending the interface of an extant actor to implement methods defined in a new standard
✔️ Independent of programming language

We are also ready to release a fully functional sample actor for the community, just waiting for a stable release of FVM sdk so that the community can play around with it on a localnet/testnet.

anorth:
@Tom-OriginStorage thank you for responding. Could you please re-post your comment in #382 for discussion?

FRCs/frc-xxxx-calling-convention.md
```diff
@@ -0,0 +1,139 @@
...
+However, it can support dynamic resolution for future reflection capabilities.
+
+## Specification
+### Method number computation
+The method number for a symbolic method name is calculated as the first four bytes of `hash(method-name)`,
+interpreted as an unsigned, big-endian 32-bit integer.
+
+The `hash` function is given as:
+- if `method-name == "Constructor` then `1`, else
+- if `blake2b(method-name) == 0` then `4294967295` (`0xffffffff`), else
```
Kubuxu, line 51: As the method ID calculations will be done off-chain, we can use rejection sampling instead of plain substitution for outcomes `0` or `1`.
```rust
let mut hash Vec<u8>= blake2b(method-name)
while !hash.is_empty() {
  let method_id = u32::from_le(hash[0..4])
  if method_id != 0 && method_id != 1 {
    return method_id
  }
}
// probability of 2^{-248}, aka it won't happen
panic!("Method ID could not be determined, please change it")
```
This way we don't have additional collisions at `0xffffffff` or 1.

FRCs/frc-xxxx-calling-convention.md
```diff
@@ -0,0 +1,139 @@
...
+- `blake2b(method-name)`
```
Kubuxu, line 52: I would also suggest domain separation for hashing, just in case we want to change something in future.

`blake2b(method-name)` => ``blake2b("1|"+method-name)` - `1|` as `|` is illegal in the method name

FRCs/frc-xxxx-calling-convention.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Method name conventions
+A method is *exported* when a method number is computed for it and the actor will recognise this number for internal dispatch. 
+Conventions on exported method names are independent of any programming language conventions. 
+Note that these conventions only apply to methods exported to the VM for inter-actor dispatch. 
+Actor developers can continue to use relevant programming language conventions for simple internal function calls.
+
+Exported method names *should*:
+- Use only the ASCII characters in `[a-zA-Z0-9_]` (the same set as the C programming language). Other characters, including unicode beyond this set, are excluded in order to reduce the opportunity for misleading spelling of names in user interfaces.
```
Kubuxu, line 61: Do we allow starting with `[0-9]`?.C and many other languages don't allow it.

FRCs/frc-xxxx-calling-convention.md
anorth: It won't _always_ necessarily be done off-chain, but still thanks for the suggestion.

FRCs/frc-xxxx-calling-convention.md
anorth: Yes ok I'll add this restriction too.

anorth: I'm torn on disallowing underscore, since by construction these are for exported names. But I'll leave it allowed.

FRCs/frc-xxxx-calling-convention.md
```diff
@@ -0,0 +1,154 @@
...
+The `hash` function is given as:
+```
+let domain_separation_tag = "1|" // Note `|` is illegal in method names
+let mut digest Vec<u8>= blake2b(domain_separation_tag + method_name)
+while !digest.is_empty() {
+  let method_id = u32::from_be(digest[0..4])
+  if method_id != 0 && method_id != 1 {
+    return method_id
+  }
+  digest.remove(0) // pop-front
```
Kubuxu, line 58:
```suggestion
  digest = digest[4..] // pop-front
```

FRCs/frc-xxxx-calling-convention.md
```diff
@@ -0,0 +1,154 @@
...
+### Method number computation
+The method number for a symbolic method name is calculated as the first four bytes of `hash(method-name)`,
+interpreted as an unsigned, big-endian 32-bit integer.
+
+The `hash` function is given as:
+```
+let domain_separation_tag = "1|" // Note `|` is illegal in method names
+let mut digest Vec<u8>= blake2b(domain_separation_tag + method_name)
+while !digest.is_empty() {
+  let method_id = u32::from_be(digest[0..4])
```
Kubuxu, line 54: I'm assuming we are using big-endian to be consistent with existing protocols.

anorth: It's "network" byte order and also the order used by CBOR.

jennijuju:
The draft lgtm - 'd like to assign this FRC as `0042`, and will be happy to approve it once the assigned number is getting updated in proposal & filename.

kaitlin-beegle:
> 

Agreed! Thanks for reviewing. 

Not to state the obvious, but I think we should recognize a policy wherein any FIP editor who authors a FIP/FRC cannot also be the primary reviewer for their work. Glad you were able to take a look at this :)

kaitlin-beegle (approved):

jennijuju:
@kaitlin-beegle i think we will have to revert the PR, or create a new pr as FIP number was not updated  in the proposal yet

FRCs/frc-xxxx-calling-convention.md
```diff
@@ -0,0 +1,154 @@
...
+before significant adoption makes renaming one a significant burden.
+
+### Conceptual compatibility with Ethereum
+This scheme is inspired by the Solidity ABI for the Ethereum VM.
+It differs in the choice of hash function and exclusion of method parameter types (no overloading),
+so is not transparently compatible.
+
+Note that removing both those differences would *not* automatically make this scheme compatible with the Solidity ABI, 
+since method parameter types have fundamentally different schemas between the two environments. 
+A future FVM/EVM embedding will need to translate method selectors along with the rest of the EVM ABI.
```
raulk, line 116: I doubt we will have logic to translate method selectors; that would rely on a number of assumptions that I don't feel good encoding in the protocol (even if in actor land). Instead, Ethereum CALLs would set a fixed dynamic method number and rely on the call convention in parameters to perform dispatch.

PR #405, opened 2022-07-12
FIP-0034: Note changes to ProveReplicaUpdates

anorth:
The first draft omitted consideration of `ProveReplicaUpdates`, which needs an update to handle the new return values from market methods.

Note that `ProveReplicaUpdates` could be further optimised with a dedicated both-at-once call in the built-in market actor. But this would become irrelevant after future planned changes to the miner-market interactions.

FYI @Kubuxu


jennijuju:
could you please link the builtin actor PR in the #Implementation session? (currently, it only has a link to the go actor

jennijuju:
requested review from the co-author @Kubuxu, once ✅ will approve & merge

Kubuxu (approved):

jennijuju (approved):

PR #423, opened 2022-08-02
FIP-0043 Introduce Domain Separation FIP

geoff-vball:
First draft of a FIP for adding Signature Domain Separation Tags, see https://github.com/filecoin-project/FIPs/discussions/387.

An open question is how to deal with BLS Signature Aggregation. Feedback welcome!

### DIFF
--- a/FIPS/fip-0043.md
+++ b/FIPS/fip-0043.md
@@ -1,0 +1,128 @@
+---
+fip: FIP-0043
+title: Signature Domain Separation
+author: Geoff Stuart (@geoff-vball), Aayush Rajesekaran (@arajasek)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/387
+status: Draft
+type: Technical
+category: Core
+created: 2022-08-02 
+specs-sections: 4.5.1 Signatures
+---
+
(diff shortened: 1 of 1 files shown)


anorth:
I think this draft is not quite in a clear enough state to merge. The specification for protocol level changes is not clear, and there are quite a few open questions. It's been sitting here for a while. I'd suggest we close this PR unless it will be actively worked on in the near future. The FIP number can stay reserved for when someone does pick it up, though.

Kubuxu:
It would be great if [FIP-0044](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0044.md) included DST as well.

Kubuxu:
After thinking about it more, I would upgrade the "great" to "strong should".
FIP-0044 is a standard we want other people to adopt, the lack of DST will cause long-term security issues and the transition to DST enabled version of FIP-0044 will be hard.

cc @Stebalien

kaitlin-beegle:
Closing for now. Seems like the benefits of this FRC have been incorporated as part of follow-on FIPs, and that this draft is not currently moving towards readiness. Feel free to reopen if that changes!

PR #432, opened 2022-08-14
FIP-0045 First draft of FIP for decoupling FIL+  from markets

anorth:
This is a first draft of the FIP for the proposal discussed in https://github.com/filecoin-project/FIPs/discussions/313

### DIFF
--- a/FIPS/fip-0045.md
+++ b/FIPS/fip-0045.md
@@ -3,31 +3,82 @@ fip: 0045
 title: De-couple verified registry from markets
 author: Alex North (@anorth)
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/313
-status: Draft type: Technical category : Core created: 2022-08-12 requires: FIP-0034
+status: Draft
+type: Technical
+category : Core
+created: 2022-08-12
+requires: FIP-0034
 ---
 
 TODO:
-- Method for extending a sector, SP provides allocation ids (because market doesn't retain them after deal expiration)
-- Confirm policy for built-in market actor's default term maximum
-- Confirm policy for sector migration any extension to term maximum beyond the sector's lifespan
+- Spec receiver hook on Account, Multisig actors
 - Spec methods for sector migration on miner, market, and verifreg actors
+- Spec change to SectorOnChainInfo to distinguish migrated sectors
+- Discuss changes to deal packing
 - Discuss how market delegate means no external API changes
 - Discuss mechanisms for conditional allocations, negatively priced deals
+- Confirm policy for built-in market actor's default term maximum
+- Confirm policy for sector migration any extension to term maximum beyond the sector's lifespan
 - Market method for fetching allocation ids for deals (needed?)
 - Spec hook/method for a different client extending the claim by spending data cap (needed?).
 
 
 ## Simple Summary
-<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
-TODO
+Represents Filecoin Plus data cap allocations separately from the built-in market actor.
+Adds explicit term (duration) limits to data cap allocations.
+Simplifies the quality-adjusted power calculation to remove spreading out over a sector's lifetime.
+Supports extending the terms of data cap allocations independently of market deals.
+Represents un-allocated data cap as a fungible token.
+These changes provide a foundation for user-programmed market actors to broker Filecoin Plus verified deals in the future.
 
 ## Abstract
-<!--A short (~200 word) description of the technical issue being addressed.-->
-TODO
+The current implementation of Filecoin Plus verified deals is deeply coupled to the built-in market actor,
+which is treated as trusted code.
+The built-in market is inefficient, limited in functionality and can only improve very slowly through network upgrades.
+The architectural coupling prevents any user-programmed actors from replicating, improving, or varying deal-related functionality,
+especially with any relation to Filecoin Plus.
+
+This proposal decouples the representation of Filecoin Plus data cap allocations from the built-in market actor's deals, 
+and the process of claiming it from the built-in market's processes.
+The built-in market actor remains a delegate, acting on behalf of verified clients,
+but only in a way that will be accessible to user-programmed market actors too.
+This removes some of the necessary trust vested in the built-in market actor's code. 
+
+Along the way, this proposal implements the ability to extend data cap terms beyond the market's deal terms.
+It also simplifies the quality adjusted power calculation so as to support
+the future ability to update the data in a sector regardless of its previous content.
+This proposal gets half way toward full support for user-programmed market actors.
 
 ## Change Motivation
-<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
-TODO
+The core motivation behind this proposal is to enable the development of markets and market-like actors
+on the FVM, once user-programability is supported.
+All negotiation between data clients and storage providers is currently limited by 
+the built-in market actor's limited abilities,
+and improvements to it are thus limited to FIPs and network upgrades.
+Features like deal extension, multi-sector deals, re-negotiation, deal transfer, capacity deals,
+auction markets, repair markets, insurance, derivatives etc. should all be possible in user-programmed actors.
+None should require a network-wide upgrade to realise.
+
+The built-in market actor currently occupies an exclusive position as the sole broker of
+Filecoin Plus data cap, tightly coupled with the verified registry and storage miner actors.
+In order to enable the development of useful storage-related applications and features,
+we need to break the tight coupling between the built-in actors, expose composable on-chain primitives,
+and make the capabilities of the built-in market actor available to user-programmable contracts.
+This proposal addresses the capability to broker of Filecoin Plus data cap.
+
+Breaking coupling between the built-in market actor, verified registry actor, and miner actors 
+will eventually remove the storage market from consensus-critical risk surface,
+and move subsequent market development into the realm of smart-contract upgrades,
+rather than network-wide consensus protocol upgrades.
+Minimising coupling is a key principle in software and systems engineering. 
+Decoupling the verified data rewards from storage deals provides great opportunities for 
+simplifying Filecoin’s core mechanics in order to provide a better platform for development.
+
+One benefit we can immediately realise is simplifying the calculation of quality-adjusted power. 
+The current mechanism, which is coupled to deal terms, imposes significant limits on the flexibility of sector storage,
+essentially rendering it write-once. 
+Decoupling these neatly resolves the calculation difficulties to give full flexibility, 
+and is much simpler than the Filecoin Plus premium proposal that also attempted this.
 
 ## Specification
 **Terminology**
@@ -172,15 +223,19 @@ struct SectorAllocationClaim {
     // The (padded) size of the data committed (up to the allocation's data size).
     PieceSize: uint64
     // The sector into which the piece has been committed.
-    SectorID: uint64,
+    SectorID: uint64
+    // The epoch during which the sector is scheduled to expire.
+    SectorExpiration: int64
  
 }
 struct ClaimAllocationsParams {
     Sectors: []SectorAllocationClaim
 }
 struct SectorClaimResult {
-    // Indices of allocation claims in the parameters that succeeded.
-    Successful: []int64
+    // Count of successful claims.
+    SuccessCount: uint64
+    // Index and failure code for each unsuccessful claim.
+    Failures: []{index: uint64, code: ExitCode}
 }
 // Called by storage miner actor to claim allocations for data provably committed to storage.
 // For each allocation claim, the registry checks that the provided piece CID
@@ -221,8 +276,10 @@ struct ExtendClaimTermsParams{
     Claims: []ClaimTerm
 }
 struct ExtendClaimTermsResult {
-    // Indices of extensions that succeeded.
-    Successful: []int64
+    // Count of successful extensions.
+    SuccessCount: uint64
+    // Index and failure code for each unsuccessful extension.
+    Failures: []{index: uint64, code: ExitCode}
 }
 
 // Called by allocation client to extend the maximum term of their allocations, 
@@ -236,27 +293,27 @@ fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
 The `Clients` map of addreses to balances is removed from the verified registry state.
 The balances of data cap tokens held by verified clients are extracted to a data cap token actor.
 This actor represents un-allocated data cap as a fungible token.
+This token actor is intended to implement the FRC-0046 token standard in the future,
+but specification of the full, public API is deferred to a subsequent proposal.
+The data cap token actor should be retroactively updated to support the standard later.
+
 Token balances are represented as BigIntegers with 18 decimal places of precision, 
 a representation which is likely to be adopted by future token standards.
 Token quantities are constrained to be whole multiples of `10^18`. 
 One whole data cap token corresponds to one unit of data cap to be allocated,
 which today is good for one byte of data.
-
 The data cap token is generally non-transferable, except to or from the verified registry actor.
 The token actor enforces these transfer restrictions.
 The verified registry actor can invoke privileged methods to create and destroy tokens.
 
-A token standard is expected to support delegation of control of token balances to specified
-third-party actors.
-The built-in storage market actor is pre-authorized as a delegate for all verified clients.
+FRC-0046 supports delegation of control of token balances to specified third-party actors as "operators".
+The built-in storage market actor is pre-authorized as an operator for all verified clients.
 Thus, the built-in market actor can allocate data cap on clients' behalf.
-This mechanism allows this proposal to be impelemented without changing any clients' or
-providers' workflows.
-It also will support user-programmed market actors dealing in data cap in the future.
-The data cap token actor is intended to present a public interface conforming to standards for
-fungible token behaviour and an extensible calling convention.
-Neither of those standards are pre-requisite for this proposal, however.
-The built-in actors will be retroactively updated to support those standards when they settle.
+This delegation mechanism will support user-programmed market actors dealing in data cap in the future.
+
+This token delegation mechanism allows this proposal to be implemented without changing any
+clients' or providers' workflows for existing functionality.
+Parties will continue to interact with the built-in market actor and verified registry actor directly.
 
 #### Operations
 The `AddVerifiedClient` method is changed to invoke the data cap token actor to
@@ -274,7 +331,7 @@ data cap tokens back to the client's balance for re-use.
 
 ### Simplified quality-adjusted power
 A data cap allocation/claim is valid for a range of commitment terms, between a client-specified
-minimum and maximum.
+minimum and maximum, so long as the allocation is claimed before the allocation expires.
 
 The actual term for a claim begins at the epoch the data is committed into a sector,
 proven either with proof-of-replication or replica-update. 
@@ -347,7 +404,43 @@ This facility to drop a claim can only be exercised:
 - in the final 30 days of the sector's currently committed lifespan
 
 These restrictions prevent a provider using extension as a means of escaping their commitment to 
-store the claimed piece for the sector’s full originally-committed lifetime. 
+store the claimed piece for the sector’s full originally-committed lifetime.
+
+The storage provider operator must provide the IDs of the allocations claimed by the sector when extending.
+
+```
+// The verified data claims currently supported by a sector, to be maintained or dropped
+// while processing extension.
+// All claims currently supported by the sector must be provided.
+struct SectorClaims {
+    Sector: uint64,
+    MaintainClaims: []ClaimID
+    DropClaims: []ClaimID
+}
+struct ExpirationExtension2 {
+    Deadline: uint64,
+    Partition: uint64,
+    Sectors: BitField, // IDs of non-FIL+ sectors (must have zero verified deal weight)
+    SectorsWithClaims: []SectorClaims
+    NewExpiration: int64, // epoch
+}
+struct ExtendSectorExpirationParams2 {
+    Extensions: []ExpirationExtension2,
+}
+struct ExtendSectorExpirationResult2 {
+    // Count of successful extensions.
+    SuccessCount: uint64
+    // Index and failure code for each unsuccessful extension.
+    Failures: []{index: uint64, code: ExitCode}
+}
+
+// Increases the expiration epoch for some sectors, optionally dropping expired claims.
+// For each sector, the provided claim IDs must correspond to claims that refer back to the 
+// sector, and the sum of their size must equal the sector's verified deal weight.
+// The new expiration must be prior to the earliest term maximum for any maintained claims.
+fn ExtendSectorExpiration2(params: ExtendSectorExpirationParams2) -> ExtendSectorExpirationResult2
+
+```
 
 ### Built-in storage market as a delegate
 The workflows described above are independent of any market actor, 
@@ -377,16 +470,24 @@ let dealAllocation = Allocation {
   }
 ```
 
-The market stores the allocation ID in deal state.
+The market stores allocation IDs for not-yet-activated deals in a new mapping, 
+and the claim ID for activated deals in deal state.
 
 ```
+struct State {
+  // Remainder unchanged...
+  
+  // Allocation IDs for deals that are not yet activated.
+  PendingDealAllocationIDs CID // HAMT[DealID]AllocationID
+}
+
 struct DealState {
   // Existing state
   SectorStartEpoch: int64,
   LastUpdatedEpoch: int64,
   SlashEpoch: int64,
   // New
-  AllocationID: uint64
+  ClaimID: uint64
 }
 ```
 
@@ -405,17 +506,22 @@ TODO!
 The power for existing sectors is not changed by this proposal.
 
 A provider may opt-in to migrate a sector to the new QA power calculation method.
+Each active deal is converted to a claim as if an allocation had been created when the deal was made.
+The claim's minimum term is the deal's duration, and maximum term is the sector's current expiration epoch
+(TODO: plus some buffer?).
+
+Expired deals cannot be migrated, even though the old QA power calculation method has spread out
+their power and reward over the sector's full lifetime.
+This will generally result in an increase in power for a sector, 
+but may result in a decrease in case of expired deals.
+This is why the migration is opt-in.
+
 A new method on the storage miner actor:
 - Fetches information about each not-yet-expired verified deal from the market
 - Calls the verified registry to create a claim for each such deal.
-  The claim's minimum term is the deal's duration, and maximum term is the sector's current expiration epoch
-  (TODO: plus some buffer?)
 - Re-calculates power according to the amount of space committed to active claims. 
-
-This will generally result in an increase in power for the sector.
-Note that power may be lost for deals that have already expired,
-but which weight was previously amortized over the sector lifespan.
-  
+- Re-caclulates pledge requirement according to the new power and network conditions (but never decreases) 
+ 
 After migration, the client can extend the migrated claims' terms up to the network policy maximum.
 After claims are extended, the provider can extend the sector commitment duration to match.
 Note that a migrated sector cannot be extended unless the claim terms are extended, 
--- a/FIPS/fip-0045.md
+++ b/FIPS/fip-0045.md
@@ -14,16 +14,16 @@ TODO:
 - Spec methods for sector migration on miner, market, and verifreg actors
 - Spec change to SectorOnChainInfo to distinguish migrated sectors
 - Confirm policy for built-in market actor's default term maximum
-- Market method for fetching allocation ids for deals (needed?)
-- Spec hook/method for a different client extending the claim by spending data cap (needed?).
+- Market method for fetching allocation/claim ids for deals (needed?)
+- Spec hook/method for a different client extending the claim by spending DataCap (needed?).
 
 
 ## Simple Summary
-Represents Filecoin Plus data cap allocations separately from the built-in market actor.
-Adds explicit term (duration) limits to data cap allocations.
-Simplifies the quality-adjusted power calculation to remove spreading out over a sector's lifetime.
-Supports extending the terms of data cap allocations independently of market deals.
-Represents un-allocated data cap as a fungible token.
+Represents [Filecoin Plus](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0003.md) DataCap allocations separately from the built-in market actor.
+Adds explicit term (duration) limits to DataCap allocations.
+Simplifies the quality-adjusted power calculation to remove power spreading out over a sector's lifetime.
+Supports extending the terms of DataCap allocations independently of market storage deals.
+Represents DataCap as a fungible token.
 These changes provide a foundation for user-programmed market actors to broker Filecoin Plus verified deals in the future.
 
 ## Abstract
@@ -33,13 +33,13 @@ The built-in market is inefficient, limited in functionality and can only improv
 The architectural coupling prevents any user-programmed actors from replicating, improving, or varying deal-related functionality,
 especially with any relation to Filecoin Plus.
 
-This proposal decouples the representation of Filecoin Plus data cap allocations from the built-in market actor's deals, 
+This proposal decouples the representation of Filecoin Plus DataCap allocations from the built-in market actor's deals, 
 and the process of claiming it from the built-in market's processes.
 The built-in market actor remains a delegate, acting on behalf of verified clients,
 but only in a way that will be accessible to user-programmed market actors too.
 This removes some of the necessary trust vested in the built-in market actor's code. 
 
-Along the way, this proposal implements the ability to extend data cap terms beyond the market's deal terms.
+Along the way, this proposal implements the ability to extend DataCap terms beyond the market's deal terms.
 It also simplifies the quality adjusted power calculation so as to support
 the future ability to update the data in a sector regardless of its previous content.
 This proposal gets half way toward full support for user-programmed market actors.
@@ -55,11 +55,11 @@ auction markets, repair markets, insurance, derivatives etc. should all be possi
 None should require a network-wide upgrade to realise.
 
 The built-in market actor currently occupies an exclusive position as the sole broker of
-Filecoin Plus data cap, tightly coupled with the verified registry and storage miner actors.
+Filecoin Plus DataCap, tightly coupled with the verified registry and storage miner actors.
 In order to enable the development of useful storage-related applications and features,
 we need to break the tight coupling between the built-in actors, expose composable on-chain primitives,
 and make the capabilities of the built-in market actor available to user-programmable contracts.
-This proposal addresses the capability to broker of Filecoin Plus data cap.
+This proposal addresses the capability to broker of Filecoin Plus DataCap.
 
 Breaking coupling between the built-in market actor, verified registry actor, and miner actors 
 will eventually remove the storage market from consensus-critical risk surface,
@@ -77,19 +77,19 @@ and is much simpler than the Filecoin Plus premium proposal that also attempted 
 
 ## Specification
 **Terminology**
-- *Allocation*: data cap allocated by a client to a specific piece of data.
+- *Allocation*: DataCap allocated by a client to a specific piece of data.
 - *Claim*: a provider's assertion they are storing all or part of an allocation
-- *Term*: period of time for which a data cap allocation or claim is valid or active.
+- *Term*: period of time for which a DataCap allocation or claim is valid or active.
 
 ### Data cap allocations
 
-The Filecoin Plus verified registry actor is extended to record allocations of data cap to specific
+The Filecoin Plus verified registry actor is extended to record allocations of DataCap to specific
 pieces of data to be stored by a provider. An allocation is independent of any deal with the
 built-in storage market (or any future market).
 
 ```
 struct Allocation {
-    // The verified client which allocated the data cap.
+    // The verified client which allocated the DataCap.
     Client: Address
     // The provider (miner actor) which may claim the allocation.
     Provider: Address
@@ -139,41 +139,45 @@ struct State {
 
 #### Operations
 
-An allocation is created by a verified client by transferring data cap tokens (see below)
+An allocation is created by a verified client by transferring DataCap tokens (see below)
 to the verified registry. 
 The `Allocation` metadata must be provided as transfer metadata.
 The verified registry checks the same preconditions as the existing `UseBytes` method. 
 The size of the allocation must match the number of tokens received.
 
-An allocation may be revoked after its expiration epoch has passed. 
-When revoked, the data cap tokens are transferred back to the client.
-The verified registry provides a new method to process revocations, replacing `RestoreBytes`.
+An allocation may be removed after its expiration epoch has passed (by anyone). 
+When removed, the DataCap tokens are transferred back to the client.
+The verified registry provides a new method to process removals, replacing `RestoreBytes`.
 
 ```
-struct RevokeExpiredAllocationsParams {
+struct RemoveExpiredAllocationsParams {
     Client: Address // client to clean up (need not be the caller)
     AllocationIDs: []AllocationID // empty for "all", or specify a set
 }
 
 // Removes all expired allocations for a client.
-// The data cap for cleaned-up allocations is returned to the client.
-// Note that un-expired allocations cannot be revoked.
-fn RevokeAllocations(params: RevokeExpiredAllocationsParams)
+// The DataCap for cleaned-up allocations is returned to the client.
+// Note that un-expired allocations cannot be removed.
+fn RemoveExpiredAllocations(params: RemoveExpiredAllocationsParams)
 ```
 
 Allocations are also removed when claimed by a provider (see below).
 
+The `RemoveVerifiedClientDataCap` method invokes a privileged method of the DataCap token actor
+to burn DataCap tokens held on a client's balance. 
+Only the verified registry actor is permitted to invoke this method.
+
 ### Data cap claims
 
 The verified registry actor is further extended to record each storage provider's commitment to a
-data cap allocation, called a claim. A claim represents a provider's obligation to store a piece of
+DataCap allocation, called a claim. A claim represents a provider's obligation to store a piece of
 data, and corresponding benefit of incentivized storage power.
 
 ```
 struct Claim {
     // The provider storing the data (from allocation).
     Provider: Address
-    // The client which allocated the data cap (from allocation).
+    // The client which allocated the DataCap (from allocation).
     Client: Address
     // Identifier of the data committed (from allocation).
     Data: CID
@@ -255,7 +259,7 @@ The maximum term for a claim may be increased by the client which originally mad
 up to the network policy maximum.
 
 A verified client can extend the term for a claim _beyond_ the initial maximum term by 
-spending new data cap. 
+spending new DataCap. 
 The term maximum can be extended up to the network policy maximum beyond the current epoch.
 The client extending the claim need not be the one that made the original allocation.
 This is similar to issuing a new allocation for the same data, but avoids overheads of re-sealing.
@@ -277,33 +281,33 @@ struct ExtendClaimTermsResult {
 }
 
 // Called by allocation client to extend the maximum term of their allocations, 
-// without consuming data cap.
+// without consuming DataCap.
 // The new TermMaximum must be no less than the current one,
 // and no greater than the maximum initial term.
 fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
 ```
 
 ### Data cap token
-The `Clients` map of addreses to balances is removed from the verified registry state.
-The balances of data cap tokens held by verified clients are extracted to a data cap token actor.
-This actor represents un-allocated data cap as a fungible token.
+The `VerifiedClients` map of addreses to balances is removed from the verified registry state.
+The balances of DataCap tokens held by verified clients are extracted to a DataCap token actor.
+This actor represents un-allocated DataCap as a fungible token.
 This token actor is intended to implement the FRC-0046 token standard in the future,
 but specification of the full, public API is deferred to a subsequent proposal.
-The data cap token actor should be retroactively updated to support the standard later.
+The DataCap token actor should be retroactively updated to support the standard later.
 
 Token balances are represented as BigIntegers with 18 decimal places of precision, 
 a representation which is likely to be adopted by future token standards.
 Token quantities are constrained to be whole multiples of `10^18`. 
-One whole data cap token corresponds to one unit of data cap to be allocated,
+One whole DataCap token corresponds to one unit of DataCap to be allocated,
 which today is good for one byte of data.
-The data cap token is generally non-transferable, except to or from the verified registry actor.
+The DataCap token is generally non-transferable, except to or from the verified registry actor.
 The token actor enforces these transfer restrictions.
 The verified registry actor can invoke privileged methods to create and destroy tokens.
 
 FRC-0046 supports delegation of control of token balances to specified third-party actors as "operators".
 The built-in storage market actor is pre-authorized as an operator for all verified clients.
-Thus, the built-in market actor can allocate data cap on clients' behalf.
-This delegation mechanism will support user-programmed market actors dealing in data cap in the future.
+Thus, the built-in market actor can allocate DataCap on clients' behalf.
+This delegation mechanism will support user-programmed market actors dealing in DataCap in the future.
 
 This token delegation mechanism allows this proposal to be implemented without changing any
 clients' or providers' workflows for existing functionality.
@@ -314,34 +318,34 @@ The built-in account and multisig actors are changed to implement a receiver hoo
 This hook always succeeds, without inspecting the payload.
 
 #### Operations
-The `AddVerifiedClient` method is changed to invoke the data cap token actor to
+The `AddVerifiedClient` method is changed to invoke the DataCap token actor to
 mint new tokens to the client's balance.
 
 The `UseBytes` method is removed.
-To create an allocation, a verified client transfers data cap tokens to the verified registry actor.
+To create an allocation, a verified client transfers DataCap tokens to the verified registry actor.
 The transfer must be accompanied by one or more `Allocation` records which specify 
 the allocation(s) to create.
 The registry may reject the transfer if the allocations are invalid.
 
 The `RestoreBytes` method is also removed.
 Instead, the `RevokeAllocations` method on the verified registry actor transfers any released
-data cap tokens back to the client's balance for re-use.
+DataCap tokens back to the client's balance for re-use.
 
 ### Simplified quality-adjusted power
-A data cap allocation/claim is valid for a range of commitment terms, between a client-specified
+A DataCap allocation/claim is valid for a range of commitment terms, between a client-specified
 minimum and maximum, so long as the allocation is claimed before the allocation expires.
 
 The actual term for a claim begins at the epoch the data is committed into a sector,
 proven either with proof-of-replication or replica-update. 
 There is no distinct "deal start epoch".
-A sector containing a data cap claim is immediately eligible for a power multiplier of 10 
+A sector containing a DataCap claim is immediately eligible for a power multiplier of 10 
 for that fraction of the sector's capacity occupied by the incentivized data
 (subject to the existing requirement to initially prove the sector in Window PoSt).
 
 The actual term for a claim ends the epoch that the sector containing the data expires, 
 or that data is replaced. 
 The term cannot end during a sector's lifetime, except by explicit replacement of data.
-If data is replaced, the sector immediately losses the power multiplier for that fraction
+If data is replaced, the sector immediately loses the power multiplier for that fraction
 of the sector's capacity.
 (Replacing non-zero data in a sector is not possible today).
 
@@ -375,8 +379,9 @@ Similarly, the miner actor requires that sector into which the piece is committe
 That is, a sector's commitment expiration epoch must fall between the minimum and maximum terms
 of any verified claim it holds.
 
-The longest practical _minimum_ term for a verified data cap allocation is
-the maximum initial sector commitment duration.
+The longest practical _minimum_ term for a verified DataCap allocation is the larger of
+the maximum initial sector commitment duration, and maximum sector extension duration
+(which today are identical).
 
 #### Failure and repair
 There is **no change to termination penalty** calculations. 
@@ -390,14 +395,16 @@ the provider may commit the same piece into a new sector and regain the quality-
 The new sector’s committed lifespan is still constrained to fit within the claim’s term.
 
 #### Sector extension
-A provider cannot extend the scheduled expiration for a sector past the maximum term 
-of any piece of verified data that it holds. 
-The provider must forgo the verified power boost when processing such a sector extension. 
-This does not reduce the sector’s pledge requirement, 
-but does reduce penalties to be paid on any future fault to reflect the new, lower power.
+A sector's scheduled commitment term cannot be extended beyond the maximum term of any
+verified claim attributed to it.
+Thus, to extend a sector, a provider must simultaneously drop any verified claims with 
+terms that would be exceeded by the extended commitment.
+Dropping a verified claim reduces the sector's power accordingly, as well as
+reducing penalties to be paid on any future fault to reflect the new, lower power.
+Dropping a claim does not reduce the sector’s pledge requirement.
 
-This facility to drop a claim can only be exercised:
-- for claims that have passed their minimum commitment,
+This facility to drop a claim while extending a sector's commitment can only be exercised:
+- for claims that have passed their minimum term,
 - for non-faulty sectors, and
 - in the final 30 days of the sector's currently committed lifespan
 
@@ -440,6 +447,15 @@ fn ExtendSectorExpiration2(params: ExtendSectorExpirationParams2) -> ExtendSecto
 
 ```
 
+#### Sector termination
+This proposal makes no changes to sector termination.
+When a sector terminates at the end of its scheduled commitment, no specific action is necessary.
+When a sector terminates before its scheduled commitment is complete, 
+the provider pays penalties according to the sector's power at that time.
+
+The verified registry actor will continue to hold `Claim` records for terminated sectors
+until the provider removes them with `RemoveExpiredClaims`.
+
 ### Built-in storage market as a delegate
 The workflows described above are independent of any market actor, 
 and take place through interactions by the client and provider with 
@@ -448,9 +464,9 @@ the verified registry actor directly.
 However, in order to maintain current client and provider workflows, 
 the built-in market actor can act as a delegate.
 
-The built-in storage market actor is pre-authorized as a data cap token delegate for all clients.
+The built-in storage market actor is pre-authorized as a DataCap token delegate for all clients.
 When a deal proposal is published with `VerifiedDeal=true`, the market actor transfers
-the corresponding data cap tokens from the client to the verified registry actor.
+the corresponding DataCap tokens from the client to the verified registry actor.
 This transfer is accompanied by an `Allocation` record populated with data from the deal.
 The built-in market specifies an allocation's minimum term to be equal to the deal duration,
 and maximum term to be some amount greater than the deal duration.
@@ -492,7 +508,7 @@ struct DealState {
 }
 ```
 
-When a verified deal expires without being committed, the data cap allocation remains.
+When a verified deal expires without being committed, the DataCap allocation remains.
 The market does not revoke the allocation during cron handling.
 Revocation is a manual operation that the client must perform.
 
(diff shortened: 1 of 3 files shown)


FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
+---
+fip: <to be assigned>
```
jennijuju, line 2: This is going to be fip 0045

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+to the verified registry. The `Allocation` metadata must be provided as transfer metadata. The
+verified registry checks the same preconditions as the existing `UseBytes` method. The size of the
+allocation must match the number of tokens received.
+
+An allocation may be revoked after its expiration epoch has passed. When revoked, the data cap
+tokens are transferred back to the client. The verified registry provides a new method to process
+revocations.
+
+```
+struct RevokeExpiredAllocationsParams {
```
ZenGround0, line 103: Even while maintaining the market actor as intermediary to verifreg we should take this opportunity to remove this call from market cron and have users manage their allocation table bloat explicitly.  In general the system should be pruning work from cron whenever it is technically feasible to do so safely and not onerous or unfair on operators.  

Safety
* SPs that do not revoke allocations can only squat on datacap they can never use
* In this case verified clients will lose out on efficient use of datacap and this is good, a VC's job is to find trustworthy SPs and it is ok for them to be punished in terms of datacap by a defecting SP that they chose
* This is not a great DOS vector for an attacking SP as each expired deal burns provider collateral.  In the current factoring of the system there is also much more significant gas cost in posting a precommit and burning PCD.

Fair for operators
* Following through on verified deals is in every SPs best interest -- no honest provider has a good reason or intention to do this
* Letting deals expire is already an action the system explicitly penalizes (by slashing provider collateral)
* The extra burden of paying for gas cost of state removal scales with the amount of garbage already in that provider's allocation table.  This means honest SPs with occasional operational problems and consistent allocation table cleanup won't be penalized much at all.

ZenGround0: Note this requires node-level work.  In lotus we should have / glob onto a configurable process that watches for deal expiries and makes the appropriate `RevokeExpiredAllocations` call.

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+data cap tokens back to the client's balance for re-use.
+
+### Simplified quality-adjusted power
+A data cap allocation/claim is valid for a range of commitment terms, between a client-specified
+minimum and maximum.
+
+The actual term for a claim begins at the epoch the data is committed into a sector,
+proven either with proof-of-replication or replica-update. 
+There is no distinct "deal start epoch".
+A sector containing a data cap claim is immediately eligible for a power multiplier of 10
```
ZenGround0, line 285: Modulo the "power on first post rule" right?  I don't think we're planning to change that for newly committed sectors.

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+TODO
+
+## Specification
+**Terminology**
+- *Allowance*: data cap provided to a verifier by the root, or to a client by a verifier.
+- *Allocation*: data cap allocated by a client to a specific piece of data.
+- *Term*: period of time for which a sector/deal/allocation is active or valid.
```
jennijuju, line 32: would it be better to have it in the form of (sector, deal, allocation) ? given these are like at tuple?

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+TODO
+
+## Incentive Considerations
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
+TODO
+
+## Product Considerations
+<!--All FIPs must contain a section that discusses the product implications/considerations relative to the proposed change. Include information that might be important for product discussion. A discussion on how the proposed change will enable better storage-related goods and services to be developed on Filecoin. FIP submissions missing the "Product Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Product Considerations discussion deemed sufficient by the reviewers.-->
```
jennijuju, line 439: product consideration should include the additional cost occurred for clients to manage verified deals.

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+	Client: Address
+    // Identifier of the data committed (from allocation).
+    Data: CID
+    // The (padded) size of data (from allocation).
+    Size: uint64
+    // The minimum period which the provider must commit to storing the piece (from allocation).
+    TermMinimum: uint64
+    // The maximum period for which the provider can earn QA-power for the piece (from allocation).
+    TermMaximum: uint64
+	// The epoch at which the (first chunk of the) piece was committed.
```
jennijuju, line 136: what do you mean by "first chuck of the piece?"

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+    // early-termination penalties (epochs).
+    TermMinimum: uint64
+    // The maximum period for which a provider can earn quality-adjusted power 
+    // for the piece (epochs).
+    TermMaximum: uint64
+    // The latest epoch by which a provider must commit data before the allocation expires.
+    Expiration: Epoch
+    // Whether the provider must commit the entire piece in a single sector.
+    // This is initially un-used, but supports allocations larger than a full sector in the future. 
+    AllOrNothing: bool
```
jennijuju, line 61: what do you mean by `initially unused`? do you mean upon upgrade, this value can only be `true`?

FIPS/fip-filplus.md
anorth: Thanks, I will remove this from cron explicitly. Note that it's verified clients, rather than providers, that lose out by not doing this though. I'll make it so anyone can call the method, though.

FIPS/fip-filplus.md
anorth: I mean these as alternates. A sector, or deal, or allocation can have a term.

FIPS/fip-filplus.md
anorth: Which costs? There may be slightly more gas cost for verified deals, but the operational aspects remain constant because the market actor acts as a delegate.

FIPS/fip-filplus.md
anorth: Changed "chunk" to "range". ranges are explained below.

FIPS/fip-filplus.md
anorth: Yes, clarified. I will invert the sense so that the default `false` is the value the market will always use.

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+
+An allocation is successfully claimed if at least one range of the allocation 
+is committed before the allocation expires.
+Other ranges may be committed after the expiration epoch. 
+This provides symmetry with resumption after faults.
+
+This larger-than-sector allocation capability will a more scalable representation within
+the verified registry actor.
+Full realisation of this capability will require changes to the
+storage miner actor that are beyond the scope of this proposal.
```
jennijuju, line 182:
```suggestion
This larger-than-sector allocation capability will a more scalable representation within
the verified registry actor.
Full realisation of this capability will require changes to the
storage miner actor that are beyond the scope of this proposal.
```

ah..... I wish "out of scope" was stated the earlier.  im also wondering, shouldn't related mechanism to be introduced in a followup FIP later?

FIPS/fip-filplus.md
```diff
@@ -0,0 +1,470 @@
...
+    Size: uint64
+    // The minimum period which the provider must commit to storing the piece (from allocation).
+    TermMinimum: uint64
+    // The maximum period for which the provider can earn QA-power for the piece (from allocation).
+    TermMaximum: uint64
+	// The epoch at which the (first chunk of the) piece was committed.
+	TermStart: uint64
+    // The sector(s) in which the data is committed.
+    // This representation permits a claim to span pieces of data in multiple sectors.
+    // Ranges are over the allocation data, and are non-overlapping. (0, 0) means all.
```
jennijuju: given `All or nothing` is defined as out of scope of this fip, do we want to introduce sector arrays in this FIP? I personally think we should try to simplify the fip given its quite big already.

or at least, given that `AddorNothing` is always true, we should also say that `range start and range size` will also be 0 & 0?

FIPS/fip-filplus.md
jennijuju: (I will come back to this after finish reading all then.. im leaving comments as I read it through, so might be missing full pic in the beginning

FIPS/fip-filplus.md
jennijuju: im not too sure about sector & deals, in the protocol today, we use `duration` for deals and `lifetimes` for sectors, im not sure updating the wording in this fip is the best 🤔 

so maybe only use `term` on allocation?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+	
+    // Sequence number for allocation IDs.
+    NextAllocationId: AllocationID // uint64
+}
+```
+
+#### Operations
+
+An allocation is created by a verified client by transferring data cap tokens (see below)
+to the verified registry.
```
jennijuju, line 99: whats the method for this operation? i.e: do we have a `CreateAllocation`

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+
+
+
+#### Operations
+An allocation is claimed when a storage miner actor proves the corresponding piece of data 
+is committed to a sector.
+
+```
+struct SectorAllocationClaim {
+    // The allocation being claimed.
```
jennijuju, line 198: we probably need a client id here 


given
```    
// Allocations indexed by client, then by ID.
    Allocations: HAMT[Address]AMT[AllocationID]Allocation
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+    // The sector into which the piece has been committed.
+    SectorID: uint64,
+ 
+}
+struct ClaimAllocationsParams {
+    Sectors: []SectorAllocationClaim
+}
+struct SectorClaimResult {
+    // Indices of allocation claims that succeeded.
+    Successful: []int64
```
jennijuju, line 219: same here? (clientID, allocationID)?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+
+Claims are stored in the verified registry actor state, grouped by provider address. A claim's ID is
+inherited from the allocation that created it.
+
+```
+struct State {
+    // ... Existing verified registry state ...
+
+    // Claims by provider ID, then Claim ID.
+    Claims HAMT[Address]AMT[ClaimID]Claim
```
jennijuju, line 165: why dont we keep this information in https://github.com/filecoin-project/FIPs/blob/52e31bb51edd18dfdbe56dd5bfaaf5475449cff8/FIPS/fip-0045.md?plain=1#L46?
```
Allocation {

Claimed: boolean
}
```

as the amount of the datacap and verified deals increase in the network, the size of the verifreg state might grow a lot. have one less state variable might reduced the state size -> save the gas cost when loading the state object

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+    // The latest epoch by which a provider must commit data before the allocation expires.
+    Expiration: Epoch
+    // Whether the provider can commit the data in multiple pieces.
+    // This is initially un-used, but supports allocations larger than a full sector in the future. 
+    AllowRanges: bool
+}
+```
+
+An allocation specifies a range of terms for which the provider may commit the data, between some
+minimum and maximum. An allocation's minimum term must be at least as large as its maximum term.
```
geoff-vball, line 70: Is this a typo?

FIPS/fip-0045.md
geoff-vball: I think this is answered on line 287

FIPS/fip-0045.md
ZenGround0: This is a good question, one thing that helps with the concern you are raising is:

> Allocations are also removed when claimed by a provider (see below)

so information for each use of datacap (allocation or claim) is on chain only once.  However the motivation behind separate maps when we could do it as you propose is also not fully clear to me.

FIPS/fip-0045.md
ZenGround0: I don't thinks so, we can just look up the provider's claim table and then the claim by id.

anorth: These are indices in the parameter, not allocation IDs.

FIPS/fip-0045.md
anorth: Claims have different data than Allocations

FIPS/fip-filplus.md
anorth: It could be, but that would push larger-than-sector deals out at least a year, whereas I can make it happen alongside this and the subsequent change.

However, all I really need is a placeholder in the Claim here, so I could defer many the details to the next FIP.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+}
+```
+
+#### Operations
+
+An allocation is created by a verified client by transferring data cap tokens (see below)
+to the verified registry. 
+The `Allocation` metadata must be provided as transfer metadata.
+The verified registry checks the same preconditions as the existing `UseBytes` method. 
+The size of the allocation must match the number of tokens received.
```
AxCortesCubero, line 102: I think there is specification missing for me to fully understand this data allocation token. Here are key points I think I understand:

1) Token is minted by verifiers/ notaries, to give to a client they verified to be a real data client. They mint a certain amount to the person to use (not to a particular data) at their discretion, I guess the amount they mint depends on how valuable a client this is verified to be?

2) Token cannot be transferred from one user address to another? only transferable to the verified registry? Is this something that will be impossible to attempt to do? penalized if done?

3) Amount of token will be mapped to "amount of data" to seal in an allocation. I am less clear on this, will this be sort of stabilized, "always one token corresponds to 1GiB" or something? Will it be subject to market force? "this is how much data my token can buy me right now to allocate because too much token was minted"?   Does the token buy an amount of data "space" or an amount of data "space-time" as in "I have one token so I can choose to allocate 10GiB of data for one year, or instead 5 GiB for two years"?

4) The token is destroyed/burnt when it is deposited for an allocation AND then that becomes a sealed claim.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+    // for the piece (epochs).
+    TermMaximum: uint64
+    // The latest epoch by which a provider must commit data before the allocation expires.
+    Expiration: Epoch
+    // Whether the provider can commit the data in multiple pieces.
+    // This is initially un-used, but supports allocations larger than a full sector in the future. 
+    AllowRanges: bool
+}
+```
+
```
AxCortesCubero, line 68: Should this have also something about the amount of tokens used to create this allocation? Amount to be burnt if claimed later?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+The term cannot end during a sector's lifetime, except by explicit replacement of data.
+If data is replaced, the sector immediately losses the power multiplier for that fraction
+of the sector's capacity.
+(Replacing non-zero data in a sector is not possible today).
+
+Aligning the claim's term with the sector's expiration exactly in this way removes the
+"spreading-out" of QA power over a sector's life in the current method of computation.
+Providers gain 10x power for each byte of verified data exactly while that data is proven.
+The miner actor computes a sector’s power as `SectorSize + (9 * VerifiedSize)`.
+
```
AxCortesCubero, line 319: It is unclear to me how this works with the idea of the maximum term. Say you have a one year sector and we make an allocation to store my data with maximum term of 8 months.  You Claim that to you sector start earning 10X for 8 months. It is unclear what happens at the end of the 8 months. Is this like your previous FIL+ indefinite proposals where even if our allocation is for 8 months, you will get the full QAP for the whole 1 year sector?

Also, I know you previously wanted to decouple QAP from consensus power, and I dont know how exactly the 10X reward will work if the information is kept in this separate actor. will it be that:
a) SP will be 10 times more likely to win a tipset and write a block?
b) Probability for writing a block is not enhanced, but when they win they get 10X the jackpot?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+When the minimum term for a verified piece expires, the pledge requirement does not change.
+The provider's commitment to storing the verified piece extends until 
+the sector's scheduled expiration. 
+If the sector is terminated early, the termination penalty includes the boosted power.
+
+If a verified piece is transferred to new sector (see below),
+the initial pledge for the new sector is calculated using the network conditions at the time.
+The pledge requirement for the old sector is not decreased, 
+but any subsequent penalty is calculated using the current (reduced) power.
+
```
AxCortesCubero, line 331: I think this point is very important to fully clarify. I would expect that an additional piece of collateral needs to be placed at the time when SP starts earning 10X, when the claim is sealed. But this is not really specified here.

If QAP can change throughout the lifetime of a sector then I would imagine the collateral also changes, increases when the QAP increases?

If collateral will work in this "not smeared over sector lifetime" way, this has a potentially very large economic impact to the network that we would need to evaluate. 

The idea is economically, collateral from sectors can be viewed as a "loan" to the network, freeze these funds, which is good for the rest of the network, for a period of time. The point is that the terms of these loans are changing (and perhaps maybe rightfully so, this may be right thing to do) but it strictly resuts in a "bad deal" for the rest of the network. If you loan me 10 dollars that I have to pay back in one year, it is not the same as if you loan me 5 dollars that I have to pay you back in 2 years. The second loan is a more valuable one for which I should pay you back more interest.

Currently collateral works in the second "smeared way", which economically speaking is "better for the network" than the new version.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+### Storage miner actor
+The built-in miner actor is trusted to enforce the verified and the continual storage of the data.
+
+The miner actor requires that the sector into which a claimed piece is committed must have a
+**remaining lifespan that is sufficient to fulfil the minimum term**.
+Thus, when a sector terminates at its scheduled expiration epoch, the claim's term has been met.
+
+Similarly, the miner actor requires that sector into which the piece is committed must have a
+**remaining lifespan no longer than that which would reach the maximum term**.
```
AxCortesCubero, line 340: Ah, I think this answers one of my previous questions! about the  claim ending before the sector ends.

FIPS/fip-0045.md
Kubuxu: The data cap token functions the same as current allocation tracking, it is just using the token interface.

> I guess the amount they mint depends on how valuable a client this is verified to be?

Notaries allocate some amount of data cap (in bytes) to a given client by minting some amount of the token. The unit of the data cap (byte IIRC) is the same unit as the unit of the token.

> Token cannot be transferred from one user address to another?

Yes.

> only transferable to the verified registry?

Yes.

> Is this something that will be impossible to attempt to do?

The Token contract will only allow for transfers to the verified registry or by the verified registry to some other party (to facilitate the refund flow).
See https://github.com/filecoin-project/FIPs/pull/432/files#diff-da7f03d0785adde81d03c9123f3d0cf50d8633ca6ac08d826a561239ca7dc66cR259

> always one token corresponds to 1GiB

The current plan is 1 token to 1 unit of data cap (which AFAIK is 1byte)

> Will it be subject to market force?

No such plans at the current time

> The token is destroyed/burnt when it is deposited for an allocation AND then that becomes a sealed claim.

Yes,

anorth: And these are all unchanged from the current mechanism, except for moving the location in state of the data cap balance table.

FIPS/fip-0045.md
anorth: > maximum term

You discovered the answer below. A sector can't claim an allocation that's shorter than the sector's lifetime.

> decouple QAP from consensus power

This proposal doesn't decouple QAP from consensus power, but simplifying the calculation solves many of my original motivations for doing so. The SP will get 10x consensus power.

FIPS/fip-0045.md
anorth: > I would expect that an additional piece of collateral needs to be placed at the time when SP starts earning 10X

I meant to imply this with "the initial pledge for the new sector is calculated using the network conditions at the time". The new sector pledge accounts for its FIL+ contents. I can clarify in the body.

> the terms of these loans are changing

They can only change in a way that is "good for the network". The amount can only increase, and the term can only be extended. Could you state a concrete scenario that you're worried about?

FIPS/fip-filplus.md
jennijuju: >  so I could defer many the details to the next FIP.

I think this would be a good option!

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+    // Indices of allocation claims that succeeded.
+    Successful: []int64
+}
+// Called by storage miner actor to claim allocations for data provably committed to storage.
+// For each allocation claim, the registry checks that the provided piece CID
+// and size match that of the allocation.
+fn ClaimAllocations(params: ClaimAllocationsParams) -> ClaimAllocationsResult
+```
+
+The record of a claim may be removed by the provider after it has been completed.
```
jennijuju, line 227: could you please define whats "been completed" mean?
and how can one remove it?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+Neither of those standards are pre-requisite for this proposal, however.
+The built-in actors will be retroactively updated to support those standards when they settle.
+
+#### Operations
+The `AddVerifiedClient` method is changed to invoke the data cap token actor to
+mint new tokens to the client's balance.
+
+The `UseBytes` method is removed.
+To create an allocation, a verified client transfers data cap tokens to the verified registry actor.
+The transfer must be accompanied by one or more `Allocation` records which specify
```
jennijuju, line 289: when an allocation is claimed - does the token get to be transferred from the verifreg -> datacap token contract under the miner actor's balance?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+TODO!
+
+The power for existing sectors is not changed by this proposal.
+
+A provider may opt-in to migrate a sector to the new QA power calculation method.
+A new method on the storage miner actor:
+- Fetches information about each not-yet-expired verified deal from the market
+- Calls the verified registry to create a claim for each such deal.
+  The claim's minimum term is the deal's duration, and maximum term is the sector's current expiration epoch
+- Re-calculates power according to the amount of space committed to active claims.
```
geoff-vball, line 433: @anorth It seems to me to be a little arbitrary that a deal that expires tomorrow would give 10x until the sector expires, but a deal that expired yesterday would give no power at all. It would be a lot simpler and allow for a network-wide migration if we treated expired and non-expired fil+ deals the same here.

FIPS/fip-0045.md
jennijuju: > Allocations are also removed when claimed by a provider 

hrm.. dont we need to keep it in case the client wants to do extend term max?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+struct State {
+    // ... Existing verified registry state ...
+
+    // Claims by provider ID, then Claim ID.
+    Claims HAMT[Address]AMT[ClaimID]Claim
+}
+```
+
+A data cap allocation may identify data that is larger than a single sector.
+Such an allocation may be satisfied by multiple claims, committed to different sectors,
```
hannahhoward, line 170: > an allocation may be satisfied by multiple claims
> A claim's ID is inherited from the allocation that created it.

These two statements seem in contradiction for the current state data structure which is:

```
 Claims HAMT[Address]AMT[ClaimID]Claim
```

Am I missing something? Can an allocation be satisfied by multiple providers?

hannahhoward: In other words, if an allocation is satisfied by two claims that both have a ClaimID that is the AllocationID, how are these stored in the AMT?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+TODO
+
+## Incentive Considerations
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
```
jennijuju, line 473: the proposal should include:

the gas change of a PSD message that triggers an allocation

the gas change of a Provecommit/PCA/PRU message that triggers a claim

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+title: De-couple verified registry from markets
+author: Alex North (@anorth)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/313
+status: Draft type: Technical category : Core created: 2022-08-12 requires: FIP-0034
+---
+
+TODO:
+- Method for extending a sector, SP provides allocation ids (because market doesn't retain them after deal expiration)
+- Spec methods for sector migration on miner, market, and verifreg actors
+- Confirm whether to include the super-sector representations now, or defer to later
```
jennijuju, line 12: Defer to later +1

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+- Find a better (compressed) representation of the piece ranges in a claim
```
jennijuju, line 13: Defer to later ?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+- Discuss how market delegate means no external API changes
+- Discuss mechanisms for conditional allocations, negatively priced deals
```
jennijuju, line 15: Why is this a requirement? Can we defer to later?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+- Market method for fetching allocation ids for deals (needed?)
+- Spec hook/method for a different client extending the claim by spending data cap (needed?).
```
jennijuju, line 17: This seems nice and can be something that DP folks  like  @dkkapur would be interested for programs like EG. 
however, is there any soon-to-land use case exists to support  this feature to  a high priority that must land in nv17? If not, I’d defer to a later fip.

FIPS/fip-0045.md
AxCortesCubero: >I meant to imply this with "the initial pledge for the new sector is calculated using the network conditions at the time". The new sector pledge accounts for its FIL+ contents. I can clarify in the body.

Ok yes I think this can be more explicitly specified. Let's say I seal a 1 year sector, and I make a 6 month claim on it, which starts on the second half. So I place 1X pledge at the beginning of the sector, then when the claim starts halfway through the sector, I need to deposit the remaining 9X of the pledge (roughly but not exactly because this would be based on current network conditions, different from 6 months ago) before I start earning 10X reward. Is this how it works?

AxCortesCubero: >They can only change in a way that is "good for the network". The amount can only increase, and the term can only be extended. Could you state a concrete scenario that you're worried about?

Let's take the same example, 1 year sector, with 6 month claim on its second half. Currently the network would have 5.5X locked for 1 year duration. With this proposal the network would have 1X locked for 6 months, and 10X locked for 6 months. Financially the first loan is "better".

AxCortesCubero: Another possible issue I am thinking is the maximum term mechanism can introduce an incentive for sealing shorter sectors. If a client offers an SP an allocation with 6 month max term, and the SP now has to put it in a sector. Currently the SP can seal that in a new sector and might as well seal it in a longer 1 year sector or whatever. Specially if Duration multiplier passes, they would seal it in the longest sector they can, so they can multiply their multipliers together and earn more.

with this proposed version, they only have the options of a) add it to a sector they have that only has 6months or less left, or b) seal a new 6 month sector. 

All FIL+ will be added to sectors that are exactly the right size, while otherwise SP's could have sealed longer sectors (which aligns more with the sectors as containers idea).


I appreciate the point, I think this is an elegant solution to simplifying the QAP calculation, and I generally prefer this current proposal to some of the previous iterations like FIL+ forever.  Just figuring out still what may be the compromises and if they are worth it.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+    // The provider storing the data (from allocation).
+    Provider: Address
+    // The client which allocated the data cap (from allocation).
+    Client: Address
+    // Identifier of the data committed (from allocation).
+    Data: CID
+    // The (padded) size of data (from allocation).
+    Size: uint64
+    // The minimum period which the provider must commit to storing the piece (from allocation).
+    TermMinimum: uint64
```
misilva73, line 139: Why do claims also have a min and max term? Aren't claims already associated with specific sectors?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+}
+```
+
+A data cap allocation may identify data that is larger than a single sector.
+Such an allocation may be satisfied by multiple claims, committed to different sectors,
+each for a distinct, non-overlapping range of the allocation's data.
+Each sector hosting a range of an allocation gains power individually.
+All ranges from a single allocation must be claimed by the same provider.
+If a sector hosting one range of an allocation is lost, 
+the provider will pay the usual termination penalty but can re-seal the range into a new sector
```
misilva73, line 175: I am not sure what this means. As an example, let's imagine that we have claim for storing a deal for 2 years. The sector is sealed and the SP starts getting rewards. After one year, the sector is terminated and the termination fee is paid. What happens to the claim? Can the SP seal another 2-year sector with the same claim?  Or do we consider that one year of data cap was already spent?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+}
+// Called by storage miner actor to claim allocations for data provably committed to storage.
+// For each allocation claim, the registry checks that the provided piece CID
+// and size match that of the allocation.
+fn ClaimAllocations(params: ClaimAllocationsParams) -> ClaimAllocationsResult
+```
+
+The record of a claim may be removed by the provider after it has been completed.
+
+The maximum term for a claim may be increased by the client which originally made the allocation,
```
misilva73, line 229: Is this possible for a claim related to data already sealed in a sector? If yes, does the SP need to change the list of sectors of this claim?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+}
+
+// Called by allocation client to extend the maximum term of their allocations, 
+// without consuming data cap.
+// The new TermMaximum must be no less than the current one,
+// and no greater than the maximum initial term.
+fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
+```
+
+### Data cap token
```
misilva73, line 259: Can you give more details about the interface for this?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+The actual term for a claim begins at the epoch the data is committed into a sector,
+proven either with proof-of-replication or replica-update. 
+There is no distinct "deal start epoch".
+A sector containing a data cap claim is immediately eligible for a power multiplier of 10 
+for that fraction of the sector's capacity occupied by the incentivized data
+(subject to the existing requirement to initially prove the sector in Window PoSt).
+
+The actual term for a claim ends the epoch that the sector containing the data expires, 
+or that data is replaced. 
+The term cannot end during a sector's lifetime, except by explicit replacement of data.
```
misilva73, line 310: To clarify, does this mean that a claim with a max term of 1 year cannot be stored in a 2-year sector?

misilva73: Also, if we have a claim with a min term of 1 year, we cannot store it in a 6 months sector, right?

FIPS/fip-0045.md
anorth: The claim includes the client and the term max, so the client can extend it.

FIPS/fip-0045.md
anorth: I've added a couple of clarifying statements in the body, especially the 1 datacap = 1 byte relationship.

FIPS/fip-0045.md
anorth: > Let's take the same example, 1 year sector, with 6 month claim on its second half. Currently the network would have 5.5X locked for 1 year duration. With this proposal the network would have 1X locked for 6 months, and 10X locked for 6 months. Financially the first loan is "better".

Thanks for the example. I understand the point but think the example is not representative of real behaviour. Given a deal that doesn't start until six months in the future, the network doesn't get to choose between these cases even today. An SP would just wait and put the 6-month deal is a 6-month sector when it was actually due to start. (I'm ignoring the idea of "storage futures", which would be better implemented higher level than concrete sectors).

> the maximum term mechanism can introduce an incentive for sealing shorter sectors. If a client offers an SP an allocation with 6 month max term, and the SP now has to put it in a sector. Currently the SP can seal that in a new sector and might as well seal it in a longer 1 year sector or whatever

I concur that there is potential incentive for _some_ sectors to have short commitments, up to the client demand for low  term maximums (i.e. clients that want their data unincentivised after some period, for some reason). An SP doesn't have to take such a deal though – if they can get better return on a long commitment sans-deal, they should take it. So I'm actually a little more worried that short deals will find a scarce supply and have to pay, because SDM pays miners to prefer long ones.

Note that the incentive to short sector already exists because of the spread-out QA power. Up to sealing costs, a rational SP picks the shortest sector duration for their deals so their payout and risk of penalty is as short as possible. So all up I don't think this will exert any significant pressure on sector commitment duration.

FIPS/fip-0045.md
anorth: I mean the term maximum has passed. I'll add in the method for cleaning up.

FIPS/fip-0045.md
anorth: No, it's burnt

FIPS/fip-0045.md
jennijuju: Ah - I see what it’s like now!

FIPS/fip-0045.md
anorth: We cannot treat expired deals the same because they're gone. The data is no longer on chain.

I don't disagree about some amount of arbitrariness here, which is why it has to be opt-in.

Note there's also a rough edge about sectors that expire very soon after the migration: there's little time to seek client extension of the terms. We could adopt a policy of extending the migrated term somewhat past the sector's expiration. See discussion in #313.

FIPS/fip-0045.md
anorth: > Can an allocation be satisfied by multiple providers?

No, they had to be the same provider.There was a single claim record in the AMT with an internal array of sector IDs.

 But I'm removing this now anyway.

FIPS/fip-0045.md
anorth: I can probably note the directional impact, but I can't calculate any numbers until it's fully implemented and integrated.

FIPS/fip-0045.md
anorth: It's automatically possible here, I just wish to explain it.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+<!--A short (~200 word) description of the technical issue being addressed.-->
+TODO
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+TODO
+
+## Specification
+**Terminology**
+- *Allowance*: data cap provided to a verifier by the root, or to a client by a verifier.
```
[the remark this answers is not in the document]
anorth: Allowance referred to the pure numerical values, while an allocation is a structured object with rules. But seems I've purged the term allowance anyway, so can drop this.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+    // This is initially un-used, but supports allocations larger than a full sector in the future. 
+    AllowRanges: bool
+}
+```
+
+An allocation specifies a range of terms for which the provider may commit the data, between some
+minimum and maximum. An allocation's minimum term must be at least as large as its maximum term.
+
+Due to the implementation of term enforcement (details below), clients should leave a large buffer
+between the minimum and maximum term to make the allocation practical for a provider.
```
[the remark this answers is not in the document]
anorth: Clients can pick. I guess we could chose a minimum but it would only be for practicality, not correctness.

FIPS/fip-0045.md
jennijuju: https://github.com/filecoin-project/builtin-actors/issues/566

FIPS/fip-0045.md
```diff
@@ -0,0 +1,508 @@
...
+
+**Parameters**
+
+- The minimum allowed value for `TermMinimum` is 6 months.
```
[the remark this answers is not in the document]
anorth: Yes, exactly.

FIPS/fip-0045.md
anorth: Preserving the information from the allocation so the sector can be extended (if allowed by the terms)

FIPS/fip-0045.md
anorth: When it was first sealed, the claim got a `TermStart` epoch. It's valid only until `TermStart + TermMaximum`. So the new sector must have a scheduled expiration before that. So yes, we basically consider that one year of the allocation was already used.

FIPS/fip-0045.md
anorth: No (we lack the technique today to prove inclusion of some data in an already-sealed sector).

FIPS/fip-0045.md
anorth: https://github.com/filecoin-project/FIPs/pull/435

I'll reference an FRC number here when I get it.

FIPS/fip-0045.md
anorth: Yes and yes.

FIPS/fip-0045.md
anorth: The SDM proposal could cause a high priority need for it.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+Allocations are stored in the verified registry actor's state, grouped by client address.
+Nesting by client promotes more efficient lookups at scale, 
+and supports per-client iteration for revocation.
+
+```
+struct State {
+    // ... Existing verified registry state ...
+    
+    // Allocations indexed by client, then by ID.
+    Allocations: HAMT[Address]AMT[AllocationID]Allocation
```
ZenGround0, line 88: Becasue AllocationIDs are global I think this and claims should be HAMT[Address]HAMT[AllocationID].  AMTs do best with sequential keys.

Kubuxu: Agreed, it used to be just `AMT[AllocationID]Allocation` initially IIRC.

FIPS/fip-0045.md
misilva73: If I understand this correctly, does this mean that sectors with deals sealed before this FIP is introduced won't be able to extend?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+    // Client of the allocation being claimed.
+    Client: Address
+    // The allocation being claimed.
+    AllocationID: AllocationID
+    // The piece data committed (either the allocations data, or a part of it).
+    PieceCID: CID
+    // The (padded) size of the data committed (up to the allocation's data size).
+    PieceSize: uint64
+    // The sector into which the piece has been committed.
+    SectorID: uint64,
```
ZenGround0, line 175: @anorth I think we need sector expiration here so we can check that the allocation's max and min terms are satisfied

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+  TermMaximum: deal.EndEpoch - deal.StartEpoch + EPOCHS_IN_YEAR
+  Expiration: deal.StartEpoch
+  AllowRanges: false
+  }
+```
+
+The market stores the allocation ID in deal state.
+
+```
+struct DealState {
```
ZenGround0, line 383: Because we get the allocation ID on PSD and deal state is not yet written this isn't quite enough. We'll either need to rethink writing deal state earlier which seems harder, or add a new map HAMT[DealID]AllocationID which seems easy.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+}
+```
+
+When a verified deal expires without being committed, the data cap allocation remains.
+The market does not revoke the allocation during cron handling.
+Revocation is a manual operation that the client must perform.
+
+When a verified deal is activated by the provider, the market actor returns 
+the allocation ID and piece metadata to the storage miner actor.
+The miner actor then invokes `ClaimAllocations` and, if successful,
```
ZenGround0, line 399: I think we should keep the market actor as the delegate for this during activate deals.  If we do it this way I think we'll be introducing logic from market to miner that we are just going to remove when we start calling ClaimAllocations directly from miner.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+A provider may opt-in to migrate a sector to the new QA power calculation method.
+A new method on the storage miner actor:
+- Fetches information about each not-yet-expired verified deal from the market
+- Calls the verified registry to create a claim for each such deal.
+  The claim's minimum term is the deal's duration, and maximum term is the sector's current expiration epoch
+  (TODO: plus some buffer?)
+- Re-calculates power according to the amount of space committed to active claims. 
+
+This will generally result in an increase in power for the sector.
+Note that power may be lost for deals that have already expired,
```
ZenGround0, line 416: TODO: policy for handing opt-in on sectors with expired verified deals
1. All expired deals don't have a claim but non-expired verified ones still create an claim
2. Any expired deal errors the whole opt in -- simpler, safer(?)

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+The new sector’s committed lifespan is still constrained to fit within the claim’s term.
+
+#### Sector extension
+A provider cannot extend the scheduled expiration for a sector past the maximum term 
+of any piece of verified data that it holds. 
+The provider must forgo the verified power boost when processing such a sector extension. 
+This does not reduce the sector’s pledge requirement, 
+but does reduce penalties to be paid on any future fault to reflect the new, lower power.
+
+This facility to drop a claim can only be exercised:
```
ZenGround0, line 344: I think we are aligned that we'll move away from this approach and instead allow extension in the case all relevant claims have permissive max terms.

ZenGround0: **notes**

Node keeps map[sectorID][]allocationID.  We could keep explicitly on sector info but we should avoid for scaling.  This info is onchain in verifreg claims table but operator can keep a fast index offchain.

Migrate sector infos to contain a boolean on whether they have opted into new FIL+ accounting paradigm.  Then the opt in migration will flip this bit when it does the migration computation.

The old extension endpoint remains unchanged for extending sectors that have not opted into the new FIL+ accounting paradigm.

Introduce a new endpoint that is mostly the same as old extension but it has an additional check.  If sector has verified power then go to verified registry and gather claim data for provided allcoation IDs (that operator passes in from offchain map).  Verifreg claims must match provider and sector id.  Furthermore total size of allocations must match exactly the QA power of sector.  Because of simplified QA power rules in new paradigm this should be straightforward.  After passing these checks this new method knows it is looking at all relevant claims.  Now for all claims this method checks that term_start + term_max doesn't exceed the new intended expiration.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,488 @@
...
+    SectorID: uint64,
+ 
+}
+struct ClaimAllocationsParams {
+    Sectors: []SectorAllocationClaim
+}
+struct SectorClaimResult {
+    // Indices of allocation claims in the parameters that succeeded.
+    Successful: []int64
+}
```
ZenGround0, line 184: I think I agree that it will be useful to allow miner actor to continue committing a sector whose deals pass all by verifreg claim allocation.  So the batching error behavior here is correct.  What I really don't like is the continuation of this pattern that I've been adding to actors.  I've added this to a few methods and the lack of error information is not good.  This shows up in testing, and especially in user debugging for user facing methods like PSD.  Its foreseeable that error info could impact actor behavior too.

Here's a proposal for something compact but errorful:

```golang
type BatchReturn struct {
    // we could also use []int64 but bitfield is nice for the common case
    // where there are no failures
    Success: BitField, // indexes of successful params
    FailCodes: []ExitCode, // every failure appends an exitcode to this list
}
```
len of params == len(Success) + len(FailCodes).  Indices start at 0.  So this fully determines mapping from index to success / failure.

We can implement methods like `count()` to return total index, `success(index: u_size)` to return success or not at index and `code(index: u_size)` to return Ok for succeses, error code for failures.

FIPS/fip-0045.md
Kubuxu: > I think we are aligned that we'll move away from this approach and instead allow extension in the case all relevant claims have permissive max terms.

@ZenGround0 I don't understand, the point of this was to allow an SP to keep a sector without the QA adjustment past the `TermMaximum` which would be disallowed without that. Is that still considered or are we dropping this mechanism completely?

FIPS/fip-0045.md
Kubuxu: This is internal communication so BitField encoding/decoding costs might be not worth it. I like the idea of returning fail at least fail codes, I would suggest going with for example: `FailCodes: []struct{index: usize, code: ExitCode}`, this way the happy case is the cheapest.

ZenGround0: Nice I like just using FailCodes.  If we add on one integer field for total index then this struct can be used standalone which I think might be worth the tradeoff for external use too.  It would be simple to standardize this type across internal / external use.

FIPS/fip-0045.md
ZenGround0: Thanks, I have not been considering this as important though maybe I should be.  In light of verifreg claim extension I have been proposing dropping this mechanism for simplicity.

ZenGround0: notes 

If there is a real need to allow sectors to exist with verified deal data without verified deal incentives then this should be a small change to the plan outlined above.  We do the same claim gathering as above.  If there are claims that don't cover the new sector lifetime the method checks if the sector is in the grace period at the end of life, and if so it drops verified weight and recalculates QA power, otherwise failing.

FIPS/fip-0045.md
anorth: Sorry, I misread your original question.

Yes, the term for a claim may be extended – all claims are already sealed in a sector. But this doesn't imply any change to the sector.

FIPS/fip-0045.md
anorth: I'll add something like this, we can finesse in implementation

FIPS/fip-0045.md
anorth: I'm going to keep the claim-dropping functionality in the spec for now. Everything else above sounds good.

FIPS/fip-0045.md
anorth: If the new map is just for un-activated deals, sounds good.

FIPS/fip-0045.md
anorth: We discussed and agreed to try doing it from the miner, unless it turns out very clumsy.

FIPS/fip-0045.md
anorth: I think we have to do (1) for fairness, especially with the sector duration multiplier change.

FIPS/fip-0045.md
anorth: -> https://github.com/filecoin-project/FIPs/discussions/313#discussioncomment-3462044

anorth:
@jennijuju I've added a lot of the missing content here. While not final, I think this PR is now ready to merge as a draft. @ZenGround0 and I can then collaborate on improvements as implementation progresses.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+the provider will pay a penalty related to the expected reward earned by the sector (up to some limit).
+
+The verified registry can support re-commitment of the same data
+(even though the built-in storage market actor does not support resumption of a terminated deal).
+If a claim’s maximum term has not been reached when a sector terminates,
+the provider may commit the same piece into a new sector and regain the quality-adjusted power.
+The new sector’s committed lifespan is still constrained to fit within the claim’s term.
+
+#### Sector extension
+A provider cannot extend the scheduled expiration for a sector past the maximum term
```
arajasek, line 393: I'm having a hard time understanding this paragraph. We say that `a provider cannot extend the scheduled expiration for a sector past the maximum term of any piece of verified data that it holds`, but it sounds like a provider _can_ do so (under the circumstances listed in the subsequent paragraph), but will automatically lose the power boost for the extended period if they do so.

Is that correct?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+### Multi-sector allocations
+The verified registry actor and its allocations and claims are not coupled to sector size.
+A future iteration to this schema could support allocations which are larger than a single sector,
+and may be claimed by a collection of data commitments spread across multiple sectors.
+
+Such an allocation would be far more gas efficient than the multiple per-sector deals
+required for a large data set today.
+
+### Support for FIP-0036
+Take alone, FIP-0036 could introduce an unfortunate disadvantage to storage providers with many verified deals,
```
arajasek, line 622:
```suggestion
Taken alone, FIP-0036 could introduce an unfortunate disadvantage to storage providers with many verified deals,
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+The Filecoin Plus verified registry actor is extended to record allocations of data cap to specific
+pieces of data to be stored by a provider. An allocation is independent of any deal with the
+built-in storage market (or any future market).
+
+```
+struct Allocation {
+    // The verified client which allocated the data cap.
+    Client: Address
+    // The provider (miner actor) which may claim the allocation.
+    Provider: Address
```
arajasek, line 95: So this is interesting, and a non-trivial deviation from the FilPlus process today. Today the "verification" process is:

0) Steps are taken for verifiers (notaries) to be authorized -- irrelevant / unchanged by this proposal
1) Notaries give some DataCap to Verified Clients, who can spend it on any data, with any SP as they see fit.
2) VCs negotiate deals with SPs. SPs validate the deal offer by checking that the VC has enough DataCap for the deal.
3) Data is transferred to the SP.
4) The DataCap gets "used" when the SP publishes the storage deal*. Note that the SP could get misled by a VC whose DataCap changes between the validation of the offer in Step 2, and the actual publication of the deal in Step 4.

With this FIP, the process becomes:

0) Steps are taken for verifiers (notaries) to be authorized -- irrelevant / unchanged by this proposal
1) Notaries give some DataCap (tokens) to Verified Clients, who can spend it on any data, with any SP as they see fit.
2) VCs negotiate deals with SPs. SPs validate the deal offer by checking that the VC has enough DataCap for the deal.
3?) VC publishes the Claim, with the matching SP. The SP waits for this publication
4?) Data is transferred to the SP.
5?) The DataCap gets "used" when the SP publishes the storage deal*. 

At the very least, this leads to changes in the validation processes on the SP side. Additionally, it's not clear to me in what order these steps "should" be performed -- that's not consenus-critical, but I'm interested in whether the FIP authors have an idea in their minds.

I'm concerned about the various ways both VCs and SPs could get misled by this. Is the idea for VCs to only publish claims with very short `Expiration`s, so that they don't have their DataCap locked for a long period by an uncooperative SP? Is the idea for SPs to receive all the data _before_ the VC actually publishes the Claim, leaving them vulnerable to having their time and resources wasted by malicious VCs who have no intent of actually publishing a Claim?

I suspect time and discussion has already gone into these questions, so feel free to point me to the appropriate place.

*I might be wrong about the precise details of this.

arajasek (commented):
Thank you for the detailed and well-written FIP. This is a complex change, and understanding it is a lot easier thanks to the FIP.

I do worry that certain aspects of this are overengineered, but I don't yet have anything in mind that I'd suggest simplifying. For now, some notes and questions.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+    
+    // Allocations indexed by client, then by ID.
+    Allocations: HAMT[Address]HAMT[AllocationID]Allocation
+	
+    // Sequence number for allocation IDs.
+    NextAllocationId: AllocationID // uint64
+}
+```
+
+#### Operations
```
arajasek, line 140: I think this section is missing a discussion of how the existing `RemoveVerifiedClientDataCap` will need to be modified, since the `VerifiedClients` HAMT is being removed.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+category : Core
+created: 2022-08-12
+requires: FIP-0034
+---
+
+TODO:
+- Spec methods for sector migration on miner, market, and verifreg actors
+- Spec change to SectorOnChainInfo to distinguish migrated sectors
+- Confirm policy for built-in market actor's default term maximum
+- Market method for fetching allocation ids for deals (needed?)
```
jennijuju, line 17: say this is needed - it would be fetching claims.. right?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+- Spec hook/method for a different client extending the claim by spending data cap (needed?).
+
+
+## Simple Summary
+Represents Filecoin Plus data cap allocations separately from the built-in market actor.
```
jennijuju, line 22:
```suggestion
Represents [Filecoin Plus](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0003.md) DataCap allocations separately from the built-in market actor.
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+Adds explicit term (duration) limits to data cap allocations.
```
jennijuju, line 23:
```suggestion
Adds explicit term (duration) limits to DataCap allocations.
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+Simplifies the quality-adjusted power calculation to remove spreading out over a sector's lifetime.
```
jennijuju, line 24:
```suggestion
Simplifies the quality-adjusted power calculation to remove power spreading out over a sector's lifetime.
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+Supports extending the terms of data cap allocations independently of market deals.
```
jennijuju, line 25:
```suggestion
Supports extending the terms of DataCap allocations independently of market storage deals.
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+Represents un-allocated data cap as a fungible token.
```
jennijuju, line 26:
```suggestion
Represents un-allocated DataCap as a fungible token.
```

(also whats the reason for emphasizing "unallocated" DataCap?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+These changes provide a foundation for user-programmed market actors to broker Filecoin Plus verified deals in the future.
+
+## Abstract
+The current implementation of Filecoin Plus verified deals is deeply coupled to the built-in market actor,
+which is treated as trusted code.
+The built-in market is inefficient, limited in functionality and can only improve very slowly through network upgrades.
+The architectural coupling prevents any user-programmed actors from replicating, improving, or varying deal-related functionality,
+especially with any relation to Filecoin Plus.
+
+This proposal decouples the representation of Filecoin Plus data cap allocations from the built-in market actor's deals,
```
jennijuju, line 36:
```suggestion
This proposal decouples the representation of Filecoin Plus DataCap allocations from the built-in market actor's deals, 
```

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+
+#### Operations
+
+An allocation is created by a verified client by transferring data cap tokens (see below)
+to the verified registry. 
+The `Allocation` metadata must be provided as transfer metadata.
+The verified registry checks the same preconditions as the existing `UseBytes` method. 
+The size of the allocation must match the number of tokens received.
+
+An allocation may be revoked after its expiration epoch has passed.
```
jennijuju, line 148: Could you please who can trigger revoke datacap allocations? client? provider? anyone?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,675 @@
...
+
+// Called by allocation client to extend the maximum term of their allocations, 
+// without consuming data cap.
+// The new TermMaximum must be no less than the current one,
+// and no greater than the maximum initial term.
+fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
+```
+
+### Data cap token
+The `Clients` map of addreses to balances is removed from the verified registry state.
```
jennijuju, line 287:
```suggestion
The `verified_clients` map of addreses to balances is removed from the verified registry state.
```

FIPS/fip-0045.md
jennijuju: as discussed offline, 3&4 happen within 5!

jennijuju:
@anorth  could you please specify, what happens when the sector is early terminated?
- faulty >42 days
- manually terminated 

did the claimed got removed then datacap got burned?

FIPS/fip-0045.md
anorth: Well, yes, because it stops holding (kinda, once we can resnap) that piece of data. I'll try to reword.

FIPS/fip-0045.md
anorth: Added

FIPS/fip-0045.md
anorth: Their IDs, yes

FIPS/fip-0045.md
anorth: Ugh I don't like the unnecessary WordSmooshing but I guess I'll be consistent with that FIP.

FIPS/fip-0045.md
anorth: Yes, the pseudocode spells this out

anorth:
> please specify, what happens when the sector is early terminated

There is no change from behaviour today. Datacap was already burned. The claim is not removed until the provider does so explicitly.

FIPS/fip-0045.md
```diff
@@ -0,0 +1,691 @@
...
+(which today are identical).
+
+#### Failure and repair
+There is **no change to termination penalty** calculations. 
+If a sector is terminated early (continual faults, or manual termination),
+the provider will pay a penalty related to the expected reward earned by the sector (up to some limit).
+
+The verified registry can support re-commitment of the same data
+(even though the built-in storage market actor does not support resumption of a terminated deal).
+If a claim’s maximum term has not been reached when a sector terminates,
```
arajasek, line 393: I don't understand how this would work. Is the verified registry notified when a sector fulfilling a claim expires? Or is the idea that a Claim can be re-claimed if the Sector in the original Claim has expired?

jennijuju:
> > please specify, what happens when the sector is early terminated
> 
> There is no change from behaviour today. Datacap was already burned. The claim is not removed until the provider does so explicitly.

maybe i missed it - but i cant seem to find when the token is burned in this fip yet. Upon a claim?

FIPS/fip-0045.md
```diff
@@ -0,0 +1,691 @@
...
+category : Core
+created: 2022-08-12
+requires: FIP-0034
+---
+
+TODO:
+- Spec methods for sector migration on miner, market, and verifreg actors
+- Spec change to SectorOnChainInfo to distinguish migrated sectors
+- Confirm policy for built-in market actor's default term maximum
+- Market method for fetching allocation/claim ids for deals (needed?)
```
jennijuju, line 17: ah.. we will need this for client implementation, when extend a sector, we need to be able to get the term max of all data in the deals of that sector

jennijuju:
@anorth i believe two things are missing 
- upon upgrade, we should migrate the existing verified clients datacap balance to the token contract, under the clients address. I’m not sure if we need to migrate verifiers or not, iiuc we are keeping the verifier datacap state in verifreg, if so we don’t need to do anything. 
- upon upgrade, we should migrate and add allocations for pending deals (deals that’s has been onchain via PSD), so that, when SP provecommit those deal sectors, they won’t fail the claim due to allocation not found.

FIPS/fip-0045.md
anorth: No the registry is not notified when a sector expires. The miner actor is still trusted to enforce the appropriate penalties etc.

But a claim can be re-claimed if the original sector has expired. Here again the miner actor will be trusted to only do this when the original sector _has_ gone. It's quite likely that the user-facing method to do this won't be landed as part of this nv17, but it's automatically encompassed in the follow-on designs.

FIPS/fip-0045.md
anorth: You can do state inspection. This method will eventually be needed for operators to migrate off that though.

anorth:
Thanks I've added notes about the upgrade migration, including datacap balances. I'm not yet confident about the best way to handle pending deals, so have added TODO.

jennijuju:
I think this FIP still needs a couple of iterations before it can be considered last call ready. However, it has enough information for the first draft of an FIP and has enough community interest to consider the idea for filecoin network. 
That being said, I am merging this PR now - more discussion of the FIP evolvement at https://github.com/filecoin-project/FIPs/discussions/313. FIP authors shall ensure that the unaddressed comments in this PR are either captured in the discussion or tracked elsewhere and will be addressed in the upcoming iterations.

jennijuju (approved):

PR #434, opened 2022-08-17
FRC-0042: expand rejection sampling to numbers < 256

anorth:
Expands the range of method numbers excluded by the rejection sampling mechanism to 256. This removes the possibility of collision with any existing built-in actor method number, and leaves a decent range for new actors to use for internal dispatch or other conventions.

This idea is due to @Kubuxu.


Kubuxu (approved):
It looks correct.

Stebalien:
Can we increase this even more? I.e., reserve 0x7FFFFFFF? (or even just 0xFFFF). There's really no reason to go small here.

jennijuju (approved):

PR #435, opened 2022-08-17
FRC-0046 First draft of fungible token standard.

anorth:
This is the first draft of a fungible token standard API for native actors, as discussed in https://github.com/filecoin-project/FIPs/discussions/407

An implementation of this standard is in development at https://github.com/helix-onchain/filecoin/tree/main/fil_fungible_token

FYI @jsuresh, @alexytsu @ZenGround0

### DIFF
--- a/FRCs/frc-tokenstandard.md
+++ b/FRCs/frc-tokenstandard.md
@@ -1,0 +1,290 @@
+---
+fip: <to be assigned>
+title: Fungible token standard
+author: Alex North (@anorth), Jeeva Suresh (@jsuresh), Alex Su (@alexytsu)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/277
+status: Draft
+type: FRC
+created: 2022-08-17 
+---
+
+## Simple Summary
+A standard interface for native actor fungible tokens. 
+
+## Abstract
+This proposal provides a standard API for the implementation of fungible tokens as FVM native actors.
+Its design goal is a token API conceptually similar to prior art,
+including support for delegated control of tokens,
+but with superior safety and reduced cost and complexity.
+
+The highlights are:
+- Standard token name/symbol/supply/balances, with a fixed decimal precision of 18.
+- Standard allowance protocol for delegated control, but with an API robust to [front-running](https://ieeexplore.ieee.org/document/8802438).
+- Receiver hooks informing receiving actors of incoming tokens (mandatory).
+
+The proposal learns from fungible token standards developed for other blockchain ecosystems.
+It is heavily inspired by [ERC-777](https://eips.ethereum.org/EIPS/eip-777),
+but un-encumbered by backwards-compatibility concerns.
+
+## Change Motivation
+The concept of a fungible token is widely established in other blockchains. 
+As in those ecosystems, the Filecoin actor ecosystem will benefit from a standard API implemented by most tokens.
+A standard permits easy building of UIs, wallets, tools, and higher level applications 
+on top of a variety of tokens representing different quantities.
+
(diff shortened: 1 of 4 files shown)


FRCs/frc-tokenstandard.md
```diff
@@ -0,0 +1,290 @@
...
+// Returns the balance of an address.
+// Must be non-negative.
+fn balance_of(owner: Address) -> TokenAmount
+
+// Transfers tokens from caller to another address.
+// Amount must be non-negative (but can be zero).
+// Transferring to the caller must be treated as a normal transfer.
+// Returns the sender's remaining balance.
+// Aborts if the receiver hook on the `to` address aborts.
+fn transfer({to: Address, amount: TokenAmount, data: Bytes}) -> TokenAmount
```
hunjixin, line 86: what is the usage of data parameter?

hunjixin: any piece of data for the receiver hook?

FRCs/frc-tokenstandard.md
```diff
@@ -0,0 +1,290 @@
...
+
+Methods are to be presented according to a calling convention such as that discussed in [#382](https://github.com/filecoin-project/FIPs/discussions/382), 
+so this recommendation depends on that.
+
+### Interface FRC-XXX
+
+```
+// Token amounts serialize as var-int, and have an implied 18 decimals precision,
+// matching the native token.
+type TokenAmount = BigInt
```
Kubuxu, line 61: There is an argument for using uint128 (or two uint64) for the token amount. I think FVM evaluated both at some point.

FRCs/frc-tokenstandard.md
Kubuxu: Yes, you can imagine a use-case where you pay for something with a token and you want to attach tracking information to it.

Kubuxu:
I see that `decimals()` was discarded. Should we provide something like `granularity()` from EIP-777. While rejecting transactions internally is the correct thing to do, I can see a need for knowing what is the granularity of a given token. For example a generic wallet application and users wanting to transfer half of their tokens.

FRCs/frc-tokenstandard.md
anorth: Yes, thanks I need to clarify that in the text.

jsuresh:
@Kubuxu - I understood the purpose of granularity differently. Love a few more examples from you as I may have missed important use cases.

my understanding is everything has 18 DP as  this being different is an endless source of client bugs. However, in certain use cases it doesn’t make sense to hold or transact partial tokens.

my goto example is games. Eg. Rupees in Zelda have to be ‘whole’ for use in game and therefore trading.

My assumption was in filecoin this kind of use case would be rare. However, would love to hear other motivations/examples of ways you think this would come in handy.

anorth:
Let's take granularity discussion to https://github.com/filecoin-project/FIPs/discussions/407#discussioncomment-3418757 and leave this review for FRC style/completeness.

jennijuju:
hi - this is our FRC-0046

anorth:
I have pushed a new commit which:
- adds granularity method
- expands the return values from some methods
- adds tokenData separate from operatorData for the receiver hook
- clarifies a few things

jennijuju:
is this ready for review? for fip editor?

anorth:
@jennijuju I think it is, but I think there's one more round of minor tweaks coming, so let's wait just a week more.

anorth:
I have pushed updates to
- make the receiver hook universal
- use FRC-42 compliant method names

Let's leave this for a day or two, but then I think it's ready for merge. Of course it's still in draft.

PR #438, opened 2022-08-18
FIP-0036 Draft Update with Latest Parameters

vkalghatgi:
Revised draft of FIP for community consideration. Includes changes reflective of community feedback and further CEL analysis and modeling:
- SDM Function of slope 1
- InitialPledge calculation with 50% multiplier

### DIFF
--- a/FIPS/fip-0036.md
+++ b/FIPS/fip-0036.md
@@ -5,7 +5,7 @@ status: Draft
 type: Technical
 author: Axel C (@AxCortesCubero), jbenet (@jbenet), Maria S (@misilva73), Molly M. (@momack2), Tom M. (@tmellan), Vik K. (@vkalghatgi), ZX @zixuanzh)
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/421 (prev https://github.com/filecoin-project/FIPs/discussions/386)
-created: 2022-07-29
+created: 2022-08-18
 ---
 
 ## Simple Summary
@@ -14,7 +14,7 @@ created: 2022-07-29
 - The Duration Multiplier is multiplicative on the existing Quality Multiplier incentive (Filecoin Plus incentive).
 - Sectors with higher Quality Adjusted Power as a result of Duration Multiplier and Quality Multiplier will require higher initial pledge collateral.
 - The minimum sector duration time will increase to 1 year and the maximum sector duration to 5 years.
-- The SectorInitialConsensusPledge multiplier will increase from 30% to 40%.
+- The SectorInitialConsensusPledge multiplier will increase from 30% to 50%.
 - This policy will apply at Sector Upgrade and Extension.
 
 ## Problem Motivation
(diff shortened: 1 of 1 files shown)


FIPS/fip-0036.md
```diff
@@ -14,26 +14,26 @@ created: 2022-07-29
...
+Per the economic preference to increase the percentage value locked, we also propose adjusting the [Initial Consensus Pledge Mutliplier](https://spec.filecoin.io/systems/filecoin_mining/miner_collaterals/) to 50%. The intention is to create long-term aligned total value locked (TVL) dynamics to support stable and predictable conditions for storage provider (SP) returns. We further discuss the problem/change motivation, and explore impacts on SP profitability and network macroeconomics in the CEL analysis brief [here](https://pl-strflt.notion.site/Duration-Changes-FIP-discussion-Analysis-Summary-735ce6685b7946f0a03fc13c3fe271fa). Note, this brief focuses on a linear duration multiplier with slope 2, and a change to the SectorInitialConsensusPledge calculation from 30% to 40%. Following community feedback and input, CEL has produced a follow-on analysis brief [here](https://pl-strflt.notion.site/Duration-FIP-revisions-7426f344685940409ac513a0ffcccc86), modeling the current revised parameter choices; mainly, FIP 0036 will now propose a Sector Duration Multiplier function with slope 1, and an increase in the SectorInitialConsensusPledge calculation from 30% to 50%.
```
anorth, line 25: Could we please simplify and update resources to only talk about the proposed parameters, not other values? People will get confused.

FIPS/fip-0036.md
```diff
@@ -163,22 +163,21 @@ EpochReward := 20.53 * 5 = 102.7 FIL
...
-
-// Sector quality = 100 (values are approx 50x greater)
-StoragePledge := 0.9802 FIL
-ConsensusPledge := 21.93 FIL
-PreCommitDeposit := StoragePledge = 0.9802 FIL
-InitialPledge := StoragePledge + ConsensusPledge = 22.9102 FIL 
+// Sector quality = 1
+StoragePledge := 0.0098 FIL
+ConsensusPledge := 0.2741 FIL
+PreCommitDeposit := 50 * StoragePledge = 0.4903 FIL
```
anorth, line 169: I was a little confused by the labelling here. I am taking this to be post-FIP-0034, which is why PCD is the same between the two cases. Perhaps just note that the x50 here comes from that.

FIPS/fip-0036.md
```diff
@@ -194,34 +193,31 @@ The existing 1.5 year sector duration limit in effect provides a built-in rotati
...
+5. advPowerDailyPctGain = 0.3%
 ```
 If this scenario is maintained, the adversarial group is expected to exceed 33% of consensus power within 140 days. 
 
 Factors that mitigate this risk are that it’s unlikely a single group could achieve 50% of FIL+ power consistently, and unlikely that the adversarial group exclusively takes up the longer duration sectors with enhanced power multipliers. 
 
 A limitation is that the above calculation assumes the malicious party is starting from 0% of consensus power. If they already control 10%, time to 33% is reduced to approximately 100 days.  
 
-### Rollout Shock
-Rollout shock could occur if SPs race to extend their commitments and gain a further 10x multiplier. This could be mitigated by gradually increasing the maximum multiplier 1x to 5x during an initial period (e.g. first three months)
```
anorth, line 223: Thank you for removing the "could" so that this proposal is concrete. But you can't remove consideration of rollout shock! You can give explanation of forces that would counter it, though.

anorth (approved):
"Final" is optimistic!

jennijuju:
I agree with anorth - Based on the discussion I don’t think the fip is ready to move on to last call, I’d  suggest changing the PR title.

FIPS/fip-0036.md
```diff
@@ -5,7 +5,7 @@ status: Draft
 type: Technical
 author: Axel C (@AxCortesCubero), jbenet (@jbenet), Maria S (@misilva73), Molly M. (@momack2), Tom M. (@tmellan), Vik K. (@vkalghatgi), ZX @zixuanzh)
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/421 (prev https://github.com/filecoin-project/FIPs/discussions/386)
-created: 2022-07-29
+created: 2022-08-18
```
jennijuju, line 8: I dont think we should change the fip creation date
```suggestion
created: 2022-07-29
```

FIPS/fip-0036.md
vkalghatgi: yep fair point will do.

FIPS/fip-0036.md
vkalghatgi: Ok - yes can make this clearer where the 50x comes from on the PCD calculation. Note, it is mentioned in the beginning of this section (line 157) that these calculations are being done post FIP-0034

anorth:
I am at some point going to request that you check in the content of all the links to Notion and other external sites into the `resources/` directory here, so that they will persist with the FIP.

PR #441, opened 2022-08-22
Require full 512-bit Blake2b digest, and up rejection sampling to 24 bits.

anorth:
See discussion here: https://github.com/filecoin-project/FIPs/discussions/382#discussioncomment-3421274

### DIFF
--- a/FRCs/frc-0042.md
+++ b/FRCs/frc-0042.md
@@ -49,19 +49,19 @@ interpreted as an unsigned, big-endian 32-bit integer.
 The `hash` function is given as:
 ```
 let domain_separation_tag = "1|" // Note `|` is illegal in method names
-let mut digest Vec<u8>= blake2b(domain_separation_tag + method_name)
+let mut digest Vec<u8>= blake2b_512(domain_separation_tag + method_name)
 while !digest.is_empty() {
   let method_id = u32::from_be(digest[0..4])
-  if method_id >= 256 {
+  if method_id >= (1 << 24) { // 24-bit reserved range
     return method_id
   }
   digest = digest[4..] // pop-front
 }
-// probability of 2^{-192}, aka it won't happen
+// probability of 2^{-128}, aka it won't happen
 panic!("Method ID could not be determined, please change it")
 ```
 
-Values less than 256 are avoided via rejection sampling, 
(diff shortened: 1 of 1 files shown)


FRCs/frc-0042.md
```diff
@@ -49,19 +49,19 @@ interpreted as an unsigned, big-endian 32-bit integer.
 The `hash` function is given as:
 ```
 let domain_separation_tag = "1|" // Note `|` is illegal in method names
-let mut digest Vec<u8>= blake2b(domain_separation_tag + method_name)
+let mut digest Vec<u8>= blake2b_512(domain_separation_tag + method_name)
 while !digest.is_empty() {
   let method_id = u32::from_be(digest[0..4])
-  if method_id >= 256 {
+  if method_id >= OxFFFFFF { // 24-bit reserved range
```
Kubuxu, line 55:
```suggestion
  if method_id >= (1<<24) { // 24-bit reserved range
```

This changes the semantic, previously 0xFFFFFF was hashed method ID, this way it is a static method ID. Giving static method IDs full 24 bit range.

FRCs/frc-0042.md
```diff
@@ -49,19 +49,19 @@ interpreted as an unsigned, big-endian 32-bit integer.
...
   }
   digest = digest[4..] // pop-front
 }
-// probability of 2^{-192}, aka it won't happen
+// probability of 2^{-128}, aka it won't happen
 panic!("Method ID could not be determined, please change it")
 ```
 
-Values less than 256 are avoided via rejection sampling, 
+Values less than 0xFFFFFF are avoided via rejection sampling,
```
Kubuxu, line 64:
```suggestion
Values less than 2²⁴ are avoided via rejection sampling, 
```

Kubuxu:
IDK if it is worth adding a note over or not, but blake2b-512 and blake2b-256 don't share a common part of the output, as in `blake2b-512(x)[..16] != blake2b-256(x)`. The hash function is keyed by the requested output length.
This was changed in Blake3.

FRCs/frc-0042.md
anorth: Oops, thanks

Kubuxu (approved):

jennijuju (approved):

PR #446, opened 2022-08-31
FIP-0047: First draft of proof expiration and security policy

anorth:
This proposal presents a mechanism for the orderly replacement of sectors in case we discover a flaw in the theory or implementation of PoRep. The initial motivation comes from support for FIP-0036, but this represents a positive change even in the absence of any duration multipliers.

### DIFF
--- a/FIPS/fip-0047.md
+++ b/FIPS/fip-0047.md
@@ -9,8 +9,6 @@ category: Core
 created: 2022-08-31
 ---
 
-- TODO: should the existing ExtendSectorExpiration method also refresh proof, if it can?
-
 ## Simple Summary
 
 Adds a `ProofExpiration` parameter to each sector, 
@@ -192,7 +190,7 @@ and thus requires a network upgrade and sector state migration.
 Impact on tooling should be minimal beyond changes to sector management software.
 
 ## Test Cases
-(Will be) Included with implementation to be presented.
+Will be included with implementation to be presented.
 
 ## Security Considerations
 Decoupling sector proof expiration from their commitment duration allows longer commitment durations 
@@ -201,12 +199,20 @@ without compromising the network’s ability to respond to flaws in PoRep.
 If a flaw is discovered, an increased commitment duration would create the need for replacement sealing, 
 or penalties for non-replaced sectors.
 
+
 TODO: describe the cryptoecon modelling and the simulation outcome
 
 ## Incentive Considerations
 This proposal does not adjust incentives for providers or clients, 
 but it does make longer sector commitments possible. 
 Other proposals may then introduce incentives for such commitments.
+
+In case of an emergency response,
+charging some termination fee when a sector proof is not refreshed creates an incentive
+to seal replacement sectors vs let old sectors expired un-replaced.
+Further analysis is warranted into the _ideal_ termination penalty in this case.
+However, such analysis need not prevent us adopting this mechanism, for its other advantages,
+and policy, in order to provide greater clarity for storage provider business operations.
 
 ## Product Considerations
 If the maximum sector commitment duration is increased, 
@@ -221,7 +227,8 @@ operational risk to the provider,
 which must remember to and succeed in refreshing the proof within the refresh window.
 
 ## Implementations
-The `CommitmentExpiration` and `ProofExpiration` separation implementation is available at TBC.
+The `CommitmentExpiration` and `ProofExpiration` separation implementation is available at 
+https://github.com/filecoin-project/builtin-actors/pull/518.
 
 The replacement sealing implementation is deferred until needed.
 
--- a/FIPS/fip-0047.md
+++ b/FIPS/fip-0047.md
@@ -1,7 +1,7 @@
 ---
 fip: 0047
 title: Proof Expiration & PoRep Security Policy
-author: Jakub Sztandera (@Kubuxu), Irene (@irenegia), Alex North (@anorth)
+author: Jakub Sztandera (@Kubuxu), Irene Giacomelli (@irenegia), Alex North (@anorth), Luca Nizzardo (@lucaniz)
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/415
 status: Draft
 type: Technical
--- a/FIPS/fip-proof_expiration_porep_policy.md
+++ b/FIPS/fip-proof_expiration_porep_policy.md
@@ -1,0 +1,210 @@
+---
+fip: <to be assigned>
+title: Proof Expiration & PoRep Security Policy
+author: Jakub Sztandera (@Kubuxu), Irene (@irenegia), Alex North (@anorth)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/415
+status: Draft
+type: Technical
+category: Core
+created: 2022-08-31
+---
+
+- TODO: should the existing ExtendSectorExpiration method also refresh proof, if it can?
+
+## Simple Summary
+
+Adds a `ProofExpiration` parameter to each sector, 
+and describes a policy to be adopted in case a flaw is discovered in the theory or implementation of proof-of-replication.
+
+## Abstract
+We are not aware of any PoRep issues at this time, but it is crucial to have a plan to deal with any possibility.
+This proposal introduces a concrete mechanism for processing all sectors, 
+and describes (but does not implement) a policy that uses the mechanism.
+
+- The mechanism is to decouple sector `expiration` (renamed to `CommitmentExpiration`) 
+  and proof expiration (a new sector property is introduced called `ProofExpiration`),
+  and introduce a method `RefreshProofExpiration()` that can be used to request a refreshed proof expiration. 
+  A sector’s commitment may then be longer than its proof expiration.
+- The policy is that, in case of a PoRep flaw, 
+  the `RefreshProofExpiration` method is disallowed for sectors sealed with the vulnerable code. 
+  In order to maintain power, a provider must seal a new replacement sector before each existing sector’s `ProofExpriration` epoch is reached. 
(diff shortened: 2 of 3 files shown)


FIPS/fip-proof_expiration_porep_policy.md
```diff
@@ -0,0 +1,210 @@
+---
+fip: <to be assigned>
```
jennijuju, line 2: 0047 please

irenegia (approved):

README.md
```diff
@@ -50,15 +50,15 @@ This improvement protocol helps achieve that objective for all members of the Fi
 |[0008](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0008.md)   | Add miner batched sector pre-commit method  | FIP  |@anorth, @ZenGround0, @nicola  |Final   |
 |[0009](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0009.md)   | Exempt Window PoSts from BaseFee burn  | FIP  |@Stebalien, @momack2, @magik6k, @zixuanzh  |Final   |
 |[0010](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0010.md)   | Off-Chain Window PoSt Verification  | FIP  |@Stebalien, @anorth  |Final  |
-|[0011](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0011.md)   | Remove reward auction from reporting consensus faults  | FIP  |@Kubuxu |Final   |
+|[0011](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0011.md)   | Remove reward auction from reporting consensus faults  | FIP  |@kubuxu |Final   |
```
Kubuxu, line 53: I prefer my nickname to be capitalised.
```suggestion
|[0011](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0011.md)   | Remove reward auction from reporting consensus faults  | FIP  |@Kubuxu |Final   |
```

README.md
```diff
@@ -50,15 +50,15 @@ This improvement protocol helps achieve that objective for all members of the Fi
...
 |[0014](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0014.md)   | Allow V1 proof sectors to be extended up to a maximum of 540 days | FIP  | @deltazxm, @neogeweb3 |Final   |
 |[0015](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0015.md)   | Revert FIP-0009(Exempt Window PoSts from BaseFee burn) | FIP  | @jennijuju, @arajasek |Final   |
 |[0016](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0016.md)   | Pack arbitrary data in CC sectors | FIP  | donghengzhao (@1475) |Deferred  |
 |[0017](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0017.md)   | Three-messages lightweight sector updates | FIP  |@nicole, @lucaniz, @irenegia |Deferred  |
 |[0018](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0018.md)   | New miner terminology proposal | FIP  |@Stefaan-V |Final  |
-|[0019](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0019.md)   | Snap Deals | FIP  |@Kubuxu, @lucaniz, @nicola, @rosariogennaro, @irenegia |Final  |
+|[0019](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0019.md)   | Snap Deals | FIP  |@kubuxu, @lucaniz, @nicola, @rosariogennaro, @irenegia |Final  |
```
Kubuxu, line 61:
```suggestion
|[0019](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0019.md)   | Snap Deals | FIP  |@Kubuxu, @lucaniz, @nicola, @rosariogennaro, @irenegia |Final  |
```

README.md
```diff
@@ -83,3 +83,4 @@ This improvement protocol helps achieve that objective for all members of the Fi
 |[0041](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0041.md)   | Forward Compatibility for PreCommit and ReplicaUpdate | FIP  |@Kubuxu |Draft  |
 |[0042](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0042.md)   | Calling Convention for Hashed Method Name | FRC  |@Kubuxu, @anorth |Draft  |
 |[0045](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md)   | De-couple verified registry from markets | FIP  |@anorth |Draft  |
+|[0047](https://github.com/filecoin-project/FIPs/blob/master/FRCs/fip-0047.md)   | Proof Expiration & PoRep Security Policy | FIP  |@kubuxu, @irenegia, @anorth |Draft  |
```
Kubuxu, line 86:
```suggestion
|[0047](https://github.com/filecoin-project/FIPs/blob/master/FRCs/fip-0047.md)   | Proof Expiration & PoRep Security Policy | FIP  |@Kubuxu, @irenegia, @anorth |Draft  |
```

FIPS/fip-0047.md
```diff
@@ -0,0 +1,229 @@
...
+If `MaxSectorExpirationExtension` is increased, then the built-in storage market's 
+maximum deal duration is increased to match.
+This could permit deals up to 5 years in duration.
+
+### Policy
+If we were to discover a flaw in PoRep theory or implementation:
+
+- A new, fixed PoRep algorithm would be implemented;
+- In a network upgrade, the old algorithm is disallowed and the new one mandated for new sectors;
+- The `RefreshProofExpiration` method is disallowed for old sectors (ie, created with the vulnerable PoRep algorithm),
```
LaurenSpiegel, line 139: What if the network upgrade with the new algo occurs 1 day (or some short time period) before the ProofExpiration Date? SP's will not have time to reseal.

Kubuxu: The SP has 4 months before the network upgrade to trigger the refresh.  Which will give them 1.16yr at that point.

arajasek (commented):
I'm broadly happy with this FIP -- I especially like the simplicity of the proposal.

I have two chief notes, as discussed in the Core Devs meeting yesterday:
- As I understand it, this FIP has sectors expiring at the ProofExpiration epoch. If that is the case, can we consider making the `RefreshProofExpiration` flow automatic? I would propose doing this when processing the expiration queue at Cron (we will be running Cron to "expire" the sector anyway).
- I have misgivings about the the fact that a sector that hits it `ProofExpiration` before its `CommitmentExpiration` (and so expires) pays a termination fee, even if the proof could not be refreshed. I would like to discuss:
  -  why this needs to be the case (I have some speculative guesses why)
  - whether the termination fee in such an instance can be significantly reduced

Kubuxu:
> I would propose doing this when processing the expiration queue at Cron

The work that Cron is doing right now is optimised not to load sector infos, which is what would be required to perform the expiration. Also out of principle, I want the Cron to do as minimal job as possible, if the SP calls `RefreshProofExpiration` before cron hits that epoch, cron doesn't have to do anything.

FIPS/fip-0047.md
LaurenSpiegel: Doesn't that assume that they will have a lot of notice that this change is coming or that SP's always just refresh at the beginning of the 4-month window? In any case, would be better if there was auto-refresh and the SP only had to worry about this if there were a need to reprove in which case they were given ample time to address.

jennijuju: The community is given at least more than 6-8 weeks of notice for any network upgrade (except for the unexpected urgent ones), and that should be plenty of time for SP to extend the expiration if they want. (Tho I don’t think any good SP would want to refresh bad proofs)

anorth: Yes, there is some assumption here of some notice about the change, but I think it's reasonable. It's practically impossible for us to make such a change with no notice, although at our best we have in the past done it with <1week.

FIPS/fip-0047.md
```diff
@@ -0,0 +1,229 @@
...
+If the maximum sector commitment duration is increased, 
+then a sector committed for longer than its initial proof expiration must be refreshed, 
+or it will be terminated with a penalty. 
+This is a different operational position than a sector committed only as as long as the proof duration,
+and then extended. 
+In the latter case, because the sector was not committed to a longer period, 
+no termination penalty is applied if the extension is missed. 
+Committing a sector for longer than the initial proof duration thereby introduces some 
+operational risk to the provider, 
+which must remember to and succeed in refreshing the proof within the refresh window.
```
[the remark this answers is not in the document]
anorth: It is practically necessary for them to trigger the state accounting. I.e., this isn't proposed because we necessarily think it's a nicer product experience for SPs, but because we can't solve the problem effectively otherwise.

anorth:
I concur it's no-go to load sector infos in cron to do this automatically. We can process the termination of expired proofs in cron thanks to extensive memoization data in the expiration queue, but that doesn't include the information to appropriate re-schedule those sectors (which requires per-sector data). It is worth a little thought to discover alternative mechanisms. 

Maybe piggy-backing on another method like window post could work, but we'd need to make sure it only loaded any state when needed (i.e. we'd need to change the method params for the SP to specify the sectors to refresh etc). SPs would probably still be able to forget.

Re termination fee, yes, but please take it to a new proposal. CryptoEcon and CryptoNet are doing some analysis here and maybe there'll be a more network-optimal value. But this proposal shouldn't wait for that work, especially to the extent it's a prereq for things like FIP--036.

anorth:
I have pushed some minor updates here, but I think this is ready to merge as draft.

arajasek (approved):

PR #454, opened 2022-09-05
WithdrawBalance fails when beneficiary is expired or out of quota

geoff-vball:
Related code changes in `builtin-actors` PR [here](https://github.com/filecoin-project/builtin-actors/pull/602)

### DIFF
--- a/FIPS/fip-0029.md
+++ b/FIPS/fip-0029.md
@@ -49,6 +49,7 @@ The following channges apply to specs-actors module:
 
 ### Withdrawals to beneficiary
 - The **WithdrawBalance** method of a miner can be called by the miner's **Beneficiary** or **Owner**, but the balance always being sent to the beneficiary. The total balance withdrawed to a beneficiary address could not exceed the quota.
+- If the **Beneficiary** is expired, or has used all of its **Quota**, then no withdrawls are possible until the beneficiary is updated or changed. Calls to **WithdrawBalance** will fail in this case.
 
 ### Setting the beneficiary address
 - Add **ChangeBeneficiary** method to miner actor to propose or confirm a change of beneficiary address and/or beneficiaryTerm. When changing beneficiaryTerm, the quota must be larger than the usedQuota. The beneficiaryChange is following a proposal/confirmation manner. A **ChangeBeneficiary** proposal can only be submitted by the **Owner**, and it takes effect only after approval(s) related parties if effective. The related parties include the **Owner**, the current Beneficiary, and the proposed Beneficiary. However, there are scenarios of auto-approval (no message required) as below: 
--- a/FIPS/fip-0029.md
+++ b/FIPS/fip-0029.md
@@ -49,7 +49,7 @@ The following channges apply to specs-actors module:
 
 ### Withdrawals to beneficiary
 - The **WithdrawBalance** method of a miner can be called by the miner's **Beneficiary** or **Owner**, but the balance always being sent to the beneficiary. The total balance withdrawed to a beneficiary address could not exceed the quota.
-- If the ** BeneficiaryTerm** is expired, or has used all of its **Quota**, then no withdrawals are possible until the beneficiary is updated or changed. Calls to **WithdrawBalance** will fail in this case with `illegal_state`.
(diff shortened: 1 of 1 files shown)


FIPS/fip-0029.md
```diff
@@ -49,6 +49,7 @@ The following channges apply to specs-actors module:
 
 ### Withdrawals to beneficiary
 - The **WithdrawBalance** method of a miner can be called by the miner's **Beneficiary** or **Owner**, but the balance always being sent to the beneficiary. The total balance withdrawed to a beneficiary address could not exceed the quota.
+- If the **Beneficiary** is expired, or has used all of its **Quota**, then no withdrawls are possible until the beneficiary is updated or changed. Calls to **WithdrawBalance** will fail in this case.
```
jennijuju, line 52:
```suggestion
- If the ** BeneficiaryTerm** is expired, or has used all of its **Quota**, then no withdrawals are possible until the beneficiary is updated or changed. Calls to **WithdrawBalance** will fail in this case.
```

jennijuju (changes requested):

FIPS/fip-0029.md
jennijuju, line 52:
```suggestion
- If the **Beneficiary** is expired, or has used all of its **Quota**, then no withdrawls are possible until the beneficiary is updated or changed. Calls to **WithdrawBalance** will fail in this case, with `illegal_state`.
```

FIPS/fip-0029.md
```diff
@@ -49,6 +49,7 @@ The following channges apply to specs-actors module:
 
 ### Withdrawals to beneficiary
 - The **WithdrawBalance** method of a miner can be called by the miner's **Beneficiary** or **Owner**, but the balance always being sent to the beneficiary. The total balance withdrawed to a beneficiary address could not exceed the quota.
+- If the ** BeneficiaryTerm** is expired, or has used all of its **Quota**, then no withdrawals are possible until the beneficiary is updated or changed. Calls to **WithdrawBalance** will fail in this case with `illegal_state`.
```
anorth, line 52: Either USR_FORBIDDEN(18) or USR_INSUFFICIENT_FUNDS(19) would be better. Illegal state should not be used for things that can legitimately happen.

Please note explicitly that this behaviour is different to when the owner withdraws to themselves, where a withdrawal when 0 available will "succeed" to withdraw nothing (best effort).

anorth (commented):
I'm ok with this change.

I dunno what acceptable process is for changing an already-accepted FIP. The right time for this was the many months prior to acceptance, and during any of the two times this was initially implemented. But we shouldn't be too rigid either.

steven004:
I have no objection of this change, though I have a little (not strongly) different opinion. 
So, the reviewers, please feel free to approve it if you think it's fine. 

@anorth brought up a very good point about the change for an accepted proposal. We'd better to have one for future changes which might be needed for future FIPs. My suggestion is only for missed security issues or other major found mistakes.

jennijuju:
I feel like FIP authors should approve, or one or more core devs approve. this should not be a solely, or at all an fip editor decision.

jennijuju:
from @kaitlin-beegle in [slack](https://filecoinproject.slack.com/archives/C01EU76LPCJ/p1662433500804789?thread_ts=1662415392.474929&cid=C01EU76LPCJ)

> Great flag, thanks Jennifer.
> Is it correct to say that the change you seek to make is a technical change, discovered during implementation? Does the change substantively change the design, rationale, or problem statement of the FIP? I've reviewed, and it appears to be no. The proposed change necessarily takes into account a scenario which is probable but previously unaccounted for.  The change is necessary to properly implement the FIP.
> I also do not think this FIP needs an Active flag, since additional changes over time do not seem likely.
> My preference is that you amend the FIP in the repo, and we flag the change in the governance documentation (and potentially make a version note in the FIP itself). In the future, we will need more robust reviews to ensure these changes aren't needed, but I don't think we need a new FIP, etc. (e

steven004 (approved):

jennijuju (approved):

jennijuju:
merging as the author approved

PR #455, opened 2022-09-05
FIP-0045 updates removing opt-in migration

anorth:
- Specify `HasVerifiedClaims` field in `SectorOnChainInfo`
- Remove opt-in migration for existing sectors
- Add maximum expiration parameter for allocations
- Describe migration for pending deals

FYI @ZenGround0

### DIFF
--- a/FIPS/fip-0045.md
+++ b/FIPS/fip-0045.md
@@ -11,13 +11,10 @@ requires: FIP-0034
 ---
 
 TODO:
-- Re-spec ExtendClaimTerms as a token receiver hook value rather than explicit method
-- Spec methods for sector migration on miner, market, and verifreg actors
-- Spec change to SectorOnChainInfo to distinguish migrated sectors
-- Figure out best migration plan for pending deals, pre-committed sectors
-- Confirm policy for built-in market actor's default term maximum
+- Spec token receiver hook payload schema for allocation requests
+- Spec token receiver hook payload for extending allocations with new datacap
 - Market method for fetching allocation/claim ids for deals (needed?)
-- Spec hook/method for a different client extending the claim by spending DataCap (needed?).
+- Consider removing provision for re-commitment of same data in new sector (defer to follow-up)
 
 
 ## Simple Summary
@@ -83,7 +80,7 @@ and is much simpler than the Filecoin Plus premium proposal that also attempted 
 - *Claim*: a provider's assertion they are storing all or part of an allocation
 - *Term*: period of time for which a DataCap allocation or claim is valid or active.
 
-### Data cap allocations
+### Verified Registry DataCap allocations
 
 The Filecoin Plus verified registry actor is extended to record allocations of DataCap to specific
 pieces of data to be stored by a provider. An allocation is independent of any deal with the
@@ -114,14 +111,17 @@ struct Allocation {
 ```
 
 An allocation specifies a range of terms for which the provider may commit the data, between some
-minimum and maximum. An allocation's maximum term must be at least as large as its minimum term.
-
+minimum and maximum. 
+An allocation's maximum term must be at least as large as its minimum term.
 Due to the implementation of term enforcement (details below), clients should leave a large buffer
 between the minimum and maximum term to make the allocation practical for a provider.
+
+An allocation must expire, so that it cannot occupy non-useful DataCap indefinitely.
 
 **Parameters**
 - The minimum allowed value for `TermMinimum` is 6 months.
 - The maximum allowed value for `TermMaximum` is 5 years (can be increased by a future FIP).
+- The maximum difference between `Expiration` and the epoch at which an allocation is created is 60 days.
 
 Allocations are stored in the verified registry actor's state, grouped by client address.
 Nesting by client promotes more efficient lookups at scale, 
@@ -135,12 +135,12 @@ struct State {
     Allocations: HAMT[Address]HAMT[AllocationID]Allocation
 	
     // Sequence number for allocation IDs.
+    // The value `0` is reserved for "no allocation"
     NextAllocationId: AllocationID // uint64
 }
 ```
 
 #### Operations
-
 An allocation is created by a verified client by transferring DataCap tokens (see below)
 to the verified registry. 
 The `Allocation` metadata must be provided as transfer metadata.
@@ -169,7 +169,7 @@ The `RemoveVerifiedClientDataCap` method invokes a privileged method of the Data
 to burn DataCap tokens held on a client's balance. 
 Only the verified registry actor is permitted to invoke this method.
 
-### Data cap claims
+### Verified Registry DataCap claims
 
 The verified registry actor is further extended to record each storage provider's commitment to a
 DataCap allocation, called a claim. A claim represents a provider's obligation to store a piece of
@@ -258,24 +258,14 @@ struct RemoveExpiredClaimsParams {
 fn RemoveExpiredClaims(params: RemoveExpiredClaimsParams)
 ```
 
-
 The maximum term for a claim may be increased by the client which originally made the allocation,
 up to the network policy maximum.
-
-A verified client can extend the term for a claim _beyond_ the initial maximum term by 
-spending new DataCap. 
-The term maximum can be extended up to the network policy maximum beyond the current epoch.
-The client extending the claim need not be the one that made the original allocation.
-This is similar to issuing a new allocation/claim for the same data,
-but avoids overheads of re-sealing.
-
-TODO: rework this to be a token receiver hook rather than explicit method
 
 ```
 struct ClaimTerm {
     Provider: Address // needed to find claim in state
     ClaimID: ClaimID
-    TermMaximum: int64 // duration in epochs
+    TermMaximum: int64 // new duration in epochs from TermStart
 }
 struct ExtendClaimTermsParams{
     Claims: []ClaimTerm
@@ -293,6 +283,16 @@ struct ExtendClaimTermsResult {
 // and no greater than the maximum initial term.
 fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
 ```
+
+A verified client can extend the term for a claim _beyond_ the initial maximum term by
+spending new DataCap.
+The term maximum can be extended up to the network policy maximum beyond the current epoch.
+The client extending the claim need not be the one that made the original allocation.
+This is similar to issuing a new allocation/claim for the same data,
+but avoids overheads of re-sealing.
+
+TODO: spec token receiver hook for extending
+
 
 ### Data cap token
 The `VerifiedClients` map of addreses to balances is removed from the verified registry state.
@@ -325,16 +325,16 @@ The built-in account and multisig actors are changed to implement a receiver hoo
 This hook always succeeds, without inspecting the payload.
 
 #### Operations
-The `AddVerifiedClient` method is changed to invoke the DataCap token actor to
+The verified registry's `AddVerifiedClient` method is changed to invoke the DataCap token actor to
 mint new tokens to the client's balance.
 
-The `UseBytes` method is removed.
+The verified registry's `UseBytes` method is removed.
 To create an allocation, a verified client transfers DataCap tokens to the verified registry actor.
 The transfer must be accompanied by one or more `Allocation` records which specify 
 the allocation(s) to create.
 The registry may reject the transfer if the allocations are invalid.
 
-The `RestoreBytes` method is also removed.
+The verified registry's `RestoreBytes` method is also removed.
 Instead, the `RevokeAllocations` method on the verified registry actor transfers any released
 DataCap tokens back to the client's balance for re-use.
 
@@ -375,7 +375,30 @@ The pledge requirement for the old sector is not decreased,
 but any subsequent penalty is calculated using its current (i.e. reduced) power.
 
 ### Storage miner actor
-The built-in miner actor is trusted to enforce the verified and the continual storage of the data.
+#### Differentiated sectors
+A new field `HasVerifiedClaims: bool` is added to `SectorOnChainInfo`. 
+This is set `false` for all existing sectors during migration (below).
+
+A _new_ sector takes `HasVerifiedClaims = true` if it claims one or more verified allocations.
+Its `DealWeight` and `VerifiedDealWeight` will be a simple sum of the size of those claims.
+Its power is calculated in the simple way described above.
+
+A _new_ sector takes `HasVerifiedClaims = false` if it claims no verified allocations.
+
+A sector with `HasVerifiedClaims = false` and `VerifiedDealWeight = 0` 
+(i.e. a committed-capacity sector) takes `HasVerifiedClaims = true` if it claims 
+one or more verified allocations by `ReplicaUpdate`.
+Its new `DealWeight` and `VerifiedDealWeight` are calculated as a sum of the size of those claims.
+Its power is re-calculated in the simple way described above.
+
+A sector with `HasVerifiedClaims = false` and `VerifiedDealWeight > 0` continues to use
+the existing QA-power calculation.
+Such a sector cannot claim a verified allocation.
+Such a sector may be extended, but, as today, this will result in its power being diluted
+to "spread out" the verified reward over the sector's new commitment duration.
+
+#### Term enforcement
+The built-in miner actor enforces the verified term and the continual storage of the data.
 
 The miner actor requires that the sector into which a claimed piece is committed must have a
 **remaining lifespan that is sufficient to fulfil the minimum term**.
@@ -404,7 +427,7 @@ The miner actor must ensure that the old sector is no longer alive.
 
 #### Sector extension
 A sector's scheduled commitment term cannot be extended beyond the maximum term of any
-verified claim attributed to it.
+verified claim attributed to it (if `HasVerifiedClaims = true`).
 Thus, to extend a sector, a provider must simultaneously drop any verified claims with 
 terms that would be exceeded by the extended commitment.
 Dropping a verified claim reduces the sector's power accordingly, as well as
@@ -480,7 +503,7 @@ The built-in market specifies an allocation's minimum term to be equal to the de
 and maximum term to be some amount greater than the deal duration.
 
 **Parameters**
-- The built-in market calculates a term maximum of 1 year greater than a deal's specified duration.
+- The built-in market calculates a term maximum of 90 days greater than a deal's specified duration.
 
 ```
 let dealAllocation = Allocation {
@@ -490,10 +513,13 @@ let dealAllocation = Allocation {
   Size: deal.PieceSize
   TermMinimum: deal.EndEpoch - deal.StartEpoch
   TermMaximum: deal.EndEpoch - deal.StartEpoch + EPOCHS_IN_YEAR
-  Expiration: deal.StartEpoch
+  Expiration: min(deal.StartEpoch, CURRENT_EPOCH + 60 * EPOCHS_IN_DAY)
   AllowRanges: false
   }
 ```
+
+The market will accept verified deals with a `StartEpoch` beyond the maximum allowed allocation `Expiration`.
+If such a deal is not activated before the allocation expires, it will not confer QA power.
 
 The market stores allocation IDs for not-yet-activated deals in a new mapping, 
 and the claim ID for activated deals in deal state.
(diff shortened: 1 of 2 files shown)


FIPS/fip-0045.md
```diff
@@ -114,14 +111,17 @@ struct Allocation {
...
 
 An allocation specifies a range of terms for which the provider may commit the data, between some
-minimum and maximum. An allocation's maximum term must be at least as large as its minimum term.
-
+minimum and maximum. 
+An allocation's maximum term must be at least as large as its minimum term.
 Due to the implementation of term enforcement (details below), clients should leave a large buffer
 between the minimum and maximum term to make the allocation practical for a provider.
 
+An allocation must expire, so that it cannot occupy non-useful DataCap indefinitely.
```
jennijuju, line 119:
```suggestion
An allocation must have an expiration, so that it cannot occupy non-useful DataCap indefinitely.
```

FIPS/fip-0045.md
```diff
@@ -258,24 +258,14 @@ struct RemoveExpiredClaimsParams {
 fn RemoveExpiredClaims(params: RemoveExpiredClaimsParams)
 ```
 
-
 The maximum term for a claim may be increased by the client which originally made the allocation,
 up to the network policy maximum.
```
jennijuju, line 262: could you please specify whats the `network policy maximum` referred here?

FIPS/fip-0045.md
```diff
@@ -294,6 +284,16 @@ struct ExtendClaimTermsResult {
 fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
 ```
 
+A verified client can extend the term for a claim _beyond_ the initial maximum term by
```
jennijuju, line 287: a bit confused. above it says

> Called by allocation client to extend the maximum term of their allocations,  **without consuming DataCap**.

inconsistency maybe?

FIPS/fip-0045.md
```diff
@@ -294,6 +284,16 @@ struct ExtendClaimTermsResult {
...
+spending new DataCap.
+The term maximum can be extended up to the network policy maximum beyond the current epoch.
+The client extending the claim need not be the one that made the original allocation.
```
jennijuju, line 290: above states

_maximum term for a claim may be increased **by the client which originally made the allocation**,_

but here says need _**not** be the one that made the original allocation._

FIPS/fip-0045.md
```diff
@@ -294,6 +284,16 @@ struct ExtendClaimTermsResult {
 fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
```
jennijuju, line 284: what's `maximum initial term` ? like 5 years?

FIPS/fip-0045.md
jennijuju: from the context below it seems like we meant
```suggestion
A verified client can extend the term for a claim _beyond_ the initial maximum term without
```

FIPS/fip-0045.md
```diff
@@ -375,7 +375,30 @@ The pledge requirement for the old sector is not decreased,
 but any subsequent penalty is calculated using its current (i.e. reduced) power.
 
 ### Storage miner actor
-The built-in miner actor is trusted to enforce the verified and the continual storage of the data.
+#### Differentiated sectors
+A new field `HasVerifiedClaims: bool` is added to `SectorOnChainInfo`. 
+This is set `false` for all existing sectors during migration (below).
+
+A _new_ sector takes `HasVerifiedClaims = true` if it claims one or more verified allocations.
```
jennijuju, line 382:
```suggestion
A _new_ sector takes `HasVerifiedClaims = true` if it claims one or more allocations.
```

jennijuju: suggested cuz "verified allocation" was not an terminology used up until this point. (not a big deal tho)

FIPS/fip-0045.md
```diff
@@ -375,7 +375,30 @@ The pledge requirement for the old sector is not decreased,
...
+
+A _new_ sector takes `HasVerifiedClaims = false` if it claims no verified allocations.
+
+A sector with `HasVerifiedClaims = false` and `VerifiedDealWeight = 0` 
+(i.e. a committed-capacity sector) takes `HasVerifiedClaims = true` if it claims 
+one or more verified allocations by `ReplicaUpdate`.
+Its new `DealWeight` and `VerifiedDealWeight` are calculated as a sum of the size of those claims.
+Its power is re-calculated in the simple way described above.
+
+A sector with `HasVerifiedClaims = false` and `VerifiedDealWeight > 0` continues to use
```
jennijuju, line 394: will there ever exists a situation where a `HasVerifiedClaims` changes from _true_ to _false_, due to all claims in that sector reached term max gradually?

FIPS/fip-0045.md
```diff
@@ -375,7 +375,30 @@ The pledge requirement for the old sector is not decreased,
...
+the existing QA-power calculation.
+Such a sector cannot claim a verified allocation.
+Such a sector may be extended, but, as today, this will result in its power being diluted
+to "spread out" the verified reward over the sector's new commitment duration.
+
+#### Term enforcement
+The built-in miner actor enforces the verified term and the continual storage of the data.
```
jennijuju, line 401:
```suggestion
The built-in miner actor enforces the claim term and the continual storage of the data.
```

FIPS/fip-0045.md
```diff
@@ -404,7 +427,7 @@ The miner actor must ensure that the old sector is no longer alive.
 
 #### Sector extension
 A sector's scheduled commitment term cannot be extended beyond the maximum term of any
-verified claim attributed to it.
+verified claim attributed to it (if `HasVerifiedClaims = true`).
```
jennijuju, line 430: ```suggestion claim attributed to it (if `HasVerifiedClaims = true`).
```

File: FIPS/fip-0045.md [hunk 10]
```diff
@@ -490,11 +513,14 @@ let dealAllocation = Allocation {
   Size: deal.PieceSize
   TermMinimum: deal.EndEpoch - deal.StartEpoch
   TermMaximum: deal.EndEpoch - deal.StartEpoch + EPOCHS_IN_YEAR
```
jennijuju, line 515:
```suggestion
  TermMaximum: deal.EndEpoch - deal.StartEpoch + 90 * EPOCHS_IN_DAY
```

FIPS/fip-0045.md
ZenGround0: These comments are all good points but I think they're getting resolved by deleting `ExtendClaimTerms` which will happen when the following TODO is addressed

FIPS/fip-0045.md
```diff
@@ -375,7 +375,30 @@ The pledge requirement for the old sector is not decreased,
...
+This is set `false` for all existing sectors during migration (below).
+
+A _new_ sector takes `HasVerifiedClaims = true` if it claims one or more verified allocations.
+Its `DealWeight` and `VerifiedDealWeight` will be a simple sum of the size of those claims.
+Its power is calculated in the simple way described above.
+
+A _new_ sector takes `HasVerifiedClaims = false` if it claims no verified allocations.
+
+A sector with `HasVerifiedClaims = false` and `VerifiedDealWeight = 0` 
+(i.e. a committed-capacity sector) takes `HasVerifiedClaims = true` if it claims
```
ZenGround0, line 389: nit: these don't really exist but it could also be non verified deal sector

FIPS/fip-0045.md
```diff
@@ -490,11 +513,14 @@ let dealAllocation = Allocation {
   Size: deal.PieceSize
   TermMinimum: deal.EndEpoch - deal.StartEpoch
   TermMaximum: deal.EndEpoch - deal.StartEpoch + EPOCHS_IN_YEAR
-  Expiration: deal.StartEpoch
+  Expiration: min(deal.StartEpoch, CURRENT_EPOCH + 60 * EPOCHS_IN_DAY)
   AllowRanges: false
   }
 ```
 
+The market will accept verified deals with a `StartEpoch` beyond the maximum allowed allocation `Expiration`.
```
ZenGround0, line 521: So for miner integration: deal activation can pass and alloc claim can fail and in that case deal is still included, sector is still ProveCommitted but QA power is not added for the failed deal?

FIPS/fip-0045.md
```diff
@@ -526,43 +552,19 @@ The miner actor then invokes `ClaimAllocations` and, if successful,
 computes quality adjusted power for the sector according to the piece size and QA power multiplier.
 
 ### Migration
-#### Network upgrade
 A network state migration is required to implement this proposal.
+
 That migration must:
-- Implement the new verified registry state schema for allocations and claims.
+- Implement the new verified registry state schema for allocations and claims,
+  and set `NextAllocationId = 1`.
```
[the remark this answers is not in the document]
ZenGround0: [I think this](https://github.com/filecoin-project/FIPs/pull/455/files#diff-da7f03d0785adde81d03c9123f3d0cf50d8633ca6ac08d826a561239ca7dc66cR138)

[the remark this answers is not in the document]
ZenGround0: [And then used here](https://github.com/filecoin-project/FIPs/pull/455/files#diff-da7f03d0785adde81d03c9123f3d0cf50d8633ca6ac08d826a561239ca7dc66cR563)

FIPS/fip-0045.md
```diff
@@ -526,43 +552,19 @@ The miner actor then invokes `ClaimAllocations` and, if successful,
...
-After claims are extended, the provider can extend the sector commitment duration to match.
-Note that a migrated sector cannot be extended unless the claim terms are extended, 
-lest it violate a claim's maximum term.
-
-TODO: specify migration method and changes to SectorOnChainInfo
+- Add a new field `HasVerifiedClaim = false` to each `SectorOnChainInfo`.
+- Add the pending allocation mapping to market state, and claim ID field to deal states (all `0`)
+- For each pending (not yet activated) deal with `Verified = true`, create an Allocation
+  - The maximum term for such allocations is 540 days, in order to support any possible sector duration
+  - The expiration for such allocations is set to the maximum allowed value after the migration epoch,
```
ZenGround0, line 566: By this you mean [60 days](https://github.com/filecoin-project/FIPs/pull/455/files#diff-da7f03d0785adde81d03c9123f3d0cf50d8633ca6ac08d826a561239ca7dc66cR124)?

ZenGround0 (approved):

FIPS/fip-0045.md
ZenGround0: * in the case where prove commit happens after claim expiration but before deal start

FIPS/fip-0045.md
anorth: You're confused because there are two different mechanisms here:
1. The original client may increase term up to the maximum that could have been specified originally without spending any data cap
2. Any client can add an additional maximum term beyond the current epoch, by spending data cap

FIPS/fip-0045.md
anorth: That is the value currently specified above, yes.

FIPS/fip-0045.md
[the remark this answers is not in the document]
anorth: Answered above, 0 is reserved for a "no allocation" value

FIPS/fip-0045.md
anorth: Yes

FIPS/fip-0045.md
anorth: Yes

anorth:
I have pushed some updates:
- Clarifying some of the parameter values
- Clarifying the two different ways a claim term can be extended
- Replacing `HasVerifiedClaims` with `SimpleQAPower`, which is much simpler

FIPS/fip-0045.md
jennijuju: ah-ha!

FIPS/fip-0045.md
```diff
@@ -11,13 +11,9 @@ requires: FIP-0034
...
 
 TODO:
-- Re-spec ExtendClaimTerms as a token receiver hook value rather than explicit method
-- Spec methods for sector migration on miner, market, and verifreg actors
-- Spec change to SectorOnChainInfo to distinguish migrated sectors
-- Figure out best migration plan for pending deals, pre-committed sectors
-- Confirm policy for built-in market actor's default term maximum
+- Spec token receiver hook payload schema for allocation requests
+- Spec token receiver hook payload for extending allocations with new datacap
 - Market method for fetching allocation/claim ids for deals (needed?)
```
jennijuju, line 16: What's the potential use case here?

FIPS/fip-0045.md
```diff
@@ -135,18 +135,19 @@ struct State {
...
-
+#### Creation of allocations
 An allocation is created by a verified client by transferring DataCap tokens (see below)
 to the verified registry. 
 The `Allocation` metadata must be provided as transfer metadata.
 The verified registry checks the same preconditions as the existing `UseBytes` method. 
 The size of the allocation must match the number of tokens received.
 
+#### Removal of expired allocations
 An allocation may be removed after its expiration epoch has passed (by anyone).
```
misilva73, line 151: Why are we allowing expired allocations to be removed "by anyone"? What is the expected behavior here? Should clients remove them or are we supposed to have an automated process doing it?

FIPS/fip-0045.md
```diff
@@ -165,12 +166,12 @@ fn RemoveExpiredAllocations(params: RemoveExpiredAllocationsParams)
 
 Allocations are also removed when claimed by a provider (see below).
 
-The `RemoveVerifiedClientDataCap` method invokes a privileged method of the DataCap token actor
-to burn DataCap tokens held on a client's balance. 
+#### Revocation of un-allocated data cap
+The existing `RemoveVerifiedClientDataCap` method is changed to invoke a
```
misilva73, line 170: In what situations is this method called?

FIPS/fip-0045.md
```diff
@@ -245,7 +246,8 @@ struct SectorClaimResult {
 fn ClaimAllocations(params: ClaimAllocationsParams) -> ClaimAllocationsResult
 ```
 
-The record of a claim may be removed by the provider after the term maximum has elapsed.
+#### Removal of expired claims
+The record of a claim may be removed after the term maximum has elapsed.
```
misilva73, line 250: Who can remove expired claims? Is this an automated process?

FIPS/fip-0045.md
```diff
@@ -290,10 +282,21 @@ struct ExtendClaimTermsResult {
...
 fn ExtendClaimTerms(params: ExtendClaimTermsParams) -> ExtendClaimTermsResult
 ```
 
+#### Renewal of claim term by spending datacap
+A verified client can extend the term for a claim _beyond_ the initial maximum term by
+spending new DataCap.
+The claim's term maximum can be extended up to `MaximumVerifiedAllocationTerm` beyond the current epoch.
+The client extending the claim need not be the one that made the original allocation.
+This is similar to issuing a new allocation/claim for the same data,
+but avoids overheads of re-sealing.
```
misilva73, line 295: This mechanism makes a lot of sense. There is no reason to re-seal the same data just because one needs to spend new DataCap.

FIPS/fip-0045.md
```diff
@@ -114,14 +110,18 @@ struct Allocation {
...
 
+Network policy limits the maximum value for an allocation's expiration,
+so that an allocation cannot occupy non-useful DataCap indefinitely.
+
 **Parameters**
-- The minimum allowed value for `TermMinimum` is 6 months.
-- The maximum allowed value for `TermMaximum` is 5 years (can be increased by a future FIP).
+- `MinimumVerifiedAllocationTerm`: The minimum allowed value for `TermMinimum` is 6 months.
+- `MaximumVerifiedAllocationTerm`: The maximum allowed value for `TermMaximum` is 5 years (can be increased by a future FIP).
+- `MaximumVerifiedAllocationExpiration`: The maximum difference between `Expiration` and the epoch at which an allocation is created is 60 days.
```
misilva73, line 124: Why 60 days? Is this the current expiration term foe DataCap?

FIPS/fip-0045.md
```diff
@@ -375,7 +381,28 @@ The pledge requirement for the old sector is not decreased,
...
+- `false` otherwise
+
+All sectors committed after this proposal is activated take `SimpleQAPower = true`.
+The value of `SimpleQAPower` never changes for an individual sector.
+
+A sector with `SimpleQAPower = true` has its deal weights calculated as simple sums of deal size,
+and QA power calculated according to the simple method described above, with no spreading out effect.
+
+A sector with `SimpleQAPower = false` continues to use the existing QA-power calculation.
+Such a sector cannot claim any allocations.
```
misilva73, line 397: This is likely to result in increased expirations, with "old" sectors being replaced by new sectors that can take allocations. 

In addition, CC sectors can take advantage of this new way to package deals, while sectors with data have to wait for their deals to end. 

Could we have a mechanism to turn sectors with  `SimpleQAPower = false` into sectors with `SimpleQAPower = true`? For instance, if I have a 1-year sector with an old 1-year deal, there's no reason why I shouldn't be able to pack new deals into it, using the new logic and QAP calculation.

FIPS/fip-0045.md
```diff
@@ -480,7 +500,7 @@ The built-in market specifies an allocation's minimum term to be equal to the de
 and maximum term to be some amount greater than the deal duration.
 
 **Parameters**
-- The built-in market calculates a term maximum of 1 year greater than a deal's specified duration.
+- `MarketDefaultMaximumAllocationTerm`: The built-in market calculates a term maximum of 90 days greater than a deal's specified duration.
```
misilva73, line 503: Why 90 days?

misilva73:
It is not clear to me what happens to Allocations and DataCap tokens when there is a termination.

For instance, imagine there is a 2-year sector with a FIL+ deal that has a minimum term of 2 years. Then, after 1 year, the sector terminates. what happens here? Can the remaining 1-year DataCap allocation be used? Should DataCap tokens be given back to the client? Should a new allocation for 1 year be created?

FIPS/fip-0045.md
jennijuju: Iiuc, Anyone can remove the expired claims and it’s triggered by an onchain message.

FIPS/fip-0045.md
```diff
@@ -114,14 +110,18 @@ struct Allocation {
...
 between the minimum and maximum term to make the allocation practical for a provider.
 
+Network policy limits the maximum value for an allocation's expiration,
+so that an allocation cannot occupy non-useful DataCap indefinitely.
+
 **Parameters**
-- The minimum allowed value for `TermMinimum` is 6 months.
-- The maximum allowed value for `TermMaximum` is 5 years (can be increased by a future FIP).
+- `MinimumVerifiedAllocationTerm`: The minimum allowed value for `TermMinimum` is 6 months.
+- `MaximumVerifiedAllocationTerm`: The maximum allowed value for `TermMaximum` is 5 years (can be increased by a future FIP).
```
jennijuju, line 123:
```suggestion
- `MaximumVerifiedAllocationTerm`: The maximum allowed value for `TermMaximum` is 5 years (5 years is the current sector max lifetime, both may be increased by a future FIP).
```

kaitlin-beegle (approved):

FIPS/fip-0045.md
anorth: There is no reason to restrict it to the client address specifically, so this gives flexibility to users. E.g. if a client is a multisig or DAO or other complex entity, they don't need to go to the cost and trouble of multisigning this inconsequential message – one of the parties/delegates/employees can just do it.

There's no "should" or automation. Clients can clean up when the gas savings associated with a smaller collection justify it.

FIPS/fip-0045.md
anorth: Similar to above, anyone can. In practise this will probably be some ancillary or operations account associated with a provider. There's no automation.

FIPS/fip-0045.md
anorth: I chose 60 days as a reasonable balance between flexibility for providers scheduling deal-making, and the value being effective at reducing squatting. I'm very open to adjusting this value – please make a comment in the discussion thread if you would propose an alternative.

There's no current expiration, but it's worth introducing here because the changes make it practically a bit easier to squat on datacap (possible today, but more involved).

FIPS/fip-0045.md
anorth: It is impossible to change the data in a non-CC sector today _anyway_ (there is no "re-snap"), so there are no sectors to which this would apply. It will be easier to implement re-snapping after this change, though, but by the time we do this will be an even smaller issue.

There are [reasons](https://github.com/filecoin-project/FIPs/discussions/313#discussioncomment-3530843) not to do this. Perhaps for a narrow case of sectors where all the deals have already expired there would be fewer unwanted effects, but still this would cause a slight power loss for the SP, so is unlikely a major flow.

I agree that, for a capacity-constrained SP, a CC sector is more valuable than a sector with expired deals at the end of its life, due to the option of taking deals into it. But that's already the case today.

FIPS/fip-0045.md
anorth: Why not? This value was chosen as a compromise between flexibility for SP deal packing, and reasonably requesting the client's nominated term. The discussion thread for this is https://github.com/filecoin-project/FIPs/discussions/313#discussioncomment-3442637

anorth:
@misilva73 discussion moved to https://github.com/filecoin-project/FIPs/discussions/313#discussioncomment-3621462

FIPS/fip-0045.md
misilva73: Got it. This makes sense.

FIPS/fip-0045.md
misilva73: I don't have any specific value in mind. I was trying to understand the reasoning for the 60 days. It sounds reasonable to me as well.

FIPS/fip-0045.md
misilva73: Trying to understand the reasoning. I don't have any arguments against the 90 days.

PR #457, opened 2022-09-08
FIP-0045: spec the token receiver hook payload schema

anorth:
This complements #455, answering the remaining TODOs currently noted there.

### DIFF
--- a/FIPS/fip-0045.md
+++ b/FIPS/fip-0045.md
@@ -11,7 +11,6 @@ requires: FIP-0034
 ---
 
 TODO:
-- Re-spec ExtendClaimTerms as a token receiver hook value rather than explicit method
 - Spec methods for sector migration on miner, market, and verifreg actors
 - Spec change to SectorOnChainInfo to distinguish migrated sectors
 - Figure out best migration plan for pending deals, pre-committed sectors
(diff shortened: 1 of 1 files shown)


FIPS/fip-0045.md
```diff
@@ -598,25 +645,39 @@ can continue to use the existing `ExtendSectorExpiration` method.
...
+Since the QA power associated with a data cap allocation may be active for a longer period, 
+more of the network's total block reward will be directed toward data cap claim providers than 
+if the status quo were continued.
+
+Where an allocation is extended beyond 1.5 years, 
+this proposal will increase the incentive to the provider to extend the sector storing the associated data
+until the allocation's maximum term is reached.
+
+Since new allocations are more valuable than existing ones, this proposal may incentivise 
+deal-focussed but capital- or capacity-constrained providers to let existing sectors expires
```
misilva73, line 665:
```suggestion
deal-focussed but capital- or capacity-constrained providers to let existing sectors expire
```

FIPS/fip-0045.md
```diff
@@ -598,25 +645,39 @@ can continue to use the existing `ExtendSectorExpiration` method.
...
+in order to re-use the associated pledge and hardware for newer allocations.
+
+Storage providers may prefer deals associated with long data cap allocation terms to those with shorter expected terms.
```
misilva73, line 668: I think this will only happen if FIP36 is introduced. And, in that case, the pledge will also be higher for longer-duration sectors. Thus, SPs with lower capital or more risk averse will prefer shorter sectors (and shorter deals)

FIPS/fip-0045.md
```diff
@@ -653,18 +714,6 @@ and may be claimed by a collection of data commitments spread across multiple se
 Such an allocation would be far more gas efficient than the multiple per-sector deals
 required for a large data set today.
 
-### Support for FIP-0036
```
misilva73, line 656: Why is this section being removed? Is it no longer valid?

kaitlin-beegle (approved):

FIPS/fip-0045.md
anorth: Correct (mainly because the opt-in migration of existing sectors was removed).

PR #467, opened 2022-09-23
Moving FIP0044 and FIP0045 from "Last Call" to "Accepted".md

luckyparadise:
Moved FIP0044 and FIP0045 to "Accepted".


jennijuju (changes requested):
There is still a big todo question mark on the top. 
cc @ZenGround0 @anorth

anorth (approved):
This PR is still a correct reflection of the state.

jennijuju:
> This PR is still a correct reflection of the state.

Yes but this moves FIP status to accepted - the fip main body shall be locked by definition, no?

anorth:
> the fip main body shall be locked by definition

Where does it say that?

Look, we are FIP Editors, we make the "rules" (not that this one exists). The rules serve effective governance process, not the other way round. We are entrusted to use our best judgement when applying them. Getting stuck on tiny details like whether removing a TODO violates some idea of finality of a FIP is a waste of time.

jennijuju:
it’s not about removing the todos. Neither in any PRs or FIP discussion it was said we will NOT expose this method. as an fip editor - I’m not sure if the todo were going to be included or not, if it does - will it  change any state structure? Any breaking actor API changes? And new API? In this case, potentially not. However, the answer is unclear to me, and I’m “calling it out as I see it” given we are in LAST CALL stage of the FIP process.

jennijuju:
now you, as the author, knows the currently state of the FIP implementation and if you can confirm that TODO wont change anything for what’s described in the FIP, please simply remove it. Then LAST CALL -> accepted shall be unblocked.

anorth:
#471

jennijuju (approved):

Issue #450, opened 2022-09-02
FIP-0044: Verifreg: Consider modifying RemoveDataCapProposal


arajasek:
This is one of the 2 remaining uses of the `validate_signature` syscall in builtin actors code. There is an argument to be made that doing so allows programmatic verified clients, and an overall cleanliness argument. This is of limited consequence, but I'd like to incorporate it into FIP-0044 if people are generally favourable.

dkkapur:
@arajasek would love to understand this more, what should I look at / read to get a better handle on consequences of a new design?

arajasek:
@dkkapur Sorry about the very delayed response here!

Short answer: no consequences, it isn't a new design at all.

Long answer: FIP-0044 adds a new way for actors to validate signatures. In addition to the `validate_signature` syscall (which is [what the verifreg uses today](https://github.com/filecoin-project/builtin-actors/blob/f63b80c00d548be0c3d7bf9cbb17cd5576bef702/actors/verifreg/src/lib.rs#L913)), there is a new method called `AuthenticateMessage` that actors can choose to implement -- currently only the account actor does so.

The verified registry can thus either continue to call `validate_signature`, or instead start invoking the new `AuthenticateMessage` method. The difference is immaterial today, except that calls to `RemoveVerifiedClientDataCap` might cost slightly more gas. The long-term impact is that with the FVM, it would become possible for this method to be called by contracts (thus, it would be one step towards having programmatic verifiers if we so wanted).

Truthfully, my motivation is more around code cleanliness -- I'd rather unify around the "new" way to authenticate signatures than have two. Should be no impact at all to the UX of verifiers.

dkkapur:
@arajasek thanks! sounds like we don't need to change anything on the tooling front today and can hopefully leverage this with FVM easier in the future! appreciate the explanation!

PR #470, opened 2022-09-29
FIP-0044: Specify change to VerifiedRegistryActor

arajasek:
I opened this draft just to have a companion PR to #450, but I don't think we should actually land this:

- We already have a PR moving FIP-0044 to Accepted, and any change at this stage (even a trivial one) is bad
- @ZenGround0 points out that there really isn't gonna be any utility to this change for at least a year
- The implementation may not actually be _completely_ trivial, at which point it's definitely not worth it

### DIFF
--- a/FIPS/fip-0044.md
+++ b/FIPS/fip-0044.md
@@ -60,7 +60,9 @@ pub struct AuthenticateMessageParams {
 }
 ```
 
-Further, we propose modifying the StorageMarketActor to call this method when validating `ClientDealProposal`s instead of validating signatures directly.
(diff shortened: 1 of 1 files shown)


anorth:
I concur with deferring this to a future FIP.

jennijuju:
@arajasek yah seems like this will have to be a future fip as 44 has been Accpeted already

kaitlin-beegle:
@arajasek are you comfortable with me closing this PR for now?

arajasek:
Yes! Closing :)

PR #483, opened 2022-10-13
FIP-0049 draft: Actor events

raulk:
This FIP adds the ability for actors to emit externally observable events during execution. It is proposed for nv18, and is necessary for the Filecoin EVM runtime. It relates to other FIPs that will be submitted in the next weeks.

Events are fire-and-forget, and can be thought of as a side effect from execution. The payloads of events are self-describing objects signalling that some important circumstance, action or transition has ocurred.

Actor events enable external agents to observe the activity on chain, and may eventually become the source of internal message triggers.

This FIP introduces a _minimal_ design for actor events, including their schema, a new syscall, a new field in the `MessageReceipt` structure, and mechanics to commit these execution side effects on the chain. By _minimal_ we mean that the protocol stays largely unopinionated and unprescriptive around higher level concerns like event indexing, traversal, and proofs (e.g. inclusion proofs for light clients), and other future higher-level features.

---

Closes https://github.com/filecoin-project/ref-fvm/issues/728.

### DIFF
--- a/FIPS/fip-nnnn-actor-events.md
+++ b/FIPS/fip-nnnn-actor-events.md
@@ -1,0 +1,331 @@
+---
+fip: <to be assigned>
+title: Actor events
+author: Raúl Kripalani (@raulk), Steven Allen (@stebalien)
+discussions-to: <URL>
+status: Draft
+type: Technical Core
+category: Core
+created: 2022-10-12
+spec-sections: 
+  - https://spec.filecoin.io/#section-systems.filecoin_vm.runtime.receipts
+requires: N/A
+replaces: N/A
+---
+
+## Simple Summary
+
+This FIP adds the ability for actors to emit externally observable events during
+execution. Events are fire-and-forget, and can be thought of as a side effect
+from execution. The payloads of events are self-describing objects signalling
+that some important circumstance, action or transition has ocurred. Actor events
+enable external agents to observe the activity on chain, and may eventually
+become the source of internal message triggers.
+
+## Abstract
+
+This FIP introduces a _minimal_ design for actor events, including their schema,
+a new syscall, a new field in the `MessageReceipt` structure, and mechanics to
+commit these execution side effects on the chain. By _minimal_ we mean that the
+protocol stays largely unopinionated and unprescriptive around higher level
+concerns like event indexing, traversal, and proofs (e.g. inclusion proofs for
+light clients), and other future higher-level features.
+
+## Change Motivation
+
+Two main motivations warrant the introduction of actor events at this time.
+
+1. They have been long missing from the protocol, complicating observability and
+   forcing Filecoin network monitoring or accounting tools resort to creative
+   approaches to obtain the data they needed, e.g. forking and instrumenting
+   built-in actors code, or replaying messages and performing state tree diffs.
+   
+   While this may have been tolerable for a limited number of actors, it will no
+   longer scale for a world of user-defined actors.
+   
+   A future FIP could enhance built-in actors to emit events when power changes,
+   sectors are onboarded, deals are made, sectors are terminated, penalties are
+   incurred, etc.
+
+2. The upcoming introduction of the Filecoin EVM runtime necessitates an event
+   mechanism at the protocol layer to handle the LOG{0..4} opcodes, and the
+   corresponding Ethereum JSON-RPC methods.
+
+## Specification
+
+### New chain types
+
+We introduce a DAG-CBOR encoded `Event` type to represent an actor event.
+
+```rust
+/// Represents an event emitted throughout message execution.
+struct Event {
+    /// Carries the ID of the actor that emitted this event.
+    emitter: ActorID,
+    /// Key values making up this event.
+    entries: Vec<Entry>,
+}
+
+struct Entry {
+    /// A bitmap conveying metadata or hints about this entry.
+    flags: u8,
+    /// The key of this event.
+    key: String,
+    /// Any DAG-CBOR encodeable type.
+    value: any,
+}
+```
+
+Its main constituent is a list of ordered key-value **entries**. This approach
+is inspired by structured logging concepts. **Keys** are strings, while
+**values** can be any DAG-CBOR encodeable type.
+
+Every **entry** contains a single-byte **flags** bitmap, which conveys metadata
+or hints about the entry. These are the currently supported flags:
+
+| hex    | binary      | meaning |
+|--------|-------------|---------|
+| `0x01` | `b00000001` | indexed |
+
+Flag `0x80` is currently only used as a client hint. In the future, this flag
+and others may lead to consensus-relevant outcomes, such as populating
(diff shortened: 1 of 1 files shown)


FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
...
+
+1. They have been long missing from the protocol, complicating observability and
+   forcing Filecoin network monitoring or accounting tools resort to creative
+   approaches to obtain the data they needed, e.g. forking and instrumenting
+   built-in actors code, or replaying messages and performing state tree diffs.
+   
+   While this may have been tolerable for a limited number of actors, it will no
+   longer scale for a world of user-defined actors.
+   
+   A future FIP could enhance built-in actors to emit events when power changes,
```
ZenGround0, line 46: 💯 This will solve a series of long standing problems.

FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
...
+
+Note that there's a timing difference between the moment the client receives the
+events AMT root CID (after message execution), and the moment that the event
+data (IPLD blocks) is effectively made available in the client's blockstore (at
+Machine flush time). Consequently, events cannot be viewed/logged/inspected
+until the Machine is flushed.
+
+FVM implementations may accept a separate events store, and offer an "early
+events flush" Machine option to instruct the FVM to flush events on every
+message execution.
```
anorth, line 204: I don't think this section is part of the spec, but implementation advice.

FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
...
+Concerning the concept of `flags`, we had initially considered a top-level
+`indexed` bitmap, but later generalised it to a `flags` field for extensibility.
+We consider the consequent size tradeoff to be acceptable.
+
+## Backwards Compatibility
+
+We are adding a new field to the `MessageReceipt` type. While this type is not
+serialized over the wire nor gossiped, it is returned from JSON-RPC calls.
+Hence, users of Filecoin clients may notice a new field, which could in turn
+break some of their code.
```
anorth, line 303: I think it's worth noting that this is a consensus-breaking change, and clients _must_ upgrade to the new behaviour in order to sync the chain.

FIPS/fip-nnnn-actor-events.md
raulk: Agree. One of these problems is introspection by tools like Lily, accounting tools, and more. The lack of events forces these tools to rely on statediffing, which isn't a cheap process. There's a lot of overhead we can remove by emitting events from builtin-actors, and migrating observability tools to consume those. cc @frrist

FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
...
+
+2. The upcoming introduction of the Filecoin EVM runtime necessitates an event
+   mechanism at the protocol layer to handle the LOG{0..4} opcodes, and the
+   corresponding Ethereum JSON-RPC methods.
+
+## Specification
+
+### New chain types
+
+We introduce a DAG-CBOR encoded `Event` type to represent an actor event.
```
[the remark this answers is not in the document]
raulk: The intention with mandating DAG-CBOR (and having FVM validate the input) is to make events eventually traversable to their constituents parts for the benefit of future light clients and similar scenarios.

I chose to mandate DAG-CBOR because it's worth:
1. To model CID links univocally (although having tag 42 [officially recognised](https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml) as CIDs already achieves this without DAG-CBOR constraints)
2. To exclude other tags since there is no guarantee that future readers of events will understand them.
3. To guarantee minimal encodings to prevent potential attack vectors.

FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
...
+**values** can be any DAG-CBOR encodeable type.
+
+Every **entry** contains a single-byte **flags** bitmap, which conveys metadata
+or hints about the entry. These are the currently supported flags:
+
+| hex    | binary      | meaning |
+|--------|-------------|---------|
+| `0x01` | `b00000001` | indexed |
+
+Flag `0x80` is currently only used as a client hint. In the future, this flag
```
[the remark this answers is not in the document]
raulk: Yeah, _currently_ is the wrong word here. I will clarify.

FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
...
+(adaptable) bloom filters, committing block-level indices, supporting on-chain
+event subscriptions, and more.
+
+Future flags may include:
+- specialized indexing strategies
+- bloom inclusion
+- special type indicators (e.g. "this is an address")
+
+> TODO: make flags a varint for extensibility? Can also retype later, if we
+> overflow.
```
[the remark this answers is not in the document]
raulk: How is that? CBOR integer encoding is variable length, and DAG-CBOR requires shortest encoding possible for integers, so we definitely use the variability.

FIPS/fip-nnnn-actor-events.md
frrist: In addition to reducing (and nearly eliminating) the need to perform state-diffing, these events will serve as a definitive source of truth for activities occurring on-chain, and expose information that is currently (near-)impossible to collect, such as sector termination fees, sector expirations, etc.
We'd be happy to provide feedback/participate in writing such a FIP when the times comes.

FIPS/fip-nnnn-actor-events.md
frrist, line 204: In the vein of implementation advice, and perhaps this is what is meant by the "early events flush" operation: Could FVM implementations choose to expose events in their execution response? ([related code](https://github.com/filecoin-project/filecoin-ffi/blob/20f104e88065aae59fd212d64be5bed108604e78/fvm.go#L129-L141)).

FIPS/fip-nnnn-actor-events.md
```diff
@@ -0,0 +1,331 @@
+---
+fip: <to be assigned>
```
jennijuju, line 2: this will be FIP-0049

kaitlin-beegle (approved):

kaitlin-beegle:
FYI @raulk - going to merge.  Looks like nits and relevant feedback have been incorporated as issues, many of which have been resolved.  Want to merge the draft in before additional refinements come in the following days/weeks.

FIPS/fip-nnnn-actor-events.md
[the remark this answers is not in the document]
raulk: Coming back to this. CBOR integers are variable length encoded, and DAG-CBOR mandates minimal encoding for integers. However, at the programming level we represent integers as u64 anyway, so we can use u64 here and benefit from minimal varint encoding while we stay below decimal 24, which in bitflags would cover 4 bits (we currently only use 2).

PR #485, opened 2022-10-13
fix: Amend maximum term for migrated verified deals

geoff-vball:
Changing the FIP document to accurately reflect what has been implemented in `builtin-actors`.


anorth:
I've raised a question about this in FIL Slack, to resolve before we merge this.

geoff-vball:
Discussed with @arajasek and @ZenGround0. We have decided that `540d` is the best option for `TermMax` so we will keep the FIP as-is. Conversation here https://filecoinproject.slack.com/archives/C015KQQLQQ1/p1665970007296099

PR #510, opened 2022-11-08
Update fip-0047 

jennijuju:


anorth (approved):

anorth:
Could you also fix the template to include an example with quotes, so people are more likely not to fall for this?

PR #523, opened 2022-11-14
Add gas modelling numbers to FIP-0045

geoff-vball:
I've added gas modelling numbers for all the calls affected by this FIP.

I want to flag that the increased execution cost for each verified deal is ~50M gas. For Prove Commit, this 50M increase is going to executed in cron. We already have issues with bounding cron execution. Would it make sense to add an addition fixed gas cost to prove commits with verified deals? I don't think the marginal cost increase is too big an issue to counteract to network issues that might arise from cron bloating even more.

It's probably too late to do anything about this for nv17, but something we may want to consider in a future network upgrade.

### DIFF
--- a/FIPS/fip-0045.md
+++ b/FIPS/fip-0045.md
@@ -779,7 +779,58 @@ the final decoupling from the miner actor.
   including those far more efficient than the built-in one.
 
 By opening the field of competition for market-like actors, we can expect deal costs to decrease over time,
-while the network enforces just enough state to administer the quality-adjusted power of verified data. 
+while the network enforces just enough state to administer the quality-adjusted power of verified data.
+
+#### Publish Storage Deals
+Publish storage deals will see a large increase in gas usage in this FIP. This will be drastically reduced in the future by:
+
+- Aggregating datacap token transfers to the verifreg actor 
+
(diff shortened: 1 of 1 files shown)


arajasek (approved):

anorth:
Thanks.

The flow of https://github.com/filecoin-project/FIPs/discussions/298 will move deal activation always out of cron anyway. (We should still look to optimise it, of course).

FIPS/fip-0045.md
```diff
@@ -779,7 +779,58 @@ the final decoupling from the miner actor.
   including those far more efficient than the built-in one.
 
 By opening the field of competition for market-like actors, we can expect deal costs to decrease over time,
-while the network enforces just enough state to administer the quality-adjusted power of verified data. 
+while the network enforces just enough state to administer the quality-adjusted power of verified data.
+
+#### Publish Storage Deals
+Publish storage deals will see a large increase in gas usage in this FIP. This will be drastically reduced in the future by:
```
jennijuju, line 785:
```suggestion
`PublishStorageDeals` messages that include verified deals will see a large increase in gas usage in this FIP. This will be drastically reduced in the future by:
```

PR #527, opened 2022-11-17
FRC-0051: Create FIP: Adding (Synchronous) Consistent Broadcast to EC

guy-goren:
Related discussions at https://github.com/filecoin-project/FIPs/discussions/501


FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
+---
+fip: "00xx"
+title: mproving EC security with Consistent Broadcast
```
anorth, line 3:
```suggestion
title: Improving EC security with Consistent Broadcast
```

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+A [security analysis](https://drive.google.com/file/d/1KFykwDHQxsgbOApbLKDZL6ViOQBQoQBf/view?usp=sharing) shows that, given the current "best effort" broadcast combined with the cutoff time, an attacker with $>20\%$ of the power (storage) can break the security of Filecoin. The proposed improvement essentially replaces the *best effort* broadcast (*beb*) with a variant of *consistent* broadcast (cb). The difference is, in short, that *beb* broadcast allows two honest miners to deliver different blocks published by an equivocating Byzantine miner in the same epoch,  which *cb* does not. The result will be that an attacker [will require $>44\%$](https://hackmd.io/@consensuslab/ByltVcfAc) of the total power (adjusted storage) to successfully mount this particular attack.
```
anorth, line 26: Please check in both this security analysis file and the document linked from "will require >44%", in this PR, to the `resources/fip-xxxx` directory (renaming when given a FIP number). Links to external resources will rot.

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Design Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+The design goal is to introduce a simple change without severe repercussion to other functionalities of the systems. This means approaching the core of the matter (the mechanism that enable equivocation). This allows for orthogonal changes in the incentive design mechanism, since the change does not tie them together.
+### Potential strategies
+The ConsensusLab team explored potential mitigation strategies and narrowed down the list for the following possible strategies: (for more details see [this document](https://hackmd.io/@consensuslab/HyaNMcMAq))
```
anorth, line 52: Please include the linked document in this PR, in `resources`.

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Removing tipsets from EC, effectivelly boiling down the protocol to a longest-chain instead of a longest-tipset protocol. By removing tipsets and having a traditional blockchain design (i.e., one block per epoch) the problem is mitigated.
+
+    :+1: Simple to implement -- changing the expected number of blocks from 5 to 1.
```
anorth, line 56: I think discussion elsewhere has agreed that this is not simple to implement. Changing the _expected_ number of blocks doesn't remove tipsets.

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Backwards Compatibility
+The proposal does not introduce backward compatibility issues. A miner that adopted the change will still accept blocks from honest non-adopting miners, and non-adopting miners will still accept blocks from adopting miners. The systems' security will improve monotonically with the share of adopting miners.
```
anorth, line 92:
```suggestion
The proposal does not introduce backward compatibility issues. A miner that adopted the change will still accept blocks from honest non-adopting miners, and non-adopting miners will still accept blocks from adopting miners. The system's security will improve monotonically with the share of adopting miners.
```

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Incentive Considerations
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
+The proposal is intentionally decoupled from incentive considerations. It is neither dependent on them, nor does it limit changes to incentive mechanisms.
+
+A question may arise: isn't equivocation slashable in Filecoin, and therfore should not concern us?
+Equivocating is indeed slashable, however, there are 2 issues regarding slashing as a solution.
+**tl;dr: 1) conceptually, it is dangerous to rely on rationality that is based only on internal incentives. 2) Slashing/Blacklisting does not solve the problem.**
```
anorth, line 118: Delete this line, it's not adding anything

anorth (commented):
Regarding the title, I think we should generally avoid value judgements in the proposal title and aim to be strictly declarative, so "Improving" is a problem here. Could we call it something like "Synchronous Consistent Block Broadcast" or similar?

FIPS/fip-nnn.md
guy-goren: The current writing is indeed confusing. I corrected it to clarify that it corresponds to removing tipsets in expectation, which will already mitigate the problem. 
Nevertheless, changing the validity rule to actually remove tipsets completely (by having a single parent block) should also be rather straight forward to implement. It does not require that a single block is created per epoch. However, such a change will have dramatic side effects (e.g., slower progress and worse security parameters convergence), not to mention economic effects on the miners (increased variance in rewards). Therefore, we do not recommend it.

FIPS/fip-nnn.md
guy-goren: I initially thought the same as you, that this is unnecessary and might only confuse people. But after being asked the same question again and again by different stakeholders, I tend to think it is worthwhile to have it (although it is redundant).

FIPS/fip-nnn.md
guy-goren: Done

FIPS/fip-nnn.md
guy-goren: Done

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+    
+    :+1: Does not require a fork.
+    
+    :-1: Depends on timing assumptions. That is, gossip propagation within 6 seconds.
+    
+---
+    
+**The ConsensusLab research team recommends the use of synchronous consistent broadcast as the mitigation strategy** (i.e., the last of the three proposed mitigations). The benefits of easy adoption due to only "behind the scence changes" and no <!--fork-->Filecoin-upgrade required, outweigh the shortcomings (timing assumptions). Moreover, even if the timing assumptions are violated, the consistency degrades gracefully (that is, a slightly longer than expected gossip delay would lead to only a small amount of equivocating blocks being accepted rather than to a "cliff phenomenon" where all equivocating blocks are accepted). Also, using the synchronous approach does not require significant changes to the [underlying broadcast protocol](https://blog.ipfs.tech/gossipsubv1.1-eval-report-and-security-audit/), which is already well tested and has built-in failure detecting mechnisms.
```
jennijuju, line 89: I think it is excellent that the FIP authors proposed multiple alternative solutions for the community to consider in the FIP discussion. However, i personally think a FIP draft should have a more concrete proposal. Therefore - i'd like to suggest  instead of "recommends the use of synchronous consistent broadcast as the mitigation strategy", the authors include  "considering the trade-off, CL would like to propose...."

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
+---
+fip: "00xx"
```
jennijuju, line 2:
```suggestion
fip: "0051"
```

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+title: mproving EC security with Consistent Broadcast
+author:  Guy Goren <guy.goren@protocol.ai>, Alfonso de la Rocha <alfonso@protocol.ai>
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/501
+status: Draft
+type: Technical
```
jennijuju, line 7:
```suggestion
type: FRC
```

see reasoning [here](https://github.com/filecoin-project/FIPs/pull/527#issuecomment-1322946492) feedback welcome

FIPS/fip-nnn.md
guy-goren: Done

FIPS/fip-nnn.md
guy-goren: Indeed, one reason for choosing the _synchronous_ CBcast option was the fact that it is not "consensus breaking" (does not cause a fork even if not all miners adopt it -- in classic blockchain terminology). Hence, it can also fit as an FRC.
What is important to notice is that the security of the entire network improves when more SPs adopt it. Therefore, I suggest that even if it goes to the FRC track, it would be implemented in the next network upgrade in order to boost adoption.

jsoares: I don't have strong feelings but would like to clarify this in principle. Beyond the inconsistency addressed in https://github.com/filecoin-project/FIPs/pull/552, there's the fact that FRC, by analogy to Ethereum, has an application-layer connotation (which is also true of the existing FRCs).

We all agree this isn't consensus breaking. The questions are:
1. It's clear that any consensus-breaking change requires a non-informational FIP. I don't read the language in FIP-0001 as requiring that all FIPs be consensus-breaking.
2. An informational/FRC FIP, again per FIP-0001, "does not propose a new feature" and "do[es] not necessarily represent ... a recommendation". I would claim neither is true here.

Provided that we're consistent about this, I don't see an issue with it being an FRC. But, as Guy said, the security of the network is improved by a large majority of nodes implementing CB, and so it would be preferable to attain community consensus on the direction even if it's fully interoperable.

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,132 @@
...
+By mitigating attack on Filecoin, the proposal improves the resilience of the system. The proposal does not require any operational changes (besides adopting the protocol update) from the users.
+
+## Implementation
+<!--The implementations must be completed before any core FIP is given status "Final", but it need not be completed before the FIP is accepted. While there is merit to the approach of reaching consensus on the specification and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.-->
+The prototype implementation lives in the following repo: 
+- [Upgraded version of Lotus with consistent broadcast](https://github.com/adlrocha/lotus/tree/adlrocha/consistent-bcast): Implementation of the synchronous consistent broadcast fix over Lotus.
```
jsoares, line 129:
```suggestion
[filecoin-project/lotus#9858](https://github.com/filecoin-project/lotus/pull/9858) brings consistent broadcast, as specified above, to lotus master
```

jsoares: @jennijuju: as promised, clean PR in https://github.com/filecoin-project/lotus/pull/9858, ready to review by your team along with the proposal 🙏

FIPS/fip-nnn.md
```diff
@@ -0,0 +1,131 @@
+---
+fip: "0051"
+title: mproving EC security with Consistent Broadcast
```
jennijuju, line 3:
```suggestion
title: Improving EC security with Consistent Broadcast
```

jennijuju (approved):

FIPS/fip-nnn.md
jennijuju, line 3:
```suggestion
title: Synchronous Consistent Block Broadcast for EC security
```

jennijuju (approved):

jennijuju:
I am merging this draft as it has been ready for a long time. I still think it is an FRC atm.

@kaitlin-beegle we shall update FIP-0001 soon with better definiation bettwen frc and fips. If needed, we can update the type of this one later. 

@ adlrocha Thank you for the lotus impl 💙 
@guy-goren @jsoaresAFAIK, the team is still doing more analysis with more aggressive parameters than 6 sec - once the team confirms the value, please update the FRC, and/or let the FIP editors know  whenever its ready for last call!

PR #542, opened 2022-11-30
FIP0052: Multicall as a built-in actor

xBalbinus:
The Multicall aggregates results from multiple contract constant function calls to reduce the number of separate JSON RPC requests that need to be sent, while also providing the guarantee that all values returned are from the same block (like an atomic read) and returning the block number the values are from. Multicall is also intended to be used with Multicall.js, a lightweight JavaScript library, in front-end dapps. This pull request is a proposal to implement the Multicall contract as a built-in actor (perhaps similar to a precompiled contract on native ETH). The utility Multicall offers can greatly improve developer experience on both the frontend and contracts side of FVM.

### DIFF
--- a/FIPS/fip-0052.md
+++ b/FIPS/fip-0052.md
@@ -1,0 +1,112 @@
+---
+fip: "0052"
+title: Change type of DealProposal Label field from a (Golang) String to a Union
+author: Xiangan He (@xBalbinus), Alex Miller (@amiller68)
+discussions-to: https://github.com/filecoin-project/FIPs/issues/187
+Status: Draft
+type: Technical
+category: Core
+created: 2021-09-29
+spec-sections: 
+  - specs-actors
+---
+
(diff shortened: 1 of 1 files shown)


kaitlin-beegle:
I reached out to the FIP author to clarify the intent of this FIP.  Since the current pushed draft is a 1:1 copy of FIP0027, I'm going to close for now. 

@xBalbinus, FYI, the first step in authoring a FIP is to open your draft in the discussion forum.  If you have any questions or need any assistance, please let me know!

xBalbinus:
No problem at all! Thanks so much, I was unsure as to what the process was
for proposing an FIP and was misinformed. I will open up this discussion in
the forums.

anorth:
Please do open that discussion, I can add support here

xBalbinus:
Discussion opened: https://github.com/filecoin-project/FIPs/discussions/560, revisions / additions to come based on further comments

PR #566, opened 2022-12-11
Add market notify deal to FIP 50 'other changes'

ZenGround0:
Discussed here #549 
Prototyped here: https://github.com/filecoin-project/builtin-actors/pull/894/files

This change is small enough and aligned with the other FIP 50 changes that I think it belong here without lots of explanation.  It's up for discussion whether this should be its own FIP or if it needs more description.

### DIFF
--- a/FIPS/fip-0050.md
+++ b/FIPS/fip-0050.md
@@ -206,7 +206,7 @@ We propose making some minor changes to built-in actor functionality that align 
   - Callers of Miner::DisputeWindowedPost and Miner::ReportConsensusFault
   - Callers of Multisig::Propose, Multisig::Approve, and Multisig::Cancel
   - Callers of Power::CreateMiner
-- Modify MarketActor::PublichStorageDeals to call out to exported method `MarketNotifyDeal` for all deal clients
+- Modify `MarketActor::PublishStorageDeals` to call out to exported method `MarketNotifyDeal` for all deal clients
   - Parameters are deal ID and serialized proposal
   - Errors are unchecked so client actor need not implement MarketNotifyDeal
 
--- a/FIPS/fip-0050.md
+++ b/FIPS/fip-0050.md
@@ -208,7 +208,7 @@ We propose making some minor changes to built-in actor functionality that align 
   - Callers of Power::CreateMiner
 - Modify `MarketActor::PublishStorageDeals` to call out to exported method `MarketNotifyDeal` for all deal clients
   - Parameters are deal ID and serialized proposal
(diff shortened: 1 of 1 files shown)


FIPS/fip-0050.md
```diff
@@ -206,6 +206,9 @@ We propose making some minor changes to built-in actor functionality that align
   - Callers of Miner::DisputeWindowedPost and Miner::ReportConsensusFault
   - Callers of Multisig::Propose, Multisig::Approve, and Multisig::Cancel
   - Callers of Power::CreateMiner
+- Modify MarketActor::PublichStorageDeals to call out to exported method `MarketNotifyDeal` for all deal clients
```
arajasek, line 209:
```suggestion
- Modify `MarketActor::PublishStorageDeals` to call out to exported method `MarketNotifyDeal` for all deal clients
```

FIPS/fip-0050.md
```diff
@@ -206,6 +206,9 @@ We propose making some minor changes to built-in actor functionality that align
...
+  - Parameters are deal ID and serialized proposal
+  - Errors are unchecked so client actor need not implement MarketNotifyDeal
```
arajasek, line 211:
```suggestion
  - Errors are unchecked so client actor need not implement `MarketNotifyDeal`
```

arajasek (approved):

jennijuju (changes requested):
Im not sure adding this api can go into this FIP, cuz unlike other apis, this one introduce behaviour changes into existing method, and may add on additional gas to the existing message.

requested change to block merge before we align on whether this needs a new fip (its gonna be a small one)

ZenGround0:
This is a great point however I believe the gas change is small enough for this to be in scope:

The latest implementation of MarketNotifyDeal calling incurs minimal new compute from the PSD side when the proposal client is an account actor (the common case we want to avoid breaking).  It reuses the serialized results from the existing serialization for authenticate message and does a single send per client.  Sends are ~30k gas at syscall level.  Assuming 100 million gas for PSD that's less than 0.1 % of current gas costs.  

In the account case the client account actor will have to do a match and throw an error since the Account actor doesn't implement MarketNotifyDeal.  We can measure this but it should be quite small.  The only other thing from the account side is the new `restrict_internal_api` call which is also a new gas cost incurred by *all* methods.

ZenGround0:
In short I will measure in a devnet and report back but don't think it's worth holding up this merge.

jennijuju:
I’m sold

jennijuju (approved):

PR #571, opened 2022-12-14
SDM Draft for Community Consideration

vkalghatgi:
- See Earlier Github [discussion](https://github.com/filecoin-project/FIPs/discussions/554)


FIPS/fip-00XX.md
```diff
@@ -0,0 +1,199 @@
...
+title: "Sector Duration Multiplier (SDM)"
+status: Draft
+type: Technical
+author: Axel C (@AxCortesCubero), Juan B (@jbenet), Maria S (@misilva73), Molly M (@momack2), Tom M (@tmellan), Vik K (@vkalghatgi), ZX @zixuanzh) 
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/554
+created: 2022-12-14
+---
+
+## Simple Summary
+- A Sector Duration Multiplier (SDM) is introduced for all sectors, including Committed Capacity (CC) sectors and sectors containing storage deals.
```
anorth, line 12:
```suggestion
- A Sector Duration Multiplier (SDM) is introduced for all newly-committed sectors, including Committed Capacity (CC) sectors and sectors containing storage deals.
```
This is stated below (I'm not advocating, just clarifying)

anorth (commented):
Looks fine to me

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,199 @@
...
+**Proposed Protocol Change**: 
+
+Introduce a multiplier based on sector duration
+ Name                                | Description                                           |
+| ----------------------------------- | ----------------------------------------------------- |
+| QualityBaseMultiplier (QBM)         | Multiplier for power for storage without deals.       |
+| DealWeightMultiplier (DWM)          | Multiplier for power for storage with deals.          |
+| VerifiedDealWeightMultiplier (VDWM) | Multiplier for power for storage with verified deals. |
+| **SectorDurationMultiplier (SDM)** | **Multiplier for power for storage based on promised sector duration** |
+
```
jennijuju, line 66: https://github.com/filecoin-project/FIPs/discussions/554#discussioncomment-4596644

jennijuju:
required additional information on collateral handling upon extension here https://github.com/filecoin-project/FIPs/discussions/554#discussioncomment-4596658

FIPS/fip-00XX.md
```diff
@@ -161,27 +209,26 @@ The proposed rewards multiplier increases potential risk to consensus. The main
...
-
-Factors that mitigate this risk are that it’s unlikely a single group could achieve 50% of FIL+ power consistently, and unlikely that the adversarial group exclusively takes up the longer duration sectors with enhanced power multipliers. 
-
-A limitation is that the above calculation assumes the malicious party is starting from 0% of consensus power. If they already control 10%, time to 33% is reduced to approximately 80 days.  
+````
+- If this scenario is maintained, the adversarial group is expected to exceed 33% of consensus power within 105 days, and 51% in 180 days. To do so needs 6.5 and 11 EiB respectively.
+- On one hand it’s unlikely a single group could achieve 50% of FIL+ power consistently, and unlikely that the adversarial group exclusively takes up the longer duration sectors with enhanced power multipliers.
+- A limitation is that the above calculation assumes the malicious party is starting from 0% of consensus power. If they already control 10%, time to 33% and 51% is reduced to approximately 70 and 150 days. And they have to gain 4.5 and 9.1 EiB.
```
ZenGround0, line 231: One thing Im not seeing that could be useful is a comparision between this and the same scenario without SDM, i.e. SDM = 1.  Its more useful to see the impact of this FIP.

FIPS/fip-00XX.md
```diff
@@ -29,6 +31,50 @@ The following aims to help better align the economic incentives of the network w
...
+Upon onboarding, the SDM applies from activaion epoch to expiration epoch: 
+
+```
+let duration_commitment = EPOCHS_IN_DAY * (expiration_epoch - activation_epoch)
+```
+### SDM Upon Sector Extension
+Upon extension, SDM applies to a commitment duration between current extension epoch and new expiration epoch: 
+
+``` 
+let duration_commitment = EPOCHS_IN_DAY * (expiration_epoch - extension_epoch);
```
ZenGround0, line 56: This is much clearer.  It implies a change to the data model of the sector info which should either be added here or in the implementation section.

ZenGround0:
One thing not yet specced that improves the implementation is to allow for SDM recalculation upon snap up.  This allows us to simplify some logic around distinguishing pre-SDM and SDM sectors and reduce the amount of data in the sector info by an int64.  > 3 * 10^8 sectors that saves us > 2 GiB.  I think it should be an acceptable variation of the intention behind the FIP.   Do you agree?

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,247 @@
...
+```
+new_pledge = max(old_pledge, new_pledge)
+```
+
+### Termination Fee Cap
+
+The [monies actor](https://github.com/filecoin-project/builtin-actors/blob/b7ad2c55363c363f61275ca45ef255e28f305254/actors/miner/src/monies.rs) termination penalty should be changed to: 
+
+```
+let penalized_reward = expected_reward * TERMINATION_REWARD_FACTOR_NUM * SDM(duration_commitment);
```
ZenGround0, line 75: We need a bit more detail to map this simple summary to the actual termination fee spec: `max(SP(t), BR(StartEpoch, 20d) + BR(StartEpoch, 1d) * terminationRewardFactor * min(SectorAgeInDays, 140))`

ZenGround0: One idea I haven't thought through much yet: change sector info day_reward and maybe twenty day reward to the max of current value and calculated value upon extension.

ZenGround0: One downside to this is that termination fee will overcount rewards.  At the cost of complexity we could potentially do a simple accumulator sum of rewards so far on each sdm-changing event.  This would change semantics of the day_reward field but shouldn't need a migration.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,247 @@
...
+
+// Sector Quality = 45
+StoragePledge := 0.3951
+ConsensusPledge := 8.5080 FIL
+PreCommitDeposit := StoragePledge = 0.3951
+InitialPledge := StoragePledge + ConsensusPledge = 8.9031
+```
+
+## Backwards Compatability 
+This policy would apply at Sector Extension and Upgrade for existing sectors. CEL has released public modeling and analysis [here](https://hackmd.io/uDXfe35UQuaLjED9u1fxyA) investgiating the decision to apply the policy either to just newly onboarded sectors, or to sectors upon extension. While siulations and modeling indicated that both versions of the SDM policy would be net positive for the network, the policy applying to all sectors would have stronger macroeconomic supply effects, whilst also decreasing barriers to sector commitment (pledges) supporting robust SP return on pledge profiles.
```
ZenGround0, line 196: Note the implementation complexity of allowing SDM recalculation for existing sectors looks essentially negligible so that shouldn't stop us from implementing the full thing from the start.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,247 @@
...
+Upon extension, SDM applies to a commitment duration between current extension epoch and new expiration epoch: 
+
+``` 
+let duration_commitment = EPOCHS_IN_DAY * (expiration_epoch - extension_epoch); 
+```
+
+The minimum sector extension time is 1 year. In the [policy actor](https://github.com/filecoin-project/builtin-actors/blob/45c56ed57190349f1856d3258af6c09a24ea1395/runtime/src/runtime/policy.rs#L358): 
+
+```
+pub const MIN_SECTOR_EXPIRATION: i64 = 365 * EPOCHS_IN_DAY;
```
anorth, line 62: Somewhere you should also specify `MAX_SECTOR_EXPIRATION_EXTENSION` going to 5 years.

FIPS/fip-00XX.md
anorth: > In the past this was probably acceptable because power could only go down so BR calculated at prove commit was an over estimate. As I understand SDM this property no longer holds which means we potentially need to update / add new state to sector info to maintain an upper bound on expected rewards for use in the termination fee.

Termination fee operates on the block reward calculated at the same time as the sector's power. The ReplicaUpdate operation already recalculates the expected day reward and 20-d projection using the _current_ network conditions, overwriting the sector info. This same recalculation should apply whenever the multiplier changes too. 

We don't currently use the max of current and new calculations because the previous value is copied to the replaced_day_reward field, and then both can be used when calculating termination fees. So I don't think we'll overcount. As discussed elsewhere, we'll need a new field so we can track both the original activation epoch and the last power-upgrade epoch. I think the simplification of only keeping track of the immediately previous reward values is ok, even though a sector could go though multiple changes in power.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,247 @@
...
+
+```
+fn sdm(duration_commitment: f64) -> f64 {
+    if duration_commitment <= (365.0 * 1.5).round() {
+        1.0
+    } else {
+        (duration_commitment - 183.0) / 365.0
+    }
+}
+```
```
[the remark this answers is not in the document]
anorth: Thanks for doing this with integers. I think it's not matching your intention in a number of ways due to truncation (e.g. `3/2 = 1`, and the resulting multiplier can only be an integer). The intention is relatively clear, but the implementation is going to need to be more complex, either to use a fixed-point representation with some precision (possibly embedded in i64) or delay divisions until after the SDM multiplies whatever it's multiplying.

The best version of this FIP would fully specify that. I'm not sure whether to ask for that, or accept perhaps reverting to the floating point version along with commentary that implementation will need to be in fixed-point. WDYT @ZenGround0 ?

kaitlin-beegle:
Spoke with @vkalghatgi; after reviewing for completeness, will merge.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,291 @@
+---
+fip: "00XX"
```
kaitlin-beegle, line 2: FIP0056 - will change after merging.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,291 @@
...
+
+This change shouldn’t be seen as precluding future termination fee changes. As the network incentives evolve with time, we should also expect the termination fees to be reassessed to ensure incentive alignment in the future.
+
+## Design Rationale
+
+### Supporting Longer-Term Commitments
+The current maximum commitment of 1.5 years limits the ability for SPs to make a long-term commitment to the network (or get rewarded for it). We expect that increasing the maximum allowable commitment to 5 years, while also introducing incentives to seal sectors for longer, can increase the stability of storage and predictability of rewards for SP’s. This is further discussed in the sections below.
+
+Note, that [FIP-0052](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0052.md) increases the maximum sector commitment duration from 1.5 years to 3.5 years as well, but does not introduce additional cryptoeconomic incentives for longer commitments. If this SDM proposal were to be accepted, then FIP-0052 would be withdrawn as it is a subset of the protocol changes outlined in this FIP.
```
kaitlin-beegle, line 174: This is quite an interesting precedent- I think it's worth considering the implication of so quickly superseding a FIP that was only just passed.  FIP0052 hasn't even been accepted long enough to be eligible for incorporation in a network upgrade.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,291 @@
...
+- The Duration Multiplier is multiplicative on the existing Quality Multiplier incentive (Filecoin Plus incentive). [Filecoin Plus](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0003.md) already offers up to a 10x multiplier for proving storage of data from verified clients. This FIP introduces an independent Duration Multiplier up to 4.5x. Thus, the maximum multiplier available to any sector will increase from 10x to 45x.
+- Sectors with higher Quality Adjusted Power as a result of the Sector Duration Multiplier and Quality Multiplier will require higher initial pledge collateral.
+- The minimum sector duration time will increase from 6 months to 1 year and the maximum sector duration will increase from 1.5 years to 5 years. The upper bound of [Deal Duration Bounds](https://github.com/filecoin-project/builtin-actors/blob/b5c101ab94f562ba43c1eca31bd1e73c6fc35794/actors/market/src/policy.rs#L33-L35) will increase to 5 years as well. 
+- CEL and the community will monitor the network and look to increase the maximum sector duration in a future FIP if network conditions merit, pending community-driven support for such measures.
```
kaitlin-beegle, line 18: This is not a requirement or standard that we have ever held FIPs to, but it's worth moving this direction in the future. 

If this FIP is accepted and implemented, it would be great to have auxiliary files added to the repo describing the monitoring and/or triaging of network conditions that CEL does post-implementation.

kaitlin-beegle (approved):

FIPS/fip-00XX.md
momack2: I think the proposed approach (withdraw and replace 52 with this proposal) is the simplest and most logical approach. This proposal is a superset of the FIP52 proposal, and is planned to replace it for implementation and effect in the network. Therefore withdrawing 52 is most accurate for any other implementation that might seek to align with protocol evolution / maintain consensus.

FIPS/fip-00XX.md
```diff
@@ -0,0 +1,291 @@
...
+2. advPower =  0.5 * 10 * 5 * 0.5 = 12.5
+3. honestPower = 0.5 * 10 * 1 * 5 * 0.5 + 1 * 1 * 5 * 0.5 = 15
+4. advPowerDailyPctGain = advPower/(19*1024 + honestPower)
+5. advPowerDailyPctGain = 0.06%
+```
+- If this scenario is maintained, the adversarial group is expected to exceed 33% of consensus power within 851 days, and 51% in 2046 days. To do so needs 10.4 and 25 EiB of storage power respectively.
+
+### Rollout Shock
+
+A risk is that the SDM slope is too steep, inadvertently restricting longer commitment. If this happens, optionality on the multiplier can be introduced in a future upgrade. CEL is committed to continually monitoring the economic health of the network, and adjusting parameter spaces as needed while including community input. We have also publicly released various analyses investigating the impact of potentially sudden changes in the network were this policy to be implemented [here](https://hackmd.io/uDXfe35UQuaLjED9u1fxyA)
```
momack2, line 279: Note that this is FIP is also scheduled for implementation / mainnet upgrade after FVM is live - meaning there is an opportunity for DeFi loan marketings and liquidity solutions to scale to meet the increase demand for SP collateral.

momack2 (approved):

FIPS/fip-00XX.md
anorth: FIP-0052 was presented with this potential to be obsolete in mind, as a fall-back. I think it's fine to withdraw if this one passes.

FIPS/fip-00XX.md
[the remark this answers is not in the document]
anorth: I'm ok with reverting to the float version that was here before

anorth (approved):

momack2:
Looks like we have more than sufficient approvals to merge! @kaitlin-beegle do you want to do the renaming honors?

@vkalghatgi - looks like a few follow up changes to land for @anorth, but can make as a follow-up PR.

PR #574, opened 2022-12-16
FIP0048: extend f4 address restriction to embryos

Stebalien:
In addition to restricting `Exec4` to "blessed" address managers, this change now restricts embryo creation to addresses namespaced under "blessed" address manager IDs.

This restriction will be lifted in a future FIP.


FIPS/fip-0048.md
```diff
@@ -71,6 +71,8 @@ There are three key parts to the design:
 2. The introduction of embryo actors.
 3. A new `Exec4` method on the init actor to support the `f4` address class.
 
+**NOTE**: For now, f4 addresses may only be assigned by and in association with specific "blessed" address managers, the first of these will likely be the this will likely be the "Ethereum Address Manager (EAM)" (in NV18). Once users are able to deploy custom WebAssembly actors, this restriction should be relaxed (in a future FIP).
```
raulk, line 74: > NV18

nit: avoid making assumptions about FIP acceptance/shipping in particular network versions.

raulk: (I see this is moved content, but still)

FIPS/fip-0048.md
```diff
@@ -105,8 +107,8 @@ pub extern "C" fn invoke(_: u32) -> u32 {
 
 On `send` to an unassigned `f4` address `f4{actor-id}f{subaddress}`, the FVM will:
 
-1. Validate that an actor with ID `actor-id` exists.
-2.  Create a new "embryo actor" to hold the received funds, assigning the `f4` address to that new actor.
+1. Validate that an actor with ID `actor-id` exists. For now, the `actor-id` will be further restricted to "blessed" address managers.
+2. Create a new "embryo actor" to hold the received funds, assigning the `f4` address to that new actor.
```
raulk, line 111: Since we are in draft review time, I will use this opportunity to review the term "embryo" more generally. This term evokes confusion and "can you explain?" reactions, which is tells me that it's a not a good factual name.

A possible issue is that the term "embryo" is highly **_metaphoric_**. Metaphores can be effective when used as part of a story, but generally ineffective when used standalone. We really want easy comprehension for protocol concepts, so I'd avoid using metaphores for protocol names.

Can we consider the name "placeholder actor"? This term already denotes that it's a temporary entity subject to some future replacement. And the replacement, or transition out of this state into a _concrete_ actor (EEOA or EVM runtime in this case), can be termed "reification".

anorth (approved):
Looks good module resolution of @raulk's comments.

FIPS/fip-0048.md
ribasushi: @raulk isn't [`stub`](https://en.wikipedia.org/wiki/Method_stub) the exact CS-ey term you are looking for here?

raulk: Stub works for me too. And it's shorter than placeholder or embryo.

Stebalien: Yeah, you're right.

> Sending to an f4 address creates a placeholder actor.
> Sending to an f4 address creates a stub actor.
> Sending to an f4 address creates an uninitialized actor.

> Exec4 can deploy code to an existing placeholder actor.
> Exec4 can deploy code to an existing stub actor.
> Exec4 can deploy code to an existing uninitialized actor.

Stebalien: Discussion opened in https://github.com/filecoin-project/FIPs/discussions/473#discussioncomment-4453474.

PR #569, opened 2022-12-12
FIP-0054: Filecoin EVM runtime (FEVM)

raulk:
## Simple Summary

We introduce a new built-in actor: the Filecoin EVM (FEVM) runtime actor, capable of running EVM smart contracts on top of the Filecoin Virtual Machine. We also introduce various changes to the Filecoin Virtual Machine and to client implementations, necessary to support the operation of this new actor.

## Abstract

The Filecoin EVM (FEVM) runtime built-in actor runs EVM smart contracts compatible with the [Ethereum Paris fork].

It achieves this by embedding a special-purpose EVM interpreter, implementing the integration logic with the Filecoin environment, translating environment-dependent opcodes into their corresponding Filecoin primitives, and mapping all state I/O to the underlying IPLD model.

The EVM interpreter strives for maximal as-is portability. To such avail, it The EVM interpreter supports all required opcodes and Ethereum precompiles to strive for maximal portability. Functional and technical departures from Ethereum's standard expectations are documented herein.

The FEVM runtime actor motivates some changes in the FVM. Syscalls are modified, syscalls are added, new environmental data is required, and new exit codes are created. The concept of a TipsetCID is also introduced formally, and is of required implementation by clients.

This FIP is dependent on [FIP-0048] (f4 address class), [FIP-0049] (Actor events), and [FIP-0055] (Supporting Ethereum Accounts, Addresses, and Transactions).

[Ethereum Paris fork]: https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/
[FIP-0048]: https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0048.md
[FIP-0049]: https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md
[FIP-0055]: https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0055.md

PR #580, opened 2022-12-21
FIP for Ethereum accounts, addresses, and signatures (broken out of FIP-0054)

raulk:
This FIP breaks out the Ethereum Account (formally EEOA), EAM, Delegated signatures, and Ethereum addresses out of the monolith [FIP-0054](https://github.com/filecoin-project/FIPs/pull/569). This will allow for more focused and efficient community review.

--- 

## Simple Summary

We introduce two actors: the **Ethereum Address Manager (EAM)** and the **Ethereum Account** actors.
Both actors are necessary to attain compatibility with Ethereum tooling. We also introduce **Ethereum addresses** over the f4 address class, as well as a new **Delegated signature type** to accept native Ethereum transactions.

## Abstract

The Ethereum Address Manager (EAM) actor manages namespace `10` within the `f4` address class as defined in [FIP-0048](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0048.md). This namespace hosts Ethereum addresses in Filecoin. Only Ethereum Accounts and EVM smart contracts deployed through the EAM acquire an Ethereum address. These addresses begin with `f410` in textual form.We use this notation throughout this document.

The Ethereum Account actor represents an externally-owned Ethereum account, backed by a secp256k1 key. It enables authenticating native RLP-encoded [EIP-1559 Ethereum transaction] submitted from Ethereum wallets, within the Filecoin protocol. These messages are converted to Filecoin messages carrying a Delegated signature type.

While baked into the protocol at this stage, the Ethereum Account and the Delegated signature type are extension points to eventually transition to future fully-fledged Account Abstraction (AA).

### DIFF
--- a/FIPS/fip-nnnn-eam-eth-accounts.md
+++ b/FIPS/fip-nnnn-eam-eth-accounts.md
@@ -1,0 +1,386 @@
+---
+fip: <to be assigned>
+title: Supporting Ethereum accounts, addresses, and transactions
+author: Raúl Kripalani (@raulk), Steven Allen (@stebalien)
+discussions-to: <URL>
+status: Draft
+type: Technical Core
+category: Core
+created: 2022-12-02
+spec-sections:
+requires: N/A
+replaces: N/A
+---
+
+# Supporting Ethereum accounts, addresses, and transactions
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Simple Summary](#simple-summary)
+- [Abstract](#abstract)
+- [Change Motivation](#change-motivation)
+- [Specification](#specification)
+  - [Ethereum Address Manager (EAM)](#ethereum-address-manager-eam)
+    - [Installation and wiring](#installation-and-wiring)
+    - [EVM smart contract deployment via the EAM](#evm-smart-contract-deployment-via-the-eam)
+    - [State](#state)
+    - [Actor interface (methods)](#actor-interface-methods)
+      - [Method number `1` (`Constructor`)](#method-number-1-constructor)
+      - [Method number `2` (`Create`)](#method-number-2-create)
+      - [Method number `3` (`Create2`)](#method-number-3-create2)
+  - [Delegated signature type](#delegated-signature-type)
+    - [Transitory Delegated signature validation](#transitory-delegated-signature-validation)
+    - [Message transliteration](#message-transliteration)
+  - [Ethereum Account](#ethereum-account)
+    - [Installation and wiring](#installation-and-wiring-1)
+    - [Instantiation via promotion](#instantiation-via-promotion)
+    - [Actor interface (methods)](#actor-interface-methods-1)
+    - [State](#state-1)
+- [Design Rationale](#design-rationale)
+  - [f410 address scope](#f410-address-scope)
+  - [Upgrade path towards Account Abstraction (AA)](#upgrade-path-towards-account-abstraction-aa)
+- [Backwards Compatibility](#backwards-compatibility)
+- [Test Cases](#test-cases)
+- [Security Considerations](#security-considerations)
+- [Incentive Considerations](#incentive-considerations)
+- [Product Considerations](#product-considerations)
+- [Implementation](#implementation)
+- [Copyright](#copyright)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Simple Summary
+
+We introduce two actors: the **Ethereum Address Manager (EAM)** and the **Ethereum Account** actors.
+Both actors are necessary to attain compatibility with Ethereum tooling.
+We also introduce **Ethereum addresses** over the f4 address class, as well as a new **Delegated signature type** to accept native Ethereum transactions.
+
+## Abstract
+
+The Ethereum Address Manager (EAM) actor manages namespace `10` within the `f4` address class as defined in [FIP-0048].
+This namespace hosts Ethereum addresses in Filecoin.
+Only Ethereum Accounts and EVM smart contracts deployed through the EAM acquire an Ethereum address.
+These addresses begin with `f410` in textual form.
+We use this notation throughout this document.
+
+The Ethereum Account actor represents an externally-owned Ethereum account, backed by a secp256k1 key.
+It enables authenticating native RLP-encoded [EIP-1559 Ethereum transaction] submitted from Ethereum wallets, within the Filecoin protocol.
+These messages are converted to Filecoin messages carrying a Delegated signature type.
+
+While baked into the protocol at this stage, the Ethereum Account and the Delegated signature type are extension points to eventually transition to future fully-fledged Account Abstraction (AA).
+
+## Change Motivation
+
+Foreign runtimes, like the EVM runtime introduced in FIP-TODO, often make assumptions about addressing schemes.
+Programs deployed on those runtimes (e.g. EVM smart contracts) also inherit those assumptions.
+Such assumptions come in the form of:
+
+- data types expectations
+- cryptographic expectations (e.g. addresses derived from public keys)
+- address predictability expectations (e.g. upon contract deployment)
+
+Fulfilling these expectations involves making Filecoin aware of Ethereum addresses, in a way that does not bake that awareness in the core protocol.
+This is achieved by modelling Ethereum addresses over the newly-introduced `f4` delegated address class, under a dedicated namespace, managed by the EAM.
+
+Furthermore, supporting foreign runtimes in Filecoin is hardly useful if the tools and libraries built for those runtimes were unable to interact with Filecoin without changes.
+Such handicap would harm easy reuse.
+Supporting the submission and validation of transactions issued from wallets native to the original ecosystem is a must.
+These capabilities are delivered by the Ethereum Account and the Delegated signature type, as a stepping stone towards full Account Abstraction (AA).
+
+## Specification
+
+### Ethereum Address Manager (EAM)
+
+We introduce the Ethereum Address Manager (EAM) actor as a singleton built-in actor, sitting at ID address `f010`.
+
+This actor manages the Ethereum address space, anchored at the f4 address namespace equalling its actor ID (`10` in decimal).
+The address payload is the literal Ethereum address in binary form.
+```
+# Ethereum address in Filecoin
+0x04 || 0x10 || <binary Ethereum address>
+```
+
+The EAM also acts like a factory for EVM smart contracts, respecting Ethereum rules and semantics for the `CREATE` and `CREATE2` mechanisms for address generation.
+
+#### Installation and wiring
+
+During migration:
+
+1. The Wasm bytecode of the EAM (referred by its CodeCID) is linked under the System actor's registry under the key `"eam"`.
+2. A single copy of the EAM is deployed as a singleton actor under ID address `f010`.
+
+The Init actor statically recognizes the EAM as an _authorized_ delegated address manager, and permits calls to its `InitActor#Exec4` method originating in it.
+Address managers like the EAM use the `InitActor#Exec4` method to deploy new actors with delegated addresses under the namespace matching their actor ID (`10` in this case), with the address payload being provided as an argument.
+
+#### EVM smart contract deployment via the EAM
+
+The EAM offers two Filecoin methods `Create` and `Create2`, equivalent to their Ethereum counterparts.
+In Ethereum, `CREATE` and `CREATE2` differ in their address generation rules:
+
+- `CREATE` computes the address in this manner: `Keccak-256(rlp_list(sender, nonce))`. It is thus non-deterministic.
+- `CREATE2` computes the addres in this manner: `Keccak-256(rlp_list(0xff, sender, salt, bytecode))`. It is suitable for counterfactual deployments.
+
+EVM smart contracts are deployed to the Filecoin chain as follows:
+
+1. A contract deployment message is directed to the `Create` or `Create2` methods of the EAM.
+2. The EAM computes the `f410` address as per the rules of the invoked method.
+   The resulting f410 address is semantically equivalent to the address Ethereum would've generated for the same inputs, thus retaining tool compatibility.
+2. The EAM invokes the Init actor's `Exec4` method (see [FIP-0048]), supplying:
+      - The CodeCID of the EVM runtime actor.
+      - The EVM runtime actor's constructor parameters, concretely the EVM init code and original creator address.
+      - The assigned f410 address.
+
+#### State
+
+None.
+
+#### Actor interface (methods)
+
+##### Method number `1` (`Constructor`)
+
+The Constructor asserts that:
+
+1. It is being called by the System actor (f00).
+2. It has been deployed on address `f10`.
+
+These checks prevent instantiation by users.
+
+_Input parameters_
+
+None.
+
+_Return value_
+
+None.
+
+_Errors_
+
+- `USR_FORBIDDEN` (18) when attempts to instantiate this actor are made, or if deployed at an address other than f010.
+
+##### Method number `2` (`Create`)
+
+Deploys a smart contract taking EVM init bytecode and accepting a nonce from the user.
+The Ethereum address (and therefore the f410 address) is calculated from these parameters.
+
+The `Create` method is to be used in two cases:
+
+1. When an EEOA deploys an EVM smart contract by submitting a native Ethereum message.
+2. When a smart contract (running within the EVM runtime actor) calls the `CREATE` opcode.
+
+Note that this differs from Ethereum in that we take a user-provided nonce as a parameter.
+This is necessary for the EAM to learn the nonce of a contract, when the `CREATE` opcode is used.
+Technically, this allows an EEOA to deploy a contract with an arbitrary nonce, potentially using nonces in a non-contiguous manner.
+
+_Input parameters_
+
+```rust
+// DAG-CBOR tuple encoded.
+pub struct CreateParams {
+    /// EVM init code.
+    pub initcode: Vec<u8>,
+    /// Nonce with which to create this smart contract.
+    pub nonce: u64,
+}
+```
+
+_Return value_
+
+```rust
+// DAG-CBOR tuple encoded.
+pub struct Return {
+    /// The ID of the EVM runtime actor that was constructed.
+    pub actor_id: ActorID,
+    /// Its f2 address.
+    pub robust_address: Address,
+    /// Its Ethereum address, translatable to an f410 address.
+    pub eth_address: [u8; 20],
+}
+```
+
+_Errors_
+
+TODO.
+
+##### Method number `3` (`Create2`)
+
+The deployment procedure is the same as `Create`, but this method takes a user-provided salt.
+A deterministic Ethereum address (and therefore the f410 address) is calculated from these parameters.
+
+`Create2` only used when a smart contract (running within the EVM runtime actor) calls the `CREATE2` opcode.
+
+_Input parameters_
+
+```rust
+pub struct Create2Params {
+    /// EVM init code.
+    pub initcode: Vec<u8>,
+    /// User provided salt.
+    pub salt: [u8; 32],
+}
+```
+
+_Return value_
+
+Same as [`Create`](#method-number-1-create).
+
+### Delegated signature type
+
+We introduce a new `Delegated` signature type for Filecoin messages with serialized value `3`.
+Delegated signatures are used in combination with abstract account senders, once Account Abstraction (AA) is fully realized.
+Delegated signatures are opaque to the protocol, and are validated by actor logic.
+However, we establish a transitory period where Delegated signatures are special-cased.
+
+#### Transitory Delegated signature validation
+
+We establish a transitory period whereby `Delegated` signatures are assumed to be secp256k1 signatures over the [RLP] representation of a native [EIP-1559 Ethereum transaction] sent from an Ethereum Acount actor.
+This effectively couples the `Delegated` signature type to its only possible use at this stage, but provides an extension point for the future.
+
+Such transitory `Delegated` signatures must be verified by:
+
+1. Repacking the native RLP-encoded [EIP-1559 Ethereum transaction] from the Filecoin message (original signature payload).
+2. Recovering the secp256k1 public key from such payload and the ECDSA signature.
+3. Computing the Ethereum address from the secp256k1 public key by hashing the public key using Keccak-256, and retaining the last 20 bytes.
+4. Calculating the corresponding f410 address (see conversion rules [EAM rules](#ethereum-address-manager-eam)).
+5. Asserting that the sender's f410 address matches the expected f410 address.
+
+The client must perform this verification before chain inclusion, and before handing off the message to the FVM for execution.
+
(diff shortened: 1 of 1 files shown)


raulk:
@jennijuju @kaitlin-beegle can I have a FIP number here, please?

FIPS/fip-nnnn-eam-eth-accounts.md
```diff
@@ -0,0 +1,386 @@
+---
+fip: <to be assigned>
```
jennijuju, line 2: 0055

jennijuju:
Is this ready  for editor review?

raulk:
> Is this ready for editor review?

Yes. FIP number incorporated.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,401 @@
+---
+fip: "0055"
+title: Supporting Ethereum accounts, addresses, and transactions
```
jennijuju, line 3:
```suggestion
title: Supporting Ethereum Accounts, Addresses, and Transactions
```

raulk: Applied.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,413 @@
+---
+fip: "0055"
+title: Supporting Ethereum Accounts, Addresses, and Transactions
+author: Raúl Kripalani (@raulk), Steven Allen (@stebalien)
+discussions-to: <URL>
```
kaitlin-beegle, line 5: There's not a clear need for this FIP to have its own discussion post, except that there ought to be a landing place for someone who wants to ask questions about this particular spec, etc.

Unless you want to open up a distinct comment thread, let's just direct them here: https://github.com/filecoin-project/FIPs/discussions/287

kaitlin-beegle (approved):

FIPS/fip-0055.md
```diff
@@ -0,0 +1,413 @@
...
+In order to enable existing Ethereum tools to seamlessly interact with the Filecoin network, it is necessary for Filecoin to:
+
+1. Recognize and deal with Ethereum addresses.
+2. Fulfill Ethereum address expectations, both for accounts (secp256k1 public key derived) and contracts.
+3. Model Ethereum's Externally-Owned Accounts (EOAs).
+4. Support and validate native transactions issued by (EOAs) from Ethereum wallets.
+
+This FIP delivers these requirements by leveraging the f4 extensible address class (newly introduced in [FIP-0048]) to model Ethereum Addresses, managed by the **Ethereum Address Manager (EAM)** actor introduced herein.
+
+It also introduces the **Ethereum Account** actor to represent Ethereum Externally Owned accounts, capable of acting as message senders, and a new **Delegated signature type**, which enables validating native EIP-1559 Ethereum transactions carrying an secp256k1 ECDSA signatures over their RLP encoding.
```
jennijuju, line 69: add reference link to EIP-1559 etherum transaction plz

FIPS/fip-0055.md
```diff
@@ -0,0 +1,413 @@
...
+
+This FIP introduces three technical elements:
+
+1. It formally defines the `f410` Ethereum address space, nested under decimal namespace `10` within the `f4` address class as defined in [FIP-0048].
+   This address space is managed by the Ethereum Address Manager (EAM) actor, which also offers public methods to act as a factory for EVM smart contracts.
+   The address payload of an `f410` address is the original Ethereum address.
+   Ethereum addresses can be cast as `f410` addresses, and viceversa.
+
+2. The Ethereum Account actor, representing an Externally-Owned Ethereum account, backed by a secp256k1 key, and capable of acting as a sender of native EIP-1559 Ethereum transactions.
+   This actor will eventually become an Abstract Account (AA) actor, once such framework is introduced.
```
jennijuju, line 83:
```suggestion
   This actor may become an Abstract Account (AA) actor, once such framework is introduced.
```

given that we dont necessarily have a full AA design yet, the tune here is too assertive imo

FIPS/fip-0055.md
```diff
@@ -0,0 +1,413 @@
...
+
+Both these elements act as extension points for an eventual fully-fledged [Account Abstraction] framework.
+
+## Abstract
+
+This FIP introduces three technical elements:
+
+1. It formally defines the `f410` Ethereum address space, nested under decimal namespace `10` within the `f4` address class as defined in [FIP-0048].
+   This address space is managed by the Ethereum Address Manager (EAM) actor, which also offers public methods to act as a factory for EVM smart contracts.
+   The address payload of an `f410` address is the original Ethereum address.
```
jennijuju, line 79:
```
/// Payload is the data of the Address. Variants are the supported Address protocols.
#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
#[cfg_attr(feature = "arb", derive(arbitrary::Arbitrary))]
pub enum Payload {
    /// f0: ID protocol address.
    ID(u64),
    /// f1: SECP256K1 key address, 20 byte hash of PublicKey.
    Secp256k1([u8; PAYLOAD_HASH_LEN]),
    /// f2: Actor protocol address, 20 byte hash of actor data.
    Actor([u8; PAYLOAD_HASH_LEN]),
    /// f3: BLS key address, full 48 byte public key.
    BLS([u8; BLS_PUB_LEN]),
    /// f4: Delegated address, a namespace with an arbitrary subaddress.
    Delegated(DelegatedAddress),
}
```

When I see `address payload` in Filecoin, above is what comes to my mind. However, I believe you are saying that _xxxxxxxx_ in  `f410<xxxxxxxx>`  is an eth address. in the f4 fip, we refer that as `subaddress` ([here](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0048.md#hashing-v-prefixing), I think we shall stay consistent

jennijuju (changes requested):

jennijuju:
(im only half way through and will finish 👁️ tmr morning. @raulk could you please confirm if this PR is thorough review read in preparation of last call, or its just a draft update?

FIPS/fip-0055.md
raulk: Hm, we gotta capture the optionality of subjects accurately here. _If_ AA is introduced, then the Eth Account will _absolutely_ become an Abstract Account. I feel like the original text is more aligned with that interpretation.

raulk: How about:

> This actor will become an Abstract Account (AA) actor, if such framework is eventually introduced in the protocol.

FIPS/fip-0055.md
raulk: @jennijuju Well pointed! Indeed this should use f4 specific terminology here (subaddress). The _payload_ of f4 is `10<xxxxxxxx>` in your example above.

raulk:
> (im only half way through and will finish 👁️ tmr morning. @raulk could you please confirm if this PR is thorough review read in preparation of last call, or its just a draft update?

Feel free to do a Core Dev review in addition to a FIP Editor review here since my understanding is that we're pushing this over to Last Call imminently. But please separate both reviews as FIP Editor review may be blocking for merge, and Core Dev review might not be.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,401 @@
...
+- [Implementation](#implementation)
+- [Copyright](#copyright)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Simple Summary
+
+We introduce two actors: the **Ethereum Address Manager (EAM)** and the **Ethereum Account** actors.
+Both actors are necessary to attain compatibility with Ethereum tooling.
+We also introduce **Ethereum addresses** over the f4 address class, as well as a new **Delegated signature type** to accept native Ethereum transactions.
```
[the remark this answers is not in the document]
raulk: This is outdated now.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,413 @@
...
+## Simple Summary
+
+In order to enable existing Ethereum tools to seamlessly interact with the Filecoin network, it is necessary for Filecoin to:
+
+1. Recognize and deal with Ethereum addresses.
+2. Fulfill Ethereum address expectations, both for accounts (secp256k1 public key derived) and contracts.
+3. Model Ethereum's Externally-Owned Accounts (EOAs).
+4. Support and validate native transactions issued by (EOAs) from Ethereum wallets.
+
+This FIP delivers these requirements by leveraging the f4 extensible address class (newly introduced in [FIP-0048]) to model Ethereum Addresses, managed by the **Ethereum Address Manager (EAM)** actor introduced herein.
```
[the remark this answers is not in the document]
raulk: Ack, will add.

FIPS/fip-0055.md
raulk: It's heavily referenced elsewhere, but will add it here too.

FIPS/fip-0055.md
raulk: I've linked to these three discussion threads here:

  - https://github.com/filecoin-project/FIPs/discussions/388
  - https://github.com/filecoin-project/FIPs/discussions/490
  - https://github.com/filecoin-project/FIPs/discussions/287

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+
+- The actor implementation lives in [`filecoin-project/builtin-actors`].
+- The reference FVM implementation lives in [`filecoin-project/ref-fvm`].
+- The reference client implementation lives in [`filecoin-project/lotus`].
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+[Externally-Owned Accounts (EOAs)]: https://ethereum.org/en/developers/docs/accounts/
```
arajasek, line 436: Totally didn't know one could do this!

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+
+The EAM also acts like an EVM smart contract factory, offering methods that respect the Ethereum address generation rules and semantics for the original `CREATE` and `CREATE2` mechanisms.
+
+#### Installation and wiring
+
+During migration:
+
+1. The Wasm bytecode of the EAM (referred by its CodeCID) is linked under the System actor's registry under the key `"eam"`.
+2. A single copy of the EAM is deployed as a singleton actor under ID address `f010`.
+3. The EAM is allowed-listed in the Init actor as an _authorized_ delegated address manager.
```
arajasek, line 157: BLOCKER: We also need to create the Ethereum Zero address (`f410faaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaonc6iji`) as an EthAccount:
- Assign it an ID address in the init actor's state
- Put it as an EthAccount in the state tree

Can resolve by either adding to this FIP, or pointing to another FIP that already covers this / where we plan to cover this. I think it's most relevant to this FIP though.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+
+Failure to meet chain inclusion gas will prevent the inclusion of the message and, therefore, will not promote the Placeholder into an Ethereum Account, even if we technically ascertained that the Placeholder corresponds to an Externally-Owned Account.
+
+#### Actor interface
+
+None.
+
+#### State
+
+None.
```
arajasek, line 331: Similar note about `State` as for the EAM above.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+We prototyped [Account Abstraction] solutions extensively before submitting this FIP.
+This FIP preserves the key extension points that will facilitate a (projected) elegant transition to fully-fledged [Account Abstraction] (AA).
+
+We forecast these rough steps to conduct the transition:
+
+1. Remove Delegated signature validation logic from clients, treating the signature as an opaque blob at the client level.
+2. Perform AA-specific gas preflight checks in the client before accepting the message for inclusion, or propagating via the mpool.
+3. Move the Ethereum transaction signature verification logic inside the `ethaccount` actor, so that it runs on-chain.
+4. Generalise the Placeholder actor promotion path so that the relevant address manager gets to indicate the promotion target. Reconcile with (3).
+5. Trigger the signature validation logic within the FVM the sender is an Abstract Account.
```
arajasek, line 346:
```suggestion
5. Trigger the signature validation logic within the FVM when the sender is an Abstract Account.
```

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+
+#### Actor interface
+
+None.
+
+#### State
+
+None.
+
+## Design Rationale
```
arajasek, line 333: I don't think the upgrade path towards "full" AA is the only relevant design rationale to discuss. At the very least, I think we should reiterate (or extract and move here) some of the salient points under the Change Motivation section about wanting to to retain full Ethereum tooling and ecosystem stack compatibility with Filecoin.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+
+1. Remove Delegated signature validation logic from clients, treating the signature as an opaque blob at the client level.
+2. Perform AA-specific gas preflight checks in the client before accepting the message for inclusion, or propagating via the mpool.
+3. Move the Ethereum transaction signature verification logic inside the `ethaccount` actor, so that it runs on-chain.
+4. Generalise the Placeholder actor promotion path so that the relevant address manager gets to indicate the promotion target. Reconcile with (3).
+5. Trigger the signature validation logic within the FVM the sender is an Abstract Account.
+
+## Backwards Compatibility
+
+Backwards compatibility is not affected, as this FIP introduces strictly additive features.
```
arajasek, line 350: FIP authors usually discuss the need for a network upgrade in this section (see a very simple example [here](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0044.md#backwards-compatibility)). TBH, I'm not sure this is especially useful, since I don't think we've ever had a FIP that didn't need a network upgrade, but you can match precedent if you'd like.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+
+### Exchange adaptation
+
+Exchanges must support receiving funds from EVM smart contracts.
+
+Exchange may choose to enable sending to `f410` addresses, and may support taking in Ethereum addresses and converting them to `f410` addresses.
+
+## Implementation
+
+- The actor implementation lives in [`filecoin-project/builtin-actors`].
```
arajasek, line 428: Can we link the appropriate branches here, with a note that they will land in main branches upon this FIP's acceptance?

arajasek (changes requested):
Broadly looks good to me, various suggestions, only a couple blockers.

As a general note, this FIP is a little too forward-looking in terms of Account Abstraction. I think I understand the motivation for this approach to the FIP, and that future enhancement is germane to the design & history of this FIP, but it would be nicer if one could read & understand the specification of the _immediate_ proposed change without thinking too much about future changes.

On an aside, it's really helpful for folks implementing migrations for network upgrades if a FIP specifies all necessary migration changes in a single section. This isn't especially relevant to this FIP (the migration is relatively simple, and has already been implemented in Golang), but I think it might be a good idea to modify the FIP template to add an explicit section for migrations. Note for future discussion.

FIPS/fip-0055.md
raulk: This is already covered in the Migration section.

raulk: No, sorry, it's covered in the new Migration section of FIP-0054 (FEVM runtime). I was torn about where to put it, but likely resides better here with a reference from FIP-0054?

FIPS/fip-0055.md
raulk: I considered that, but discarded it to avoid repetition. Can do if folks really want that repetition here.

FIPS/fip-0055.md
raulk: What's the relationship between a network upgrade and backwards compatibility?

raulk: There's a Migration section which covers the transition during the network upgrade.

FIPS/fip-0055.md
arajasek: I haven't read FIP-0054 yet (sorry!), but I think it belongs here? Seems pretty related to the specifics of the EAM and EthAccounts, but defer to you.

So long as it's specified _somewhere_ (and I confirmed that it is [here](https://github.com/filecoin-project/FIPs/blob/f99bd54f8755ef4a0059ada6133f150fa19e81be/FIPS/fip-0054.md#installation-and-wiring)), I'm happy to drop this as a blocker (and am doing so now).

FIPS/fip-0055.md
arajasek: Yeah, I've had similar uncertainty when writing FIPs myself -- there's often major overlap between the Change Motivation and the Design Rationale (with the latter coming down to "designed so as to achieve the motivated change".

Defer to you / anyone else who feels strongly one way or the other.

FIPS/fip-0055.md
arajasek: It's "backwards incompatible" in that it's "v9 incompatible", and so needs new actors version / network version -- that's often all this section has come down to.

Don't think you should necessarily match precedent here, though.

FIPS/fip-0055.md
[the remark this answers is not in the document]
raulk: I think the discrepancy lies in that you interpreted this as an explicit action (e.g. there is some kind of transaction that happens after which the EAM becomes authorized), whereas I intended to write this as a constraint of the code. I can see why there's confusion here. Let me rejigger this.

raulk: Ack, I'll add it here too. It's a bit hard to choose for things that are in between the seams.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+In order to enable existing Ethereum tools to seamlessly interact with the Filecoin network, it is necessary for Filecoin to:
+
+1. Recognize and deal with Ethereum addresses.
+2. Fulfill Ethereum address expectations, both for accounts (secp256k1 public key derived) and contracts.
+3. Model Ethereum's [Externally-Owned Accounts (EOAs)].
+4. Support and validate native transactions issued by (EOAs) from Ethereum wallets.
+
+This FIP delivers these requirements by leveraging the f4 extensible address class (newly introduced in [FIP-0048]) to model Ethereum Addresses, managed by the **Ethereum Address Manager (EAM)** built-in actor introduced herein.
+
+It also introduces the **Ethereum Account** built-in actor to represent Ethereum Externally Owned accounts, capable of acting as message senders, and a new **Delegated signature type**, which enables validating native [EIP-1559 Ethereum transaction]s carrying an secp256k1 ECDSA signatures over their RLP encoding.
```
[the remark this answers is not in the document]
raulk: Fixed, thanks!

FIPS/fip-0055.md
raulk: Fixed, thanks!

FIPS/fip-0055.md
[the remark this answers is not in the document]
raulk: I've tried to bring more accuracy to this in 007f126.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,444 @@
...
+#### Relationship with the Init actor
+
+The Init actor statically recognizes the EAM as an _authorized_ delegated address manager, and allows it to call its `InitActor#Exec4` method.
+Note that address managers like the EAM use the `InitActor#Exec4` method to deploy new actors with a delegated f4 address under the namespace matching their actor ID (`10` in this case).
+The caller provides the subaddress as an argument to such method.
+For more details, refer to [FIP-0048].
+
+#### State
+
+This actor has no state.
```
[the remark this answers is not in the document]
raulk: Fixed in 682e0f7.

FIPS/fip-0055.md
raulk: Fixed in 682e0f7.

FIPS/fip-0055.md
raulk: See 75b22a6.

FIPS/fip-0055.md
raulk: Are there any specific design decisions you would like to call out here? If not, let's leave it as-is. The general address manager stuff should be in FIP-0048 (the one that introduces the notion of address managers), and I don't think there are any other noteworthy decisions specific to the EthAccount actor here beyond AA.

FIPS/fip-0055.md
raulk: Added.

arajasek (approved):

FIPS/fip-0055.md
raulk: Done.

FIPS/fip-0055.md
anorth: I agree with avoiding repetition. I interpret these sections as:
- Motivation: what's the problem and why is it worth solving at all
- Design rationale: why solve it _this way_, instead of some other ways

As a technical reviewer, I would like to see more explanation of why promotion is done this way, why it can't be done a more general way from the beginning. The section could probably highlight each of the Ethereum-specific mechanisms proposed (e.g. also the transitory signature delegation), explain they are temporary, and explain why the shortcut was needed.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,447 @@
...
+
+### Explorer adaptation
+
+Explorers will need to adapt to display `f410` addresses bound to actors and their respective Ethereum addresses, on actor pages, message pages, and more.
+
+Explorers will need to recognize Ethereum Account actors and Delegated signatures.
+
+### Exchange adaptation
+
+Exchanges must support receiving funds from EVM smart contracts.
```
anorth, line 425: Some don't even support receiving funds from native built-in actors like a multisig. There's not much we can do about it. I suggest dropping this, or at least softening to just being one more thing that exchanges might want to consider.

FIPS/fip-0055.md
```diff
@@ -0,0 +1,447 @@
...
+This FIP introduces a new signature type and a new transaction format to the protocol.
+That is, we introduce new pathways to initiate transactions in the system, and extend the kinds of payloads that are entitled to do so.
+
+It is worthy to note that Filecoin already supports secp256k1 ECDSA signatures under signature type 1.
+Clients are expected to reuse existing signature verification and key recovery logic.
+However, what differs is the authentication logic.
+The original signature payload must be reconstructed by repacking the RLP-encoded [EIP-1559 Ethereum transaction].
+Any flaws in this logic is subject to security events.
+
+### Inability to stably address some actors from EVM smart contracts
```
anorth, line 374: I don't understand this section, but discussion taken to https://github.com/filecoin-project/FIPs/discussions/490#discussioncomment-4648606

FIPS/fip-0055.md
```diff
@@ -0,0 +1,447 @@
...
+
+#### Transitory validation
+
+We establish a transitory period whereby `Delegated` signatures are assumed to be secp256k1 signatures over the [RLP] representation of a native [EIP-1559 Ethereum transaction] sent from an Ethereum Acount actor.
+The RLP representation is reconstituted from the Filecoin message as specified below.
+This effectively couples the `Delegated` signature type to its only possible use at this stage, but provides an extension point for the future introduction of [Account Abstraction] (AA).
+
+Such transitory `Delegated` signatures must be verified by:
+
+1. Reforming and repacking the native RLP-encoded [EIP-1559 Ethereum transaction] from the Filecoin message (original signature payload).
```
q9f, line 269: EIP-1559 implies EIP-2718 transaction envelopes but you might want to specifically require that to reserve the potential to add more custom transaction types in future; either your own types, or if necessary, upstream Ethereum transaction types.

PR #582, opened 2022-12-30
Create CODEOWNERS

jennijuju:
Id like to propose to  enable "2 approval from fip editors", the goal is to encourage deeper and more thorough reviews before DRAFTs move to LAST CALLs


raulk (commented):
I've kicked off a more nuanced reflection with Core Devs around the FIP editing process. I'd suggest not making changes until we recognise the issues collectively and design an action plan.

anorth:
I don't think this will achieve the goal, and possibly the opposite. Thanks @raulk for prompting discussion about the motivation first.

PR #597, opened 2023-01-20
FIP-0055: Expand on delegated signature validation

arajasek:
https://github.com/filecoin-project/ref-fvm/issues/1513

### DIFF
--- a/FIPS/fip-0055.md
+++ b/FIPS/fip-0055.md
@@ -340,7 +340,7 @@ We do so as follows:
 | Pos   | Filecoin Message Field | Source                                                                                                                                      |
 | ----- |------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|
 | 0     | Version                | 0, the only supported Message version on Filecoin as of this FIP                                                                            |
-| 1     | Recipient              | Filecoin address converted from the `EthereumTransaction#To`, or `f010` if nil as this denotes a contract deployment                        |
+| 1     | Recipient              | Filecoin address converted (f410 or ID address) from the `EthereumTransaction#To` field, or `f010` if nil as this denotes a contract deployment                        |
 | 2     | Sender                 | Copied from the original `FilecoinMessage#Sender`, since this field is not present in the `EthereumTransaction`                             |
 | 3     | Nonce                  | `EthereumTransaction#Nonce`                                                                                                                 |
 | 4     | Value                  | `EthereumTransaction#Value`                                                                                                                 |
--- a/FIPS/fip-0055.md
+++ b/FIPS/fip-0055.md
@@ -33,7 +33,6 @@ replaces: N/A
       - [Create2](#create2)
       - [CreateExternal](#createexternal)
   - [Delegated signature type](#delegated-signature-type)
-    - [Message restrictions](#message-restrictions)
     - [Transitory validation](#transitory-validation)
     - [Payload reconstitution (FilecoinMessage -> RLP-encoding)](#payload-reconstitution-filecoinmessage---rlp-encoding)
     - [Calculating the inverse (RLP-encoding -> FilecoinMessage)](#calculating-the-inverse-rlp-encoding---filecoinmessage)
@@ -280,19 +279,19 @@ Delegated signatures are opaque to the protocol, and are validated by actor logi
 Delegated signatures will be used in combination with abstract account senders, once [Account Abstraction] (AA) is fully realized.
 However, we establish a transitory period where Delegated signatures are special-cased.
 
-#### Message restrictions
-
-Only select messages are allowed to carry delegated signatures. These are:
-
-1. Messages to the Ethereum Address Manager (`f010`, calling its `CreateExternal` method, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
-2. `Send` messages with empty parameters
-3. Messages with `MethodNum` equal to the EVM built-in actor's `InvokeContract` number, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
-
-A message carrying a delegated signature that does not meet one of these criteria is considered to have an invalid signature.
-
 #### Transitory validation
 
-We establish a transitory period whereby `Delegated` signatures are assumed to be secp256k1 signatures over the [RLP] representation of a native [EIP-1559 Ethereum transaction] sent from an Ethereum Acount actor.
+We establish a transitory period whereby `Delegated` signatures are assumed to be secp256k1 signatures over the [RLP] representation of a native [EIP-1559 Ethereum transaction] sent from an Ethereum Account actor.
+During this period, all senders of messages with `Delegated` signatures must have an `f410` address, that is, an address under the EAM's namespace. 
+
+Additionally, delegated signatures are restricted to:
+
+1. Messages to the Ethereum Address Manager (`f010`), calling its `CreateExternal` method, and carrying a non-empty, validly CBOR-encoded byte array as its parameters (with no extra bytes)
+1. `Send` messages with empty parameters
+1. Messages with `MethodNum` equal to the EVM built-in actor's `InvokeContract` number, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
+
+A message carrying a `Delegated` signature that does not meet one of these criteria is considered to have an invalid signature.
+
 The RLP representation is reconstituted from the Filecoin message as specified below.
 This effectively couples the `Delegated` signature type to its only possible use at this stage, but provides an extension point for the future introduction of [Account Abstraction] (AA).
 
--- a/FIPS/fip-0055.md
+++ b/FIPS/fip-0055.md
@@ -29,9 +29,14 @@ replaces: N/A
     - [State](#state)
     - [`Constructor` (method number `1`)](#constructor-method-number-1)
     - [Contract factory methods](#contract-factory-methods)
+      - [Create](#create)
+      - [Create2](#create2)
+      - [CreateExternal](#createexternal)
   - [Delegated signature type](#delegated-signature-type)
+    - [Message restrictions](#message-restrictions)
     - [Transitory validation](#transitory-validation)
-    - [Payload reconstitution](#payload-reconstitution)
+    - [Payload reconstitution (FilecoinMessage -> RLP-encoding)](#payload-reconstitution-filecoinmessage---rlp-encoding)
+    - [Calculating the inverse (RLP-encoding -> FilecoinMessage)](#calculating-the-inverse-rlp-encoding---filecoinmessage)
   - [Ethereum Account](#ethereum-account)
     - [Installation and wiring](#installation-and-wiring-1)
     - [Sender account](#sender-account)
@@ -50,6 +55,7 @@ replaces: N/A
   - [Essential for ecosystem compatibility](#essential-for-ecosystem-compatibility)
   - [Explorer adaptation](#explorer-adaptation)
   - [Exchange adaptation](#exchange-adaptation)
+  - [Restrictions on Delegated Signature payload](#restrictions-on-delegated-signature-payload)
 - [Implementation](#implementation)
 - [Copyright](#copyright)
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0055.md
```diff
@@ -274,6 +274,16 @@ Delegated signatures are opaque to the protocol, and are validated by actor logi
 Delegated signatures will be used in combination with abstract account senders, once [Account Abstraction] (AA) is fully realized.
 However, we establish a transitory period where Delegated signatures are special-cased.
 
+#### Message restrictions
```
raulk, line 277: Let's explain the product motivations behind these restrictions. Otherwise they appear rather arbitrary to the reader. Suggest to do that in the Design Rationale section to avoid overloading this section.

raulk, line 277: Editor comment: Needs ToC update.

arajasek: Done, please take a look.

FIPS/fip-0055.md
```diff
@@ -274,6 +280,16 @@ Delegated signatures are opaque to the protocol, and are validated by actor logi
...
+
+Only select messages are allowed to carry delegated signatures. These are:
+
+1. Messages to the Ethereum Address Manager (`f010`, calling its `CreateExternal` method, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
+2. `Send` messages with empty parameters
+3. Messages with `MethodNum` equal to the EVM built-in actor's `InvokeContract` number, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
```
raulk, line 289: Should we be checking the type of the destination actor here?

FIPS/fip-0055.md
```diff
@@ -306,10 +323,32 @@ The top-level object is an [RLP] list, and integers are encoded in the RLP way (
...
+The Input data must be empty if the `FilecoinMessage#Method` indicates a `Send`, and must be a CBOR-encoded byte array with no extra bytes otherwise. 
+If these conditions are not met, the signature is considered invalid.
+
+#### Calculating the inverse (RLP-encoding -> FilecoinMessage)
+
+To fully guarantee fidelity of the signature over the RLP-encoded payload, we re-calculate the Filecoin message, 
+as an inverse of the payload calculated above.
+If the CBOR-serialization of the inverse does not precisely match that of the
+original Filecoin message, the signature is considered invalid.
+We do so as follows:
```
raulk, line 338: editor nit: deviating from MUST level [semantic line breaks](https://sembr.org/)

FIPS/fip-0055.md
```diff
@@ -306,10 +323,32 @@ The top-level object is an [RLP] list, and integers are encoded in the RLP way (
...
+
+| Pos   | Filecoin Message Field | Source                                                                                                                                      |
+| ----- |------------------------|---------------------------------------------------------------------------------------------------------------------------------------------|
+| 0     | Version                | 0, the only supported Message version on Filecoin as of this FIP                                                                            |
+| 1     | Recipient              | Filecoin address converted from the `EthereumTransaction#To`, or `f010` if nil as this denotes a contract deployment                        |
```
raulk, line 343:
```suggestion
| 1     | Recipient              | Filecoin address converted (f410 or ID address) from the `EthereumTransaction#To` field, or `f010` if nil as this denotes a contract deployment                        |
```

FIPS/fip-0055.md
```diff
@@ -306,10 +323,32 @@ The top-level object is an [RLP] list, and integers are encoded in the RLP way (
...
+| 2     | Sender                 | Copied from the original `FilecoinMessage#Sender`, since this field is not present in the `EthereumTransaction`                             |
```
raulk, line 344: Isn't this recovered from the signature?

FIPS/fip-0055.md
```diff
@@ -306,10 +323,32 @@ The top-level object is an [RLP] list, and integers are encoded in the RLP way (
...
+| 4     | Value                  | `EthereumTransaction#Value`                                                                                                                 |
+| 5     | Gas Limit              | `EthereumTransaction#GasLimit`                                                                                                              |
+| 6     | Gas Fee Cap            | `EthereumTransaction#GasFeeCap`                                                                                                             |
+| 7     | Gas Premium            | `EthereumTransaction#MaxPriorityFeeGas`                                                                                                     |
+| 8     | Method                 | `EAM::CreateExternal`, if `EthereumTransaction#To` is nil. `Send`, if `EthereumTransaction#Input` is empty. `EVM::InvokeContract` otherwise |
+| 8     | Params                 | CBOR-serialization of the byte array `EthereumTransaction#Input`, which must not be empty if `EthereumTransaction#Recipient` is nil                                                                                                                 |
```
raulk, line 351:
```suggestion
| 9     | Params                 | CBOR-serialization of the byte array `EthereumTransaction#Input`, which must not be empty if `EthereumTransaction#Recipient` is nil         |
```

FIPS/fip-0055.md
```diff
@@ -442,6 +481,16 @@ Exchanges must support receiving funds from EVM smart contracts.
...
 
+### Restrictions on Delegated Signature payload
+
+As specified above, we place several restrictions on messages that carry a delegated signature, limiting them to 
+certain specific Ethereum-related cases. This is because this FIP introduces a transitory period where 
+Delegated signatures are special-cased to meet the product goals of the FEVM. 
+Placing these restrictions allows us to avoid compromising on the longer-term goals of Account Abstraction,
+while still immediately enabling compatibility with the existing Ethereum ecosystem.
+
+These restrictions will be removed as we upgrade towards Account Abstraction.
```
raulk, line 492: editor nit: MUST-level https://sembr.org/

FIPS/fip-0055.md
arajasek: So, we could. My biggest objection to doing so is it would be the first time we place restrictions on message validity based on the recipient. We currently don't have any restrictions on message _validity_ based on the recipient (beyond that it's a well-formed Filecoin address).

We also don't have any checks based on actor type (code CID) in Filecoin today (on either sender or recipient). 

These aren't reasons not to make this change, but these are fairly critical parts of the system that we need to carefully reason about before modifying.

All in all, I would advocate for not making this change. It feels like an unnecessary change, and I think the MethodNum restriction is sufficient. But I'm definitely open to it.

arajasek: (We don't do it in the current implementation, but that can of course change to match the spec we align on here).

FIPS/fip-0055.md
arajasek: I don't _think_ we can do that because of ID-address senders. It should be possible to send messages with ID-address senders carrying delegated signature payloads.

We could recover from the signature and do an equality check _with_ resolution. But honestly the sender isn't one of the interesting fields to inspect when round-tripping, that's the one field we "know" is correct because of the siggy validation.

FIPS/fip-0055.md
arajasek: No, it doesn't? Which MUST rule does this deviate from?

Also note that Sembr is one of [many](https://mtsknn.fi/blog/4-1-wrapping-styles-for-markdown-prose-and-code-comments/#the-4-1-wrapping-styles) guidelines, and we haven't settled on one for FIPs -- I would be in favour of picking one. Right now the only convention we have is limiting line-length to 80-100 characters (eg. [FIP-0032](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0032.md)), which this FIP _doesn't_ do, but we haven't formalized it so :woman_shrugging: .

raulk: It's not the time to discuss a general style guide for FIPs, but this FIP uses Sembr at the MUST level (i.e. only full stops break lines). We should avoid mixing styles in a single FIP because it gets messy.

raulk: Where is the 80-100 max width convention? The first three FIPs (FIP-0001, FIP-0002, FIP-0003) don't use it. The next FIP, FIP-0004, uses Sembr at the MUST level too.

FIPS/fip-0055.md
```diff
@@ -274,6 +280,16 @@ Delegated signatures are opaque to the protocol, and are validated by actor logi
 Delegated signatures will be used in combination with abstract account senders, once [Account Abstraction] (AA) is fully realized.
 However, we establish a transitory period where Delegated signatures are special-cased.
 
+#### Message restrictions
```
raulk, line 283: This section should be under **transitory validation**. Otherwise it can be interpreted that these restrictions apply to the long-term **Delegated** signature type generally, which isn't the case.

FIPS/fip-0055.md
raulk: Keeping it as-is enables Eth accounts to send Ethereum native transactions to future native actors that expose an EVM compatible interface (on the `InvokeEVM` method) (assuming we move this logic to the abstract account once it exists).

FIPS/fip-0055.md
raulk: We definitely do that in the JSON-RPC endpoint. Which is also why the Sender is always an f410 address. The siginature validation logic can't work with ID addresses.

https://github.com/filecoin-project/lotus/blob/e65fae28de2a44947dd24af8c7dafcade29af1a4/lib/sigs/sigs.go#L40

FIPS/fip-0055.md
```diff
@@ -274,6 +280,16 @@ Delegated signatures are opaque to the protocol, and are validated by actor logi
...
+
+Only select messages are allowed to carry delegated signatures. These are:
+
+1. Messages to the Ethereum Address Manager (`f010`, calling its `CreateExternal` method, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
```
raulk, line 287: We are missing the constraint that the sender must be an f410 address.

FIPS/fip-0055.md
arajasek: >  but this FIP uses Sembr at the MUST level (i.e. only full stops break lines)

I'm not sure what you mean by "using Sembr at the MUST level", but no part of the Sembr spec implies "only full stops break lines". In particular, (quoting from the spec) `A semantic line break MUST occur after a sentence...` does NOT mean `A semantic line break MUST [ONLY] occur after a sentence...` See [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt) for more.

I'm willing to go with the style you have in this FIP so far, but I do want to stress that:
- it is the opposite of what is recommended by Sembr (see point 12)
- it is generally against most style guidelines
- "using Sembr at the MUST level" isn't a thing.

arajasek: > Sidenote: Where is the 80-100 max width convention? The first three FIPs (FIP-0001, FIP-0002, FIP-0003) don't use it. The next FIP, FIP-0004, uses Sembr at the MUST level too.

It's a newer thing -- pretty much every FIP in the last year or so (since FIP-0030), with the exception of the recent FEVM FIPs). 

But certainly not an _actual_ standard yet, which I would like to see adopted.

FIPS/fip-0055.md
arajasek: > The siginature validation logic can't work with ID addresses.

It can -- an ID address will be [resolved to its deterministic address](https://github.com/filecoin-project/lotus/blob/cdf3812e4081d35692b5ab81c573ab82d0df69cf/chain/stmgr/stmgr.go#L233) before signature validation.

This is _definitely_ how it works for f1 / f3 addresses today, and as far as I can tell is still true of f4 addresses as implemented in Lotus right now. 

> We definitely do that in the JSON-RPC endpoint. Which is also why the Sender is always an f410 address. 

What the JSON-RPC endpoint does is irrelevant here, and doesn't make any guarantees about what lands on chain.

FIPS/fip-0055.md
arajasek: Yup, will add that!

kaitlin-beegle:
Going to merge so we can begin last call! If there are continued conversations around formatting, etc., we can make those changes.

PR #606, opened 2023-01-27
FIP0055: Update based on user feedback

Stebalien:
- Always invoke the target actor, never "demote" to a bare-value send.
- Accept all method numbers (>= 2^24) on both account types.

This PR also fleshes out some details and fixes a couple of errors.

### DIFF
--- a/FIPS/fip-0055.md
+++ b/FIPS/fip-0055.md
@@ -332,7 +332,7 @@ Additionally, this conversion asserts that:
 
 1. The sender address is a valid `f410` address with a 20 byte payload.
 2. The recipient is either a valid `f410` address or a valid `f0` address.
-3. If the receiver is `f010` (EAM), `FilecoinMessage#Method` is `EAM#CreateExternal` (`4`). Otherwise, `FilecoinMessage#Method` is `EVM#InvokeContract` (`3844450837`).
+3. If the recipient is `f010` (EAM), `FilecoinMessage#Method` is `EAM#CreateExternal` (`4`). Otherwise, `FilecoinMessage#Method` is `EVM#InvokeContract` (`3844450837`).
 4. The message version is 0.
 5. The parameters contain no additional values or bytes beyond the CBOR encoded byte string.
 
--- a/FIPS/fip-0055.md
+++ b/FIPS/fip-0055.md
@@ -287,8 +287,7 @@ During this period, all senders of messages with `Delegated` signatures must hav
 Additionally, delegated signatures are restricted to:
 
 1. Messages to the Ethereum Address Manager (`f010`), calling its `CreateExternal` method, and carrying a non-empty, validly CBOR-encoded byte array as its parameters (with no extra bytes)
-1. `Send` messages with empty parameters
-1. Messages with `MethodNum` equal to the EVM built-in actor's `InvokeContract` number, and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes)
+2. Messages to ID addresses (`f0`) other than `f010`, and messages to `f410` addresses (`f4` addresses assigned by the EAM) with `MethodNum` equal to the EVM built-in actor's `InvokeContract` method number  (`frc42_hash("InvokeEVM")`), and carrying a validly CBOR-encoded byte array as its parameters (with no extra bytes).
 
 A message carrying a `Delegated` signature that does not meet one of these criteria is considered to have an invalid signature.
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0055.md
```diff
@@ -339,15 +345,15 @@ We do so as follows:
...
 | 6     | Gas Fee Cap            | `EthereumTransaction#GasFeeCap`                                                                                                             |
 | 7     | Gas Premium            | `EthereumTransaction#MaxPriorityFeeGas`                                                                                                     |
-| 8     | Method                 | `EAM::CreateExternal`, if `EthereumTransaction#To` is nil. `Send`, if `EthereumTransaction#Input` is empty. `EVM::InvokeContract` otherwise |
-| 9     | Params                 | CBOR-serialization of the byte array `EthereumTransaction#Input`, which must not be empty if `EthereumTransaction#Recipient` is nil         |
+| 8     | Method                 | `EAM::CreateExternal`, if `EthereumTransaction#To` is nil. `EVM::InvokeContract` otherwise                                                  |
+| 9     | Params                 | CBOR-serialization of the byte array `EthereumTransaction#Input`.                                                                           |
```
Stebalien, line 356: I'm not sure why we had the second restriction as it's technically valid to deploy empty bytecode, as far as I know.

anorth (approved):

FIPS/fip-0055.md
```diff
@@ -320,12 +319,19 @@ The top-level object is an [RLP] list, and integers are encoded in the RLP way (
...
+| 7     | Input data            | CBOR-deserialization of `FilecoinMessage#Params` as a CBOR byte string, and re-encoded as an RLP byte string                                                                                                                                               |
 | 8     | Access list           | Empty.                                                                                                                                                                   |
 
-The Input data must be empty if the `FilecoinMessage#Method` indicates a `Send`, and must be a CBOR-encoded byte array with no extra bytes otherwise. 
+Additionally, this conversion asserts that:
+
+1. The sender address is a valid `f410` address with a 20 byte payload.
+2. The recipient is either a valid `f410` address or a valid `f0` address.
+3. If the receiver is `f010` (EAM), `FilecoinMessage#Method` is `EAM#CreateExternal` (`4`). Otherwise, `FilecoinMessage#Method` is `EVM#InvokeContract` (`3844450837`).
```
arajasek, line 331:
```suggestion
3. If the recipient is `f010` (EAM), `FilecoinMessage#Method` is `EAM#CreateExternal` (`4`). Otherwise, `FilecoinMessage#Method` is `EVM#InvokeContract` (`3844450837`).
```

FIPS/fip-0055.md
```diff
@@ -320,12 +319,19 @@ The top-level object is an [RLP] list, and integers are encoded in the RLP way (
...
-| 7     | Input data            | CBOR-deserialization of `FilecoinMessage#Params`, and re-encoded as an RLP byte string                                                                                                                                               |
+| 7     | Input data            | CBOR-deserialization of `FilecoinMessage#Params` as a CBOR byte string, and re-encoded as an RLP byte string                                                                                                                                               |
 | 8     | Access list           | Empty.                                                                                                                                                                   |
 
-The Input data must be empty if the `FilecoinMessage#Method` indicates a `Send`, and must be a CBOR-encoded byte array with no extra bytes otherwise. 
+Additionally, this conversion asserts that:
+
+1. The sender address is a valid `f410` address with a 20 byte payload.
+2. The recipient is either a valid `f410` address or a valid `f0` address.
```
arajasek, line 330: If you care, the recommended way to write numbered Markdown lists is to number all items as `1.` (will be rendered based on the actual position in the list). Makes it easier to move items around if necessary.

arajasek (commented):
This change makes it _appear_ that f4 addresses will not be able to perform bare sends to account actors, miner actors, or even other EthAccount actors. My understanding is that this will not be the case, at least for accounts and EthAccounts, because they will happily accepted all unimplemented Methods ([PR](https://github.com/filecoin-project/builtin-actors/pull/1086)). Are we making this change explicit in some other FIP / can we mention that here?

Also _do_ we care about losing the ability to `Send` to miners / multisigs / etc. with delegated siggys? I don't think so, and we can always add it later if needed, but wanted to discuss.

Schwartz10:
@arajasek i may be misunderstanding, but from GLIF pov, we definitely need to enable sends to account actors. A smart contract in the FEVM needs to be able to send funds to a miner actor

Stebalien:
> Are we making this change explicit in some other FIP / can we mention that here?

Hm. Good point. I was going to add that to FIP0054, but that change makes more sense here. I'll update.

Stebalien:
> i may be misunderstanding, but from GLIF pov, we definitely need to enable sends to account actors. A smart contract in the FEVM needs to be able to send funds to a miner actor for example to provide pledge collateral

A smart contract can send funds to a miner actor using the `call_actor` precompile. However, as implemented, an EthAccount cannot.

We have a few options:

1. Do nothing.
2. Implement `InvokeEVM` on the miner actor.
3. Something like https://github.com/filecoin-project/FIPs/discussions/608.

arajasek:
> @arajasek i may be misunderstanding, but from GLIF pov, we definitely need to enable sends to account actors. A smart contract in the FEVM needs to be able to send funds to a miner actor for example to provide pledge collateral

Thanks for the engagement, it's actually very good to have your input here! This FIP is about top-level messages and not calls from contracts -- so contracts can still send funds to a miner etc.

However, "Ethereum accounts", which are basically secp keys being used with the Ethereum addressing scheme, will not be able to perform a simple transfer to miner actors. They could totally do so through a smart contract, but that's a work around.

I'm very interested to know how much this would affect your plans, if at all. We're a little in the dark about potential use cases.

Schwartz10:
Ok thanks for engaging! I need to give it more thought, but my gut instincts are:

- "doing nothing" doesn't seem like it would restrict the functionality of a protocol (i dont think it makes anything impossible that would be possible with the functionality enabled).
- if ^^ holds, enabling this feature is not launch critical as long as we can upgrade into making this possible later
- I do see how it could lead to an undesireable amount of bad/clunky UX and larger headwinds in terms of education requirements, so the feature should be upgraded into asap

In the meantime, a decent workaround might be a permissionless and un-governable relayer that's deployed to FEVM which allows you to encode some recipient address in its calldata and just send all funds through that deployed contract?

Stebalien:
The current proposal from the FVM team is to "do nothing". Basically:

1. Smart contracts _can_ send funds to Miners, they just have to use the `call_actor` precompile.
2. Ethereum Accounts won't be able to (directly) send funds to miners, but we can (as you say) implement this as a relay. However, this would be deployed as a user contract and doesn't need to be part of any FIPs. Furthermore, we don't expect this to be a common use-case (storage providers will almost certainly continue to use native accounts).

kaitlin-beegle:
Thanks @Stebalien

I assume the initial PR, then, is still correct, and no additional changes are needed. Correct? If so, I'll review and merge.

Stebalien:
> I assume the initial PR, then, is still correct, and no additional changes are needed. Correct? If so, I'll review and merge.

I need to add some (minor) changes to the account actor. I'll get to that once my current 1:1s finish.

raulk:
@Schwartz10 I tagged you on a separate thread on Slack, but would appreciate your eyes on this post as it'll clear up any residual doubts: https://github.com/filecoin-project/FIPs/discussions/592#discussioncomment-4819619

Stebalien:
Note: I realized I also left out some rather important changes to accounts: Accounts (ethereum and regular) now accept calls on all method numbers >= 2^24. This means that using  `CALL` to transfer to an account will always work.

FIPS/fip-0055.md
Stebalien: I do care and seriously object to that recommendation (it makes reading markdown documents in an editor painful).

anorth (approved):

kaitlin-beegle (approved):

PR #607, opened 2023-01-27
FIP0054: Simplify and expand CALL logic

Stebalien:
We no longer special-case depending on the target actor type in CALL and have changed the logic around solidity "transfers" to CALL with 10M Filecoin Gas.

I've also taken the opportunity to expand on the other call variants a bit, and clarify how Filecoin "send" parameters are derived.

### DIFF
--- a/FIPS/fip-0054.md
+++ b/FIPS/fip-0054.md
@@ -578,15 +578,17 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
 
 First, if the recipient is a _precompile_, FEVM invokes the precompile directly as follows:
 
-1. It invokes the precompile, ignoring the gas limit specified in the `CALL` arguments.
-2. On success, if the user has requested a non-zero value transfer, FEVM deposits the funds in the precompile's account, FEVM transfers said funds to the precompiles account.
+1. It invokes the precompile.
+    1. The `call_actor` precompile applies the specified gas limit (if any) to the subsequent actor call.
+    2. All other precompiles ignore the specified gas limit.
+2. On success, if the user has requested a non-zero value transfer, FEVM transfers the funds to the precompile's [Filecoin address](#addressing) (which will auto-create a Placeholder at that address, if necessary).
 
 Otherwise, FEVM converts the `CALL` operation into a Filecoin "send" as follows:
 
 1. The method number is always `EVM::InvokeContract` (3844450837).
 2. The receiver's address is converted into a [Filecoin address](#addressing).
-3. If input data is treated as `IPLD_RAW` if non-empty, or the empty block (block `0`) if empty.
-4. The value is treated as a Filecoin token amount.
+3. The input data is treated as `IPLD_RAW` if non-empty, or the empty block (block `0`) if empty.
+4. The value is treated as a Filecoin token amount, as per [Native currency](#native-currency).
 5. The gas limit is computed as follows:
     1. If the value zero and the gas limit is 2300, or the gas limit is zero and the value is non-zero, FEVM sets the gas limit to 10M. This ensures that bare "transfers" continue to work as 2300 isn't enough gas to perform a call on Filecoin. 10M was picked to be enough to "always work" but not so much that it becomes a security concern.
     2. FEVM then applies the 1/64th rule from Ethereum, limiting the the sent gas to at most 63/64 of the remaining gas.
@@ -604,14 +606,14 @@ Solidity no longer supports `CALLCODE`.
 
 `DELEGATECALL` behaves differently depending on the recipient:
 
-1. If the target is a precompile address, it is handled internally according to the precompile logic defined in `CALL` above.
-2. If the target actor is an EVM runtime actor, it fetches the bytecode of the target and calls `InvokeContractDelegate` on itself to create a new call stack "frame" whilst still operating on our own state.
-3. If the target is an account, placeholder, or an EthAccount actor ([FIP-0055]), it returns nothing and success (pushes 1 onto the stack).
-4. If the target is any other type of actor, it returns nothing and failure (pushes 0 onto the stack).
+- If the target is a precompile address, it is handled internally according to the precompile logic defined in `CALL` above.
+- If the target actor is an EVM runtime actor, it fetches the bytecode of the target and calls `InvokeContractDelegate` on itself to create a new call stack "frame" whilst still operating on our own state.
+- If the target is an account, placeholder, or an EthAccount actor ([FIP-0055]), it returns nothing and success (pushes 1 onto the stack).
+- If the target is any other type of actor, it returns nothing and failure (pushes 0 onto the stack).
 
 **Calls: `STATICCALL`.** 
 Performs the `CALL` in read-only mode, disallowing state mutations, event emission, and actor deletions.
-See [`send::send`](#sendsend-syscall) for more information. Specifically, it sets the "read-only" bit in the send flags (bit `0x1`).
+Specifically, it sets the "read-only" bit in the send flags (bit `0x1`). See [`send::send`](#sendsend-syscall) for more information.
 
 **External code: `EXTCODESIZE`.**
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0054.md
```diff
@@ -575,38 +575,43 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
...
+2. On success, if the user has requested a non-zero value transfer, FEVM deposits the funds in the precompile's account, FEVM transfers said funds to the precompiles account.
+
+Otherwise, FEVM converts the `CALL` operation into a Filecoin "send" as follows:
+
+1. The method number is always `EVM::InvokeContract` (3844450837).
+2. The receiver's address is converted into a [Filecoin address](#addressing).
+3. If input data is treated as `IPLD_RAW` if non-empty, or the empty block (block `0`) if empty.
+4. The value is treated as a Filecoin token amount.
+5. The gas limit is computed as follows:
+    1. If the value zero and the gas limit is 2300, or the gas limit is zero and the value is non-zero, FEVM sets the gas limit to all remaining gas. This ensures that bare "transfers" continue to work as 2300 isn't enough gas to perform a call on Filecoin.
```
Stebalien, line 591: The alternative is to go back to:

1. Not attempt to forbid method 0.
2. Turn this back into a method-0 send.

I'm currently waiting on product feedback.

Stebalien: I went with "10M gas limit".

FIPS/fip-0054.md
```diff
@@ -575,38 +575,43 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
...
-If the gas limit is 2300, no matter the target's type, the call is lowered to a bare value send.
-This is to honor the well-known Ethereum convention of setting a gas limit equal to 2300 to prevent any possible contract logic from running.
+First, if the recipient is a _precompile_, FEVM invokes the precompile directly as follows:
 
-The 1/64th gas reservation rule from Ethereum is honored.
-CALL opcodes specifying no gas limit will implicitly set a 63/64th gas limit for the callee, except when the callee is a precompile (since precompiles are resolved internally).
-Note that the reservation is applied on Filecoin gas, not Ethereum gas.
-See [Product considerations: Gas](#gas) for more information.
+1. It invokes the precompile, ignoring the gas limit specified in the `CALL` arguments.
+2. On success, if the user has requested a non-zero value transfer, FEVM deposits the funds in the precompile's account, FEVM transfers said funds to the precompiles account.
```
raulk, line 582: Refinement: we don't expect these precompiles to ever be _accounts_ (i.e. native accounts or Eth accounts); we expect them to stay as placeholders forever (otherwise, uh-oh, trouble in crypto land).

I think this would be clearer saying that the Filecoin EVM transfers said funds to the equivalent `[Filecoin address](#addressing)`, which in practice will either create a placeholder, or already be a placeholder.

FIPS/fip-0054.md
```diff
@@ -575,38 +575,43 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
...
 **Calls: `DELEGATECALL`.**
-Fetches the bytecode of the target, as long as it's an EVM runtime actor, and calls `InvokeContractDelegate` on itself to create a new call stack "frame" whilst still operating on our own state.
-If the target is an account, placeholder, or an EthAccount actor ([FIP-0055]), it returns nothing.
-If the target is another Wasm actor, we revert (aborting with exit code 33, i.e. `EVM_CONTRACT_REVERTED`).
-If the target is a precompile address, it is handled internally by first calling the precompile logic, and then transferring any value to the actual precompile address on-chain.
 
-A delegate call against a precompile has the same effect as `CALL`.
+`DELEGATECALL` behaves differently depending on the recipient:
+
+1. If the target is a precompile address, it is handled internally according to the precompile logic defined in `CALL` above.
```
raulk, line 607: On a deeper reflection, does this _transfer_ value too? In my head, those aren't the intended semantics of `DELEGATECALL`, but at this stage I wouldn't be surprised if this was a dark corner in vanilla EVM.

raulk, line 607: I tried to avoid numbering to not denote sequence, but rather a switch. Should we change to a plain bullet list?

FIPS/fip-0054.md
```diff
@@ -575,38 +575,43 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
...
+2. If the target actor is an EVM runtime actor, it fetches the bytecode of the target and calls `InvokeContractDelegate` on itself to create a new call stack "frame" whilst still operating on our own state.
+3. If the target is an account, placeholder, or an EthAccount actor ([FIP-0055]), it returns nothing and success (pushes 1 onto the stack).
+4. If the target is any other type of actor, it returns nothing and failure (pushes 0 onto the stack).
 
 **Calls: `STATICCALL`.** 
 Performs the `CALL` in read-only mode, disallowing state mutations, event emission, and actor deletions.
-See [`send::send`](#sendsend-syscall) for more information.
+See [`send::send`](#sendsend-syscall) for more information. Specifically, it sets the "read-only" bit in the send flags (bit `0x1`).
```
raulk, line 614:
```suggestion
Specifically, it sets the "read-only" bit in the send flags (bit `0x1`). See [`send::send`](#sendsend-syscall) for more information.
```

raulk (commented):
(This is a peer review and not an editor review. As an editor, I would merge this patch.)

FIPS/fip-0054.md
```diff
@@ -575,38 +575,43 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
...
-If the target is another Wasm actor, we revert (aborting with exit code 33, i.e. `EVM_CONTRACT_REVERTED`).
-If the target is a precompile address, it is handled internally by first calling the precompile logic, and then transferring any value to the actual precompile address on-chain.
 
-A delegate call against a precompile has the same effect as `CALL`.
+`DELEGATECALL` behaves differently depending on the recipient:
+
+1. If the target is a precompile address, it is handled internally according to the precompile logic defined in `CALL` above.
+2. If the target actor is an EVM runtime actor, it fetches the bytecode of the target and calls `InvokeContractDelegate` on itself to create a new call stack "frame" whilst still operating on our own state.
+3. If the target is an account, placeholder, or an EthAccount actor ([FIP-0055]), it returns nothing and success (pushes 1 onto the stack).
+4. If the target is any other type of actor, it returns nothing and failure (pushes 0 onto the stack).
```
anorth, line 610: [Discussion here](https://github.com/filecoin-project/FIPs/discussions/592#discussioncomment-4824103) about avoiding this remaining target-code-dependent behaviour switching.

FIPS/fip-0054.md
```diff
@@ -575,38 +575,43 @@ The contract isn't considered to be "dead" until it has a tombstone where the re
...
+2. On success, if the user has requested a non-zero value transfer, FEVM deposits the funds in the precompile's account, FEVM transfers said funds to the precompiles account.
+
+Otherwise, FEVM converts the `CALL` operation into a Filecoin "send" as follows:
+
+1. The method number is always `EVM::InvokeContract` (3844450837).
+2. The receiver's address is converted into a [Filecoin address](#addressing).
+3. If input data is treated as `IPLD_RAW` if non-empty, or the empty block (block `0`) if empty.
+4. The value is treated as a Filecoin token amount.
+5. The gas limit is computed as follows:
+    1. If the value zero and the gas limit is 2300, or the gas limit is zero and the value is non-zero, FEVM sets the gas limit to 10M. This ensures that bare "transfers" continue to work as 2300 isn't enough gas to perform a call on Filecoin. 10M was picked to be enough to "always work" but not so much that it becomes a security concern.
```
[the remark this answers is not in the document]
Stebalien: > it would make sense to add a point in Design rationale about having to maintain this equivalent eternally, as

I'll expand on it. The goal is to pick a number such that that it will never be an issue.

FIPS/fip-0054.md
Stebalien: It doesn't. `value` isn't even a parameter.

Stebalien: I consider switches to be sequential, but I'm happy to make it a list.

FIPS/fip-0054.md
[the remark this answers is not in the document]
Stebalien: I'm also happy to pick a higher number, like 100M. That would be 1% of the maximum block size.

anorth (approved):
Approving this diff as an editor, though I still have a technical discussion point open

kaitlin-beegle (approved):

PR #609, opened 2023-01-27
Specify float type in SDM function definition

vkalghatgi:


kaitlin-beegle (approved):

kaitlin-beegle:
Re: https://github.com/filecoin-project/FIPs/pull/571#discussion_r1071583990

Spoke with Vik, given Molly and Alex's sign-off, and the fact that they're both FIP editors, I'm going to merge and proceed with last call.  This is a minor spec change that should not required more detailed review. These types of details are usually amended during implementation.

PR #615, opened 2023-02-03
FIP0057: Update implementation & product considerations

Stebalien:
1. Include the results of the gas backtesting.
2. Link to the implementation.

### DIFF
--- a/FIPS/fip-0057.md
+++ b/FIPS/fip-0057.md
@@ -473,11 +473,22 @@ As with all gas schedule changes, this FIP introduces changes to the operation c
 
 ## Product Considerations
 
-This FIP will likely increase gas fees in some cases and needs to be carefully benchmarked.
+Overall, due to optimizations in the built-in actors, gas usage for most storage provider messages is expected to _decrease_ by ~30% on average (40% for  `PublishStorageDeals`).
+
+However, there are a couple of important cases to note:
+
+1. The message inclusion cost increases by 475,000 gas to account for updating the sending actor.
+2. Any non-trivial message will incur an additional 475,000 gas fee to update the receiving actor.
+3. Due to these changes, the cost of a simple value transfer from an account increases by ~300% (~0.5M gas to 1.5M gas).
+
+Furthermore, with respect to `PublishStorageDeals`:
+
+- PSD messages with multiple clients will be more expensive than PSD messages with fewer clients as there's a per-client 500k gas charge to lookup the client in question. However, this effect is minimal compared to the overall cost of these messages.
+- PSD messages that address clients by f1-f4 addressees will cost more than PSD messages that address said clients by their ID (f0 address) due to the cost of the additional lookup. However, again, the effects of this cost are minimal compared to the overall cost of these messages.
 
 ## Implementation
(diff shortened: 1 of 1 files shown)


FIPS/fip-0057.md
```diff
@@ -473,11 +473,22 @@ As with all gas schedule changes, this FIP introduces changes to the operation c
 
 ## Product Considerations
 
-This FIP will likely increase gas fees in some cases and needs to be carefully benchmarked.
+Overall, due to optimizations in the built-in actors, gas usage for most storage provider messages is expected to _decrease_ by ~30% on average (40% for  `PublishStorageDeals`).
```
anorth, line 476: Is this a statement about changes coming with FIP-0057, or orthogonal optimisation work that we did? I.e. is it about this FIP, or about network v18? If the latter, it probably doesn't belong here.

anorth (approved):

FIPS/fip-0057.md
Stebalien: > Is this a statement about changes coming with FIP-0057

This statement comes from:

1. Slightly lower per-block write fee.
2. Optimizations applied in the actors build system (wasn't officially a part of any FIP, although I can add it to this one, I guess).

I think it is important to note that any impact of this FIP is, for most use-cases, offset by other optimizations that will be shipping alongside it.

kaitlin-beegle: I agree that we should note the relationship this FIP has to expected gas fee changes, but the specifics of a change as caused by a non-FIP bug fix, enhancement, etc., can be accounted for in the Core Devs repo. 

A bit of a follow-on, but how formalized is network monitoring and benchmarking after an upgrade? Does EngRes or some other entity within Protocol Labs actively monitor network conditions for X amount of time, or is the monitoring more passive?  E.g., individuals check Starboard or other network usage dashboards, wait to ensure systems are working properly, and respond whenever a red flag goes up from some component of the protocol? I'm not sure how to make my question more specific so please let me know if it's unclear.  

FIPs often make reference to 'monitoring network impact', but I rarely see follow-up information about network effects post-implementation. That's not to say it isn't happening, it's just not visible to me, and I'm curious.

jennijuju: > but how formalized is network monitoring and benchmarking after an upgrade

Very good question @kaitlin-beegle. Yes - we(Lotus and Sentinel) develops  monitoring system for each upgrade, including regular network health metrics like block production and utilization, messages, and etc, and additional monitoring  according to the features/adjustment being added to the protocol from this upgrade. 
Core network deployment teams actively monitor the network for 12-24 hours(differs based on the size of each upgrade), and Lotus also continuously monitor the network health as the upgrade stabilizes over the week. We also develop alerting systems for things thats alarming for the network health.


All the information can be derived from the chain itself, and we welcome and encourage the community/ecosystem to develop toolings and monitor the network health along with PL engres. If this is something anyone would love to chat more about, feel free to reach me #fil-lotus-dev in [Filecoin slack](filecoin.io/slack).

jennijuju (approved):

PR #624, opened 2023-02-06
Add gas table to FIP0057

Kubuxu:


jennijuju:
@Kubuxu Thank you! do we have the value for PreCommitBatch and provecommitaggregation?

Kubuxu:
Not in the range of blocks I collected.

anorth (approved):
Editorial rubber stamp.

jennijuju:
@Kubuxu @Stebalien id still like some value for those three messages as they are critical storage consensus/operation  messages.

jennijuju (approved):

Kubuxu:
> id still like some value for those three messages as they are critical storage consensus/operation messages.

I'm aware, but the data collection I'm relaying on uses a live chain where these messages are not actively used.

@raulk might be able to collect some data on them with the standalone, on-demand migration software he is developing. Maybe we can apply it to a test vector of PCA and WindowPoSt dispute.

PR #635, opened 2023-02-20
Initial draft of the CDM proposal

ribasushi:


anorth:
An early editorial comment: while an enthusiastic expression is fine in the discussion forum, please tone it down in the body of the FIP document itself. Focus on the specification of the change and immediate implications on capabilities/incentives/security. It's ok to note trends, parties with different interests, potential new uses etc, but this document shouldn't really be selling anything.

FIPS/fip-draft_CappedDurationMultipler_CDM.md
```diff
@@ -0,0 +1,139 @@
+﻿---
+fip: "to be assigned"
+title: "Capped Duration Multiplier (CDM)"
+author: Peter R (@ribasushi)☨  Patrick D (@f8-ptrk)  (via superset) Fip36/56 authors
```
kaitlin-beegle, line 4: Were any of the authors of FIP0056 or FIP0036 consulted on the design of this? If not, let's remove reference to them from the authorship line. 

I think instead we should include reference to these FIPs in the FIP justification and background.  

In addition, I personally do not like the idea of this being considered a "superset" proposal, because I think it posits the CDM and SDM proposals and counterproposals to one another, and I'm not sure that's entirely accurate.  The argument could reasonably go either way; this is just my personal opinion.

ribasushi: @kaitlin-beegle authors were consulted in a way, but removing mention at your suggestion.

I also removed the "superset" text from the draft where it does not add value.

However I feel like I need to address a misconception I see repeated during the discourse. CDM or a similar **"two-pillar" mechanism can not be implemented once the gap between min-QAP and max-QAP for the same hardware greatly exceeds 10x**. If the 20x SDM, or a similar stacked-multiplier FIP proceeds beyond the acceptance stage, @f8-ptrk and I will be permanently withdrawing this CDM proposal, since it will no longer be actionable.

ribasushi:
CLosing draft-PR as per https://github.com/filecoin-project/FIPs/discussions/636#discussioncomment-5951784

PR #652, opened 2023-03-08
FIP-0063: Transition to new drand   :racing_car: `fastnet` network

AnomalRoil:
The new drand  :racing_car: `fastnet` network was launched on the League of Entropy mainnet nodes on March 1st and has already been running without any issue for the past week.
See for instance:
https://api.drand.sh/dbd506d6ef76e5f386f41c651dcb808c5bcbd75471cc4eafa3f4df7ad4e4c493/info

```
{
  "public_key": "a0b862a7527fee3a731bcb59280ab6abd62d5c0b6ea03dc4ddf6612fdfc9d01f01c31542541771903475eb1ec6615f8d0df0b8b6dce385811d6dcf8cbefb8759e5e616a3dfd054c928940766d9a5b9db91e3b697e5d70a975181e007f87fca5e",
  "period": 3,
  "genesis_time": 1677685200,
  "hash": "dbd506d6ef76e5f386f41c651dcb808c5bcbd75471cc4eafa3f4df7ad4e4c493",
  "groupHash": "a81e9d63f614ccdb144b8ff79fbd4d5a2d22055c0bfe4ee9a8092003dab1c6c0",
  "schemeID": "bls-unchained-on-g1",
  "metadata": {
    "beaconID": "fastnet"
  }
}
```

It features the new `bls-unchained-on-g1` scheme, which runs in "unchained" mode, meaning each round can now be verified independently from the previous ones, and has a longer public key but much shorter beacon signatures:
```
{
  "round": 1,
  "randomness": "ef076e4d0b9320bf3f50cb2940777ae6bbee79c3d620d8efc04195bfc0568486",
  "signature": "9544ddce2fdbe8688d6f5b4f98eed5d63eee3902e7e162050ac0f45905a55657714880adabe3c3096b92767d886567d0"
}
``` 

Discussions started here: https://github.com/filecoin-project/FIPs/discussions/567

No significant block or problem have been identified so far.

The four major changes would be:
- This is a 3 second drand network, which means Filecoin only needs every 10th beacon given its block time of 30 seconds.
- This is an unchained network, which allows for stateless verification, no need to know the previous signature and the previous beacons to verify it. As discussed in the Security section of the FIP this does not reduce the overall security of the network given drand's and the League of Entropy's current threat model which assumes there is never a threshold of malicious nodes at any time.
- It uses G2 for the public keys and G1 for the beacons' signatures, meaning shorter signatures and smaller beacons.
- The catch-up time was set to 2 seconds, which means the catch-up for 10 rounds would be 20 seconds instead of the current 15 seconds set on the default mainnet. Note that this can still be tweaked if deemed necessary, but we would prefer to stick to an integer for the catch-up time (hence why we did not set it to 1.5 seconds to keep the same as the current settings).

If anything, this new network is _more secure_ than the previous one, because its initial key generation ceremony relied on 21 nodes versus 14 for the `default` network.

### DIFF
--- a/FIPS/fip-0063.md
+++ b/FIPS/fip-0063.md
@@ -37,7 +37,7 @@ Currently drand verification is performed in two stages:
 1. by hashing the previous beacon’s signature and the current beacon round number using SHA256 and comparing the digest to the randomness in the payload, and
 2. verifying the signature over the hashed randomness is valid for the group public key (using BLS12-381).
 
-For signatures on the unchained scheme, the only change required is to omit the previous beacon’s signature when creating a SHA256 digest for comparison against the randomness. Note that the new network also have swaps the way it does BLS signatures, and uses G1 for signatures and G2 for public keys. This means shorter signatures, but it also means the signature verification algorithm needs to be updated in order to be instantiated over G1 instead of G2.
+For signatures on the unchained scheme, the only change required is to omit the previous beacon’s signature when creating a SHA256 digest for comparison against the randomness. Note that the new network also changes the way it does BLS signatures, and uses G1 for signatures and G2 for public keys. This means shorter signatures, but it also means the signature verification algorithm needs to be updated in order to be instantiated over G1 instead of G2.
 
 For Go implementations, this can be achieved by using the latest drand client library, and is transparent and handled by specifying the right “scheme” upon instantiation. At present there are no other language implementations supporting this, however they can be provided on demand (or existing open source implementations can be easily modified to be compliant).
 
--- a/FIPS/fip-0063.md
+++ b/FIPS/fip-0063.md
@@ -1,5 +1,5 @@
 ---
-fip: "to be assigned"
+fip: "FIP-0063"
 title: "Switching to new Drand mainnet network"
 author: "@yiannisbot, @CluEleSsUK, @AnomalRoil, @nikkolasg, @willscott"
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/567
--- a/FIPS/fip-00xx.md
+++ b/FIPS/fip-00xx.md
@@ -41,15 +41,9 @@ For signatures on the unchained scheme, the only change required is to omit the 
 
 For Go implementations, this can be achieved by using the latest drand client library, and is transparent and handled by specifying the right “scheme” upon instantiation. At present there are no other language implementations supporting this, however they can be provided on demand (or existing open source implementations can be easily modified to be compliant).
 
-The way beacons are retrieved will need to be adapted in order to fetch every 10th beacon to accommodate for Filecoin epochs: [chain/beacon/drand/drand.go](https://github.com/filecoin-project/lotus/blob/ccf1ba2b8a4a324744beb93d10b2162959aaf409/chain/beacon/drand/drand.go#LL132C25-L132C25)
-
-(Note that the drand team might provide an API at the library level for fetching rounds at a given different frequency, but this is not yet the case.)
-
-The following configurations will need to be adapted: 
-
-- [/build/drand.go#L45](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/drand.go#L45)
-- [/build/params_mainnet.go](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/params_mainnet.go#L17)
-- [drand configuration](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/drand.go#L45) will need to be amended to include the new `public key` and`chain hash` as well as to add the notion of `scheme` that determines how verification and signatures are done
+Additional minor changes will be needed in implementations:
+- the way beacons are retrieved will need to be adapted in order to fetch every 10th beacon to accommodate for Filecoin epochs
+- the new public key, chain hash and scheme will need updated to correctly verify beacon signatures
 
 ## Design Rationale
 
@@ -128,6 +122,14 @@ The new network is also producing smaller beacon signatures (half the size of th
 - Switching to a new chain is much easier in unchained randomness mode: it’s enough to just change the public key (bootstrapping the Filecoin chain becomes easier through use of the unchained beacon)
 
 ## Implementation
+
+The following Lotus implementation details have been highlighted:
+- the way beacons are retrieved will need to be adapted in order to fetch every 10th beacon to accommodate for Filecoin epochs: [chain/beacon/drand/drand.go](https://github.com/filecoin-project/lotus/blob/ccf1ba2b8a4a324744beb93d10b2162959aaf409/chain/beacon/drand/drand.go#LL132C25-L132C25); the drand team could provide an API at the library level for fetching rounds at a given different frequency, but this is not yet the case.
+
+The following configurations will need to be adapted: 
+- [/build/drand.go#L45](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/drand.go#L45)
+- [/build/params_mainnet.go](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/params_mainnet.go#L17)
+- [drand configuration](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/drand.go#L45) will need to be amended to include the new `public key` and`chain hash` as well as to add the notion of `scheme` that determines how verification and signatures are done
 
 <!--The implementations must be completed before any core FIP is given status "Final", but it need not be completed before the FIP is accepted. While there is merit to the approach of reaching consensus on the specification and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.-->
 
--- a/FIPS/fip-00xx.md
+++ b/FIPS/fip-00xx.md
@@ -5,9 +5,10 @@ author: @yiannisbot, @CluEleSsUK, @AnomalRoil, @nikkolasg, @willscott
 discussions-to: [https://github.com/filecoin-project/FIPs/discussions/567](https://github.com/filecoin-project/FIPs/discussions/567)
 status: Draft
 type: Core, Networking
+category: Core
 created: 2023-02-24
 spec-sections:
-- [https://spec.filecoin.io/#section-libraries.drand](https://spec.filecoin.io/#section-libraries.drand)
+ - [https://spec.filecoin.io/#section-libraries.drand](https://spec.filecoin.io/#section-libraries.drand)
 ---
 
 ## Simple Summary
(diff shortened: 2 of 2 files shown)


luckyparadise:
Hi @AnomalRoil my name is Lucky and I am the Governance Project Manager at Filecoin Foundation as well as the governance DRI for network upgrades. For this PR, what are you thinking in terms of landing this proposal (timelines)? Also, have you discussed this FIP with other teams?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,141 @@
...
+
+For further details on these new features and their implementation we refer the reader to [this blogpost](https://drand.love/blog/2022/02/21/multi-frequency-support-and-timelock-encryption-capabilities/) and the complementary, comprehensive [FAQ doc](https://docs.google.com/document/d/16QJG3Z-Kr0mN6snQz8cm0NnMXpYBpelKyvCf2oo1Zgc/edit?usp=sharing).
+
+## Change Motivation
+
+- The existing drand network is expected to be deprecated in Q4 2023 or Q1 2024.
+- The Filecoin network should switch to the new drand mainnet networks before the existing network is deprecated.
+- The new drand network is going to be producing **smaller signatures**, which means that Filecoin headers will be more lightweight.
```
jennijuju, line 30: How much smaller?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,141 @@
...
+- This new drand network also **allows to perform timelock encryption**, which will be a valuable feature for FVM.
```
jennijuju, line 31:
```suggestion
- This new drand network also **allows to perform timelock encryption**, which will be a valuable feature for user-deployed actors.
```

jennijuju: could you please provide an example of a potential use case?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Specification
+
+Currently drand verification is performed in two stages:
+
+1. by hashing the previous beacon’s signature and the current beacon round number using SHA256 and comparing the digest to the randomness in the payload, and
+2. verifying the signature over the hashed randomness is valid for the group public key (using BLS12-381).
+
+For signatures on the unchained scheme, the only change required is to omit the previous beacon’s signature when creating a SHA256 digest for comparison against the randomness. Note that the new network also have swaps the way it does BLS signatures, and uses G1 for signatures and G2 for public keys. This means shorter signatures, but it also means the signature verification algorithm needs to be updated in order to be instantiated over G1 instead of G2.
+
```
jennijuju, line 41: plz link the library, specifically the signature verification change

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,141 @@
...
+- URLs for accessing drand beacons from relays on the new network will require to include the chain hash (e.g. `[https://api.drand.sh/public/latest](http://api.drand.sh/public/latest)` may become something to the effect of `https://api.drand.sh/deadbeefcafebabe/public/latest`)
+- Filecoin implementations must consume every 10th round from drand rather than each consecutive round (to maintain a 30sec block time)
+- Beacons created with the scheme containing G1 and G2 swapped will also be unable to be verified by existing implementations.
+- Verification of beacons from existing schemes, including the G1 swap, are fully supported in the updated drand client.
+
+Therefore, switching to the new drand network needs to be done through a Filecoin network upgrade.
+
+## Test Cases
+
+In order to enable effective testing, we suggest creating a new, temporary testnets with a small number of clients from each widely supported Filecoin implementation.
```
jennijuju, line 69: Does fastnet has a testnet? Any reason why we need to start a new network instead of having filecoin testnet connects to fastnet (or its testnet) to test?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,141 @@
...
+
+Unchained randomness allows for lighter drand nodes since they effectively become “stateless” and can delegate storing historical beacons to Filecoin, IPFS or some other storage system. The longer we keep a “chained beacon” running, the bigger the burden on the nodes running it becomes over time (requires more storage, costs more, etc.) and the more likely we are to see partners leaving the League of Entropy and causing an undesirable centralization of drand nodes.
+
+The new network is also producing smaller beacon signatures (half the size of the current ones), which means lighter Filecoin headers.
+
+## Product Considerations
+
+**Future Product Consideration - why is the new drand network an improved network?**
+
+- The Timelock Encryption feature is only available when using unchained randomness, since we need to be able to predict which message will be signed at what time. Switching would allow us to integrate Timelock Encryption natively on Filecoin, once FVM lands. Expect another FIP focused on Timelock and FVM soon.
```
jennijuju, line 122:
```suggestion
- The Timelock Encryption feature is only available when using unchained randomness, since we need to be able to predict which message will be signed at what time. Switching would allow us to integrate Timelock Encryption natively on Filecoin. Expect another FIP focused on Timelock and FVM soon.
```

FIPS/fip-00xx.md
AnomalRoil: Which one do you prefer?
- helps preventing MEV (or do you say SPEV?)
- allows to run sealed bid auctions easily
- can help with airdrop and NFT minting too, to hide some data until a given time 
- data retention policy, where data is secret only for a given time, such as responsible disclosure of vulnerabilities

FIPS/fip-00xx.md
AnomalRoil: We're happy to have the filecoin testnet connecting to fastnet directly if that's fine on your side..

We do have the same scheme running in a testnet network too, but it's less resilient that our mainnet networks:
https://pl-us.testnet.drand.sh/f3827d772c155f95a9fda8901ddd59591a082df5ac6efe3a479ddb1f5eeb202c/info

jsoares: There's really no reason not to use fastnet for testing. On the Filecoin side, it's probably fine to go with testnet, yeah.

jsoares:
> Hi @AnomalRoil my name is Lucky and I am the Governance Project Manager at Filecoin Foundation as well as the governance DRI for network upgrades. For this PR, what are you thinking in terms of landing this proposal (timelines)? Also, have you discussed this FIP with other teams?

@luckyparadise Note that this is a bit different from most FIPs, in that it's an upstream service telling us to move with the times in order to keep the lights on. The key timeline is noted in the FIP's body ("The existing drand network is expected to be deprecated in Q4 2023 or Q1 2024."), though I guess there's still space to define a lower bound too. @AnomalRoil will have his own thoughts but I think we should give fastnet a ceremony period (read: a quarter) to make sure everything is stable prior to switching over, but adopting the FIP and testing should be done soon as _not adopting_ it would mean potential disruption.

And yes, other teams are aware of the need and plan to switch over. We certainly are at ConsensusLab, but I believe this has been presented somewhat extensively.

jennijuju:
> And yes, other teams are aware of the need and plan to switch over. 

Drand team has informed lotus team about the network deprecation ~mid 2022.

anorth:
Thanks, this is a solid first draft. I think there are some details missing for it to be implementable: a concrete specification of changes. Note that these needn't and shouldn't be referenced to the Lotus implementation, but specified such that all implementations can confidently make the required changes.

I could support merging this as a draft if people want to. The idea of switching networks is well understood and ready for community consideration. The actual mechanics to do so may need a little more detail.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,141 @@
+---
+fip: "to be assigned"
```
jennijuju, line 2: FIP-0063

jennijuju (approved):

jennijuju:
Im happy to merge this as a draft - FIP authors should later work with implementers closely to add more spec

momack2:
Not necessarily required in this FIP - but worth investigating at the same time what it would entail to cut down the Filecoin blocktime from 30s to something shorter (which has been heavily requested by various DeFi tools)

jsoares:
@momack2:
> Not necessarily required in this FIP - but worth investigating at the same time what it would entail to cut down the Filecoin blocktime from 30s to something shorter (which has been heavily requested by various DeFi tools)

 I agree it's worth exploring block time adjustments but it really should go in a later FIP due to the different nature of the problems:
- This one is a non-controversial dependency-driven update with zero-to-low downside. It's also something we can't _not_ do.
- Changing the Filecoin block time is an elective and complex trade-off that might turn out to be very controversial (or not; hard to guess).

Switching to the high-frequency drand network is a prerequisite to shorter block times (in the absence of deeper protocol changes), but the two are otherwise unrelated. It makes sense to get it merged cleanly and then, with the foundations in place, tackle the matter of the block time.

In principle, the block time adjustment is trivial in what concerns drand: merely a matter of changing the subsampling rate. The complexities will be on the Filecoin side (message propagation times, proof computation times, etc.).

FIPS/fip-0063.md
```diff
@@ -0,0 +1,141 @@
...
+- This new drand network also **allows to perform timelock encryption**, which will be a valuable feature for user-deployed actors.
+
+## Specification
+
+Currently drand verification is performed in two stages:
+
+1. by hashing the previous beacon’s signature and the current beacon round number using SHA256 and comparing the digest to the randomness in the payload, and
+2. verifying the signature over the hashed randomness is valid for the group public key (using BLS12-381).
+
+For signatures on the unchained scheme, the only change required is to omit the previous beacon’s signature when creating a SHA256 digest for comparison against the randomness. Note that the new network also have swaps the way it does BLS signatures, and uses G1 for signatures and G2 for public keys. This means shorter signatures, but it also means the signature verification algorithm needs to be updated in order to be instantiated over G1 instead of G2.
```
momack2, line 40:
```suggestion
For signatures on the unchained scheme, the only change required is to omit the previous beacon’s signature when creating a SHA256 digest for comparison against the randomness. Note that the new network also changes the way it does BLS signatures, and uses G1 for signatures and G2 for public keys. This means shorter signatures, but it also means the signature verification algorithm needs to be updated in order to be instantiated over G1 instead of G2.
```

What is the change from / to? Are there any security risks or considerations with the change?

CluEleSsUK: Essentially we just reverse two groups that we use for mapping purposes - it doesn't affect the security of the scheme. There are additional details in this PR for our docs updates: https://github.com/drand/website/pull/170

FIPS/fip-0063.md
```diff
@@ -0,0 +1,141 @@
...
+Upon successful testing, the network would be taken down and a new network `mixed-scheme-testnet` would be created. This network would be started using the existing drand chained network, and after creation of some blocks, would be migrated to the drand unchained network.
+
+Scenarios that must be run on this `mixed-scheme-testnet`are:
+
+- onboarding new nodes from existing snapshots (with chained randomness in the headers)
+- onboarding new nodes from newly created snapshots (containing both chained and unchained randomness in the headers)
+- transition of existing nodes from chained to unchained randomness at a set epoch
+- creation of new blocks with the unchained scheme
+
+## Security Considerations
```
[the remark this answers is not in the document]
CluEleSsUK: There are no security implications.
For syncing the past chain, two separate verification techniques will be needed. The client we provides does this transparently, so there shouldn't be any change on the Lotus side. 
For other implementations, it will depend how they do beacon verification, though we have some community implementations in e.g. rust that they can drop in.
Supporting multiple networks always has some historical precedent on filecoin, as the first bunch of blocks on filecoin actually run with randomness from drand dev (the scheme is the same, but the public key is different thus requiring a variation in the verification logic)

PR #673, opened 2023-03-22
Add Batch Balancer parameter recalibration to FIP-0057

AxCortesCubero:
FIP-0057 changes the estimated gas usage for PreCommit and ProveCommit messages, which appear as parameters in the batch balancer mechanism. If parameters are not adjusting accordingly following these changes, this would change the batching thresholds, in terms of high enough base fee and large enough number of proofs in a batch.

This change modifies FIP-0057 to include a corresponding update in the gas usage parameters of the Batch Balancer.

A sentence is added in the **Abstract** section, and a subsection is added at the end of the **Specification** section

### DIFF
--- a/FIPS/fip-0057.md
+++ b/FIPS/fip-0057.md
@@ -33,6 +33,8 @@ This FIP proposes a few adjustments to the gas charging schedule. Specifically:
 This FIP also introduces overall memory limits. However, these memory limits currently only include Wasm memories and table elements (not IPLD blocks, wasm _code_, and other metadata).
 
 Additionally, this FIP introduces a maximum block size limit of 1MiB for all newly created IPLD blocks.
+  
+Since the above changes the estimated gas usage of PreCommit and ProveCommit  messages, this FIP introduces a recalibration of the corresponding parameters in the Batch Balancer mechanism.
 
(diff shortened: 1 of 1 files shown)


anorth (approved):

arajasek (commented):
The change makes sense, but I"m confused about how we're doing this -- FIP-0057 is already live in the network.

Is this PR a _correction_ we're making to the text of FIP-0057 to reflect what already went live in nv18? Or a new change we're proposing making in nv19?

luckyparadise:
@arajasek it is an amendment/change to FIP0057 which the authors of this amendment are hoping to go live in nv19. Indeed a unique situation, but we agreed that the governance pathway would be to take this change through Last Call rather than have a new FIP as the proposed modification to maintain existing policy. I am not sure if this process makes sense from an implementation perspective.

jennijuju (approved):

Stebalien:
I'm going to _strongly_ object to this change. Cryptoecon changes should not be after-the-fact amendments.

Furthermore, this FIP was created to correct the gas model for nv18, while the issue fixed in this PR was addressing a change made in nv16.

PR #682, opened 2023-04-06
Update FIP0056 based on implementation

Stebalien:
- The minimum duration is 540 days, not 1.5 years. It's a small difference, but important.
- The implementation recalculates SDM from the current epoch on replica update (see https://github.com/filecoin-project/FIPs/discussions/554#discussioncomment-5232396).
- The `sdm` pseudo code lead to a maximum multiplier of 1e6, which was clearly a mistake.
- This now documents the update to replaced day rewards, etc.
- This now documents the correct termination penalty calculation (still under discussion, but the old version was broken).


FIPS/fip-0056.md
```diff
@@ -32,32 +32,49 @@ The following aims to help better align the economic incentives of the network w
...
+- Add a `power_base_epoch` field to miner_sector_infos. This field contains the epoch a sector power was last updated (sector commitment, extension, or replica update).
 
 ### SDM Function
-The SDM should scale initial pledge and power via the SDM function specified in pseudocode) below. Note, this will have to be implemented in fixed point arithmetic: 
+The SDM should scale initial pledge and power via the SDM function specified in pseudocode below. Specifically, it scales linearly from a 1x multiplier at 540 days to a 2x multiplier at 5 years.
 
 ```
-fn sdm(duration_commitment: float) -> float {
-    if duration_commitment <= (3/2 * EPOCHS_IN_YEAR) {
+fn sdm_scale(base_power: int, duration_commitment: int) -> int {
```
anorth, line 41: I think this should be `-> float`

anorth (approved):

anorth:
Thanks!

kaitlin-beegle (approved):
Reviewed and approving changes.  Even if the FIP is rejected, we should be sure to codify the most accurate implementation guidance.

PR #713, opened 2023-06-01
Governance Bot v1

ianconsolata:
## Proposed changes:
This introduces a governance bot github workflow to manage tags on the discussions in this repo, as requested by @kaitlin-beegle and @luckyparadise. This is step 1 of the governance discussion tag automation, more additions are planned for later this summer once requirements are more well scoped.

## Requirements to deploy / merge
This approach requires a token to be set. By default, github workflows sets a token on workflows, but, according to our testing, that token doesn't have appropriate permissions to update the tags on discussions. So instead, we set a repository secret (secrets.FILECOIN_GOVERNANCE_BOT_TOKEN) in @NateWebb03's test repo, and got it working using that token.

To add the bot to this repo, A new user needs to be configured with write access to this repo, and a scoped token should be generated for that user. We created the `filecoin-goverance` account for this, but they need to be granted the appropriate privileges, and then a scoped token needs to be generated and added to the repository secrets (different from the env secrets) by a maintainer: https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository. This token should be named `FILECOIN_GOVENANCE_BOT_TOKEN`.


.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: install python packages
+        run: |
+          python -m pip install --upgrade pip
+          cd .github/workflows 
+          pip install PyGithub
+          pip install datetime
+
+      - name: execute py script # run githublabeler.py
+        env:
+          TOKEN_SECRET: ${{ secrets.TOKEN_SECRET }}
```
ianconsolata, line 26: I think this needs to be `GITHUB_TOKEN: $ {{ secrets.TOKEN_SECRET }}` now. When we are ready to merge to `filecoin-project/FIPs`, I will create a new bot with a new token that we can add to the FIPs repo.

anorth:
I'm not familiar with the original request so will leave primary review to others, but I'm curious about the policy here. What are the rules being executed? What parameters have values? How can we see and change them?

Could you add some documentation and/or configuration?

ianconsolata:
This is just an automation of the manual labeling process @kaitlin-beegle and @luckyparadise are already following: https://github.com/filecoin-project/FIPs/discussions/292. I can link to that discussion. Is there something more you're looking for regarding documentation?

We could factor the dates out into constants too in addition to the constants we defined for the tags, but I'm not really sure what you mean by configuration. This is meant to be an MVP first step, as we start to work on figuring out how to automate some of the other tags as well.

anorth:
Thanks, even just the reference to #292 is helpful to someone encountering this code.

ianconsolata:
Ok, done!

ianconsolata:
@anorth just wanted to check in again on the status of this. We created a new user, @filecoin-governance, that we can use for the token required for this to work. I can securely share a scoped token for that account if you can give that user the appropriate access to make changes, or I can just log in as that user and make the required settings changes myself.

anorth:
Sorry, I'm not sure what's generally wanted of me here. I think this is a @kaitlin-beegle thing, and she has not briefed me on it. I have skimmed the code and it seems fine, I'm happy to rubber-stamp it. But for actually operating the thing, I think you need @kaitlin-beegle directly, and best not to use me as a go-between.

ianconsolata:
@anorth here's what needs to happen to merge:
1. The `filecoin-governance` bot account needs to be assigned `Triage` permissions so it can update labels on discussions. This permission does not allow it to change any code or checked in documents: https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/repository-roles-for-an-organization. @kaitlin-beegle doesn't want to setup new permissions unilaterally, so it would be nice if we could get another maintainer (such as yourself) to do this / sign off on us doing it.
2. Once that permission has been granted, a scoped token for the `filecoin-governance` bot needs to be set in the settings. I think anyone with `Maintain` permissions can do this, but it might require `Admin`. This is something @kaitlin-beegle can do once step 1 is complete.
3. This PR needs to be merged, at which point it will use the configured token to change the appropriate labels.

kaitlin-beegle (approved):

PR #726, opened 2023-06-15
FIP-0065: Ignore built-in market balances in circulating supply (first draft)

anorth:
## Summary
Declare that future protocol changes may commit non-zero data into sectors without requiring a corresponding deal in the built-in market actor.
Remove the built-in market actor's locked balances from the network circulating supply calculation.

Discussion: https://github.com/filecoin-project/FIPs/discussions/719


FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,177 @@
...
+    a built-in storage market deal or the associated data collateral.
+2. Removal of the built-in market's locked balance from the network circulating supply definition.
+
+## Change Motivation
+A built-in market actor deal is the only mechanism today by which storage providers (SPs) can commit data to sectors.
+This is both quite expensive, and limits their utility and value.
+More efficient and flexible storage functionality could be implemented in smart contracts 
+if they were not constrained by the built-in market actor.
+
+Because the built-in market deal is required intermediary, its minimum deal collateral constitutes a network policy for all data.
```
arajasek, line 38:
```suggestion
Because the built-in market deal is a required intermediary, its minimum deal collateral constitutes a network policy for all data.
```

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,177 @@
...
+```
+Locked = PledgeCollateral
+```
+
+## Design Rationale
+### 1. The built-in market actor is optional
+The built-in market actor constrains the data that storage providers (SPs) can store in sectors, limiting their utility and value.
+It supports only a single, simple deal schema and policies, and does not support a range of desirable market features.
+The built-in market actor is also expensive to use.
+Publishing deals is the largest on-chain cost of QA power onboarding, and currently consumes a large fraction of total chain bandwidth.
```
arajasek, line 85: Some concrete numbers would be good here.

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,177 @@
...
+Much of this cost is unnecessary in the case of simple verified deals with no client payment,
+because the verified registry actor already records information describing the datacap allocation.
+The built-in market plays no necessary role in allocating or accounting for QA power.
+
+As an alternative to declaring the built-in market optional, protocol designers could invest in it long term,
+implementing a wide range of flexible functionality in the network core, and optimising to reduce costs (these goals will often conflict).
+This path would be expensive and slow, being subject to prudent conservatism and network governance processes.
+
+Instead, this proposal prepares to for such innovation to occur in user-programmed smart contracts.
```
arajasek, line 94:
```suggestion
Instead, this proposal prepares for such innovation to occur in user-programmed smart contracts.
```

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,177 @@
...
+and in particular the built-in market will continue to require a minimum provider deal collateral when it is used.
+A data termination penalty could easily be implemented at the miner actor in the future if desired.
+
+For deals that are not FIL+ verified (a tiny minority today), the impact of allowing data commitments without deal collateral is 10x larger,
+relative to block rewards, because the deal collateral calculation does not account for QA power.
+This is a disproportionate cost/disincentive to storing data for clients that have not undergone Filecoin Plus allocation processes,
+which raises the amount that SPs must charge those clients in order to break even on storage services.
+A further motivation for this proposal is to remove this disproportionate cost in relation to block rewards for alternative storage applications.
+
+(Note: we could also adjust the built-in market's calculation to use QA Power in this FIP, if desired)
```
arajasek, line 116: Although this change is reasonable, I'd drop this line. It strikes me as a fairly orthogonal change to be making as part of this FIP.

arajasek (approved):
Looks good to me.

AFAIK, this is the first time we've FIP-ped something that is mostly a declaration of intent for future changes. While I don't oppose such a FIP, it's something we should discuss as part of the deliberation process -- if this FIP passes with little contention, but there's opposition to the future FIP that actually introduces direct commitment of data into sectors (on the grounds that the centrality of the built-in market actor shouldn't be eroded), would we entertain that discussion? Or shut it down saying "sorry, this previous FIP already announced a decision on this"?

Regardless, that doesn't affect landing this PR.

FIPS/fip-xxxx-deal-collateral.md
anorth: Will add the PSD share of total gas (44%).

anorth:
> if this FIP passes with little contention, but there's opposition to the future FIP that actually introduces direct commitment of data into sectors [...] would we entertain that discussion?

Thanks for raising. Yes, I think we would. I don't think there's anything that reasonable people could do now that should prevent other reasonable people in the future from debating a change.

With that in mind, I think I will rework this FIP to focus on the technical change, with clear pointers to the rationale being expectation of direct commitment of data. The content will mostly remain, just re-arranged.

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,161 @@
...
+This proposal, or some alternative, is logical preparation for clients and providers arranging
+Filecoin Plus storage directly with the verified registry actor (at a significant cost reduction),
+and to support for smart contract data applications that can avoid the costs and limitations associated with
+mandatory use of the built-in market actor.
+
+The built-in market actor will continue to require minimum deal collateral, so in the near term all deals will as well.
+Participants and developers will need to explicitly alter their arrangements should they wish to make 
+arrangements that uses a lesser amount.
+
+## Implementation
```
jennijuju, line 157: Does this mean we are excluding it from both regular and vm circulating supply or only the VM one?

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Simple Summary
+Exclude the built-in market actor's locked balances from the network circulating supply calculation,
+in preparation for direct commitment of data into sectors without a corresponding built-in market deal.
+
+## Abstract
+A deal with the built-in storage market actor is currently the only permitted way for a storage provider to commit data into a sector.
+The built-in market requires a minimum amount of provider deal collateral,
+which thus functions as a network mandatory data assurance mechanism.
+The minimum deal collateral amounts are tiny compared to the sector's initial pledge or termination penalty,
```
jennijuju, line 20: can we give specific numbers?

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,161 @@
...
+Work is in progress to permit SPs to commit data directly into sectors without a built-in market deal,
+upon which more efficient and flexible storage functionality could be implemented in smart contracts.
+This proposal is preparation for such changes, after which the built-in market actor would no longer occupy
+a privileged position as a required intermediary.
+
+The built-in market's minimum deal collateral amount currently equals about 0.4% of a sector's initial pledge,
+or 3% of the sector termination penalty (for Filecoin Plus verified deals).
+
+The primary motivation for this proposal is simplicity:
+the built-in market's deal collateral is only a very small part of the penalty charged by the network when a sector is terminated early,
```
jennijuju, line 45: would be lovely if we can get some specific numbers so it is easier to understand the scale of. "very small" and "even smaller". One may consider 10% a small amount and others may think 0.1% counted as a small amount.

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,161 @@
...
+author: Alex North <@anorth>
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/719
+status: Draft
+type: Technical
+category: Core
+created: 2023-06-16
+---
+
+## Simple Summary
+Exclude the built-in market actor's locked balances from the network circulating supply calculation,
```
hannahhoward, line 13: Recommend a slightly different phrasing than include/exclude

The built-in markets locked balances are currently included in the total locked amount, which is then subtracted from the total circulating supply. Another way to plausible way to say this is "the total Filecoin circulating supply EXCLUDES the balances locked in the built in actor".

I realize now the proposal is to exclude them from the calculation, but I read this several times as excluding them from the total circulating supply number itself, which I found very confusing because I know that's what already happens.

Maybe "remove the built-in market actors locked balanced from the pool of locked filecoin on the network" is a better way to say it?

jennijuju:
First editorial pass ✔️  with some nits.

re: @arajasek's good flag: This FIP's motivation and impact depends on many hypothetical future protocol changes, that may or may not be accepted by the Filecoin community. The idea of this proposal is reasonable and concrete enough to be landed as a draft, however, when this FIP can be moved to Last Call/Acceptance is worth discussing. some quick options:
1. treat this FIP as an independent FIP. 
2. treat this FIP as a dependency of future "directly commit data to sector" proposals and bound the FIPs acceptance. (stay Draft until then).
3. treat this FIP as an independent FIP now for acceptance(not finalization) & and allow revision to `Deferred` or `Reverted` when contentions raised in then future. 


@kaitlin-beegle I think this is gap in the current governance process and would love to hear our feedback.

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,161 @@
...
+- TotalProviderLockedCollateral: 317,498 FIL
+- TotalClientStorageFee: 38 FIL
+
+This amounts to 0.21% of the total locked amount (150,981,531 FIL), 
+and 0.067% of the circulating supply (473,321,502 FIL)
+
+Removing them from the calculation will thus have negligible effect on downstream calculations 
+such as the sector initial pledge requirement.
+
+Removing these balances from the calculation of network circulating supply will not actually unlock any tokens,
```
hannahhoward, line 121: Even if there's no underlying change, I assume crypto exchanges report the number the chain reports. I wonder if a sudden bump in the number affects token price?

hannahhoward (commented):
comments left per @jennijuju

jennijuju:
(I’m happy to land this DRAFT when we find a venue to discuss the raised governance process questions)

dirkmc (commented):
Overall this looks great, thanks for putting it together @anorth 👍 

My only suggestions are to change a couple of things to improve the readability and understandability (particularly for non-native speakers):
- I'd suggest massaging the language to be as concise as possible
  - eg minimize repetition of the same point in different sections
- I'd suggest using bullet points instead of paragraphs where possible
- I'd suggest using graphics where possible (better for non-native speakers) eg a pie graph of locked balances vs total locked amount vs circulating supply

FIPS/fip-xxxx-deal-collateral.md
anorth: I only know of one protocol-defined circulating supply, which is the value that is input to calculations of things like sector initial pledge. Are any other definitions of circulating supply relevant to consensus rules?

FIPS/fip-xxxx-deal-collateral.md
anorth: Given in the change motivation, below.

FIPS/fip-xxxx-deal-collateral.md
anorth: The number is given in the previous sentence.

FIPS/fip-xxxx-deal-collateral.md
anorth: Thanks, it's tricky to find something concise (FIP title is supposed to be <80 chars, though I find this restriction silly). I changed it to "ignore the built-in market locked balance..."

FIPS/fip-xxxx-deal-collateral.md
anorth: I have some reason to believe they chose their own definition (see @jennijuju comment about about which definition I'm talking about), and I don't know which if any include tokens in deal collateral or payment escrow as "circulating".

anorth:
Thanks @dirkmc , someone should make you a FIP editor. The template structure makes it hard to avoid repetition, but I will attempt to compress and tabulate a bit.

anorth:
I have addressed the editorial comments. @jennijuju please approve given your ✔️ so I can merge this draft for community consideration . We can discuss any governance issues in the the discussion #719.

jennijuju (approved):

FIPS/fip-xxxx-deal-collateral.md
```diff
@@ -0,0 +1,158 @@
+---
+fip: "<to be assigned>"
```
jennijuju, line 2: 0065

Please add this to the readme table too!

PR #733, opened 2023-06-20
Fix yaml in FIP-0019

Kubuxu:


anorth (approved):
Maybe YAML isn't the right fit here. Can you quote the FIP number too, so it's not interpreted as octal?

Kubuxu:
> Can you quote the FIP number too, so it's not interpreted as octal?

Is there a point in fixing it in just this one? If it being interpreted as octal is an issue then it happens in every document.

The primary issue is that while it is useful metadata it is not checked or really used.
https://github.com/filecoin-project/FIPs/discussions/698

jennijuju (approved):

PR #745, opened 2023-06-30
Add Filecoin Protocol Master Plan FIP

Fatman13:
Details could be found in discussion [here](https://github.com/filecoin-project/FIPs/discussions/725).

### DIFF
--- a/FIPS/fip-0000.md
+++ b/FIPS/fip-0000.md
@@ -47,7 +47,7 @@ What we like to stress again is that, on the left side of the illustration above
 
 The following specification is one potential path that a layered protocol architecture could be achieved...
 
-<img src="https://github.com/ipfs-force-community/discourse/assets/1591330/256a0baa-d738-422e-af56-bf26569f2606" width="640" />
+<img src="https://github.com/filecoin-project/FIPs/assets/1591330/891c6c46-6259-4e05-8587-865e6fbe0579" width="640" />
 
 - **Level all sector power**: CC sectors and DC sectors will all just have raw byte storage power, effectively removing concept of QAP and reducing L1 complexity.
 - **Add `subsidy` built-in actor**: with sectors normalized, the subsidized reward for storage deals is completely decoupled from storage power and will be managed by `subsidy` built-in actor. (revisit of [FIP0033](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0033.md))
--- a/FIPS/fip-0000.md
+++ b/FIPS/fip-0000.md
@@ -8,8 +8,6 @@ type: Technical (Core)
 created: 2023-06-30
 ---
 
-Continuing discussion [here](https://github.com/filecoin-project/FIPs/discussions/712#discussioncomment-6117336) with this new thread...
-
 ## Simple Summary
 
 Recalibrate Filecoin's protocol architecture vision to guide future FIPs to the betterment of the whole network.
--- a/FIPS/fip-0000.md
+++ b/FIPS/fip-0000.md
@@ -1,0 +1,66 @@
+---
+fip: "0000"
+title: Filecoin Protocol Master Plan
+author: Steven Li (@steven004), @Fatman13
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/725
+status: Draft
+type: Technical (Core)
+created: 2023-06-30
+---
+
+Continuing discussion [here](https://github.com/filecoin-project/FIPs/discussions/712#discussioncomment-6117336) with this new thread...
+
+## Simple Summary
+
+Recalibrate Filecoin's protocol architecture vision to guide future FIPs to the betterment of the whole network.
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


anorth:
This is not close to a sufficient level of technical detail for a FIP. I suggest we close this and continue the discussion in #725 first.

jennijuju:
I agree with @anorth 

> A FIP is a design document providing information to the Filecoin community, or describing a new feature for Filecoin or its processes or environment. The FIP should provide a concise technical specification of the feature and a rationale for its adoption.


I would also recommend the authors to update the fip title according to the specific features that’s it’s proposing when reopen the FIP, the current title is too vague and could be misleading imho.

Fatman13:
> This is not close to a sufficient level of technical detail for a FIP. I suggest we close this and continue the discussion in https://github.com/filecoin-project/FIPs/discussions/725 first.

Is technical details required for FIPs? [FIP0003](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0003.md) and [FIP0018](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0018.md) both seems lacking on that regard too.

kaitlin-beegle:
> > This is not close to a sufficient level of technical detail for a FIP. I suggest we close this and continue the discussion in #725 first.
> 
> > I would also recommend the authors to update the fip title according to the specific features that’s it’s proposing when reopen the FIP, the current title is too vague and could be misleading imho.
> 
> Are technical details required for FIPs? [FIP0003](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0003.md) and [FIP0018](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0018.md) both seems lacking on that regard too.

Yes, these are community FIPs.  Unlike these FIPs, the Filecoin Master Plan does include significant technical proposals. 

I believe this proposal is worth continuing to discuss, and potentially using as an edge case to inform pending FIP process improvements.  Right now, however, this is not a focused enough proposal to be opened.  I also think this should be rejected pending additional work.

Fatman13:
> Yes, these are community FIPs. Unlike these FIPs, the Filecoin Master Plan does include significant technical proposals.

> I believe this proposal is worth continuing to discuss, and potentially using as an edge case to inform pending FIP process improvements. Right now, however, this is not a focused enough proposal to be opened. I also think this should be rejected pending additional work.

Acknowledged.

jennijuju:
> > This is not close to a sufficient level of technical detail for a FIP. I suggest we close this and continue the discussion in #725 first.
> 
> > I would also recommend the authors to update the fip title according to the specific features that’s it’s proposing when reopen the FIP, the current title is too vague and could be misleading imho.
> 
> Are technical details required for FIPs? [FIP0003](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0003.md) and [FIP0018](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0018.md) both seems lacking on that regard too.

The standard of the FIPs have been evolved a lot over the past three years, and many authors have been enhancing the quality of their FIPs as they are being treated as the new spec. Instead of referring to the FIPs from early days (right around network launches), I would recommend you to take well-written FIPs like FIP-0045, all the FVM FIPs, and much more as a reference for your FIP.

Also, FIP-0018 is an organizational FIP instead of a core technical FIP, which have a totally different standard.

Fatman13:
> Also, FIP-0018 is an organizational FIP instead of a core technical FIP, which have a totally different standard.

Got it! 

> Organizational FIPs, or Filecoin Organization Proposals (FOPs) allow the Filecoin community to propose, discuss, and achieve consensus on Filecoin governance. This includes procedures, guidelines, decision-making processes, and changes to FIP processes.

I think I will amend the FIP to just have "layered protocol design approach" alone and the rest is just example path not mandated to be implemented.

kaitlin-beegle:
Closing, per the conversation above.  

Though we have fielded 'community' or 'operational' FIPs in the past- specifically, FIP0001 and FIP0003- I do not believe this FIP is specified enough to be considered an open draft.  FIP0001 and FIP0003 were introduced at the outset of network launch and were specified as programmatic entities that could be implemented.  This FIP seems to represent a blanket network policy, and it's unclear to me what 'acceptance' would actually mean.  I think a FIP proposal like this could make more  sense at a future time. 

Also, I will note: the very purpose of the Filecoin network is to serve as a decentralized storage layer.  While a proposal such as this may seem technically efficient, it is my opinion that this is not keeping with the mission or purpose of the Filecoin project. Filecoin's very purpose and technology were designed to service a storage utility.  There are plenty of other projects that provide simple blockchain transaction records, which you could design a data storage service on top of.  In fact, other people have done this very thing.  If that is infrastructure that a team wishes to work with, they are welcome and free to go do that- but, as a FIP Editor, I do not believe that is what Filecoin is supposed to be.

The-Wayvy:
Kaitlin

The expense and slowness of blockchains is justified only by permissionlessness of participation by users and miners, verifiability of all computation, and immutability of past data. If the features uniquely enabled by a blockchain are nerfed then it no longer makes sense to use one. No marketing slogan, mission statement or growth strategy can change the fact that FIL+ retains the weaknesses of blockchains while discarding their strengths. It’s analagous to buying an expensive bicycle only to carry it on your shoulder instead of riding it.

Put simply: Filecoin+ is making Filecoin less useful.

PR #763, opened 2023-07-28
FIP-0071: Deterministic State Access (IPLD Reachability)

Stebalien:
This FIP brings us a step closer to user-defined WebAssembly actors by introducing deterministic rules defining what "state" (IPLD blocks) an actor is and is not allowed to read. Specifically, this FIP specifies that an actor may only read state "reachable" from:

1. The actor's state-tree.
2. Parameters passed into the actor from other actors.
3. Blocks returned to the actor from other actors.

Where "reachable" means that the state can be "reached" by traversing IPLD links (CIDs) from the "roots" listed above.

Discussions: https://github.com/filecoin-project/FIPs/discussions/764
WIP Implementation: https://github.com/filecoin-project/ref-fvm/pull/1824

### DIFF
--- a/FIPS/fip-XXXX.md
+++ b/FIPS/fip-XXXX.md
@@ -64,6 +64,8 @@ Additionally, it exposes 2 syscalls for manipulating the actor's state-tree itse
 
 Finally, IPLD blocks can be sent between actors by as message parameters/return values (via `send::send`). Such IPLD blocks are sent _by-handle_, not by-cid.
 
+NOTE: the function signatures above are for illustrative purposes. The actual syscalls are defined in [FIP0030](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0030.md#namespace-ipld).
+
 #### Supported IPLD Codecs
 
 The FVM currently supports reading (`ipld::block_open`) and writing (`ipld::block_create`) state with the following IPLD codecs (`SUPPORTED_CODECS`):
@@ -239,6 +241,10 @@ This FIP _could_ have proposed a single "reachable" set maintained across an ent
 This FIP forbids "inlining" blocks into state-root CIDs as we'd _otherwise_ need to perform IPLD Link Analysis on the inlined block when calling `ipld::set_root`.
 
 ### Link Analysis
+
+#### Ignoring v. Rejecting Unknown CIDs
+
+The link analysis algorithm defined in this FIP _rejects_ CIDs with unknown codecs and/or hash functions instead of simply ignoring such CIDs for better forward-compatibility. This way, new codecs and hash functions can be supported in the future without worrying that such CIDs may already be present in the state (where they would have previously not been covered by link analysis).
 
 #### Native v. Wasm
 
--- a/FIPS/fip-XXXX.md
+++ b/FIPS/fip-XXXX.md
@@ -20,7 +20,12 @@ Where "reachable" means that the state can be "reached" by traversing IPLD links
 
 ## Abstract
 
-TODO
+This FIP introduces additional logic into the FVM's state management system to ensure that:
+
+1. An actor may only read its own state.
+2. When an actor writes new state, that new state may only reference an actors existing state.
+
+This ensures that an actor cannot read and/or reference random IPLD data in the client's blockstore which could lead to network forks (once users are allowed to deploy arbitrary native actors to the network).
 
 ## Change Motivation
 
--- a/FIPS/fip-XXXX.md
+++ b/FIPS/fip-XXXX.md
@@ -251,7 +251,7 @@ The link analysis algorithm defined in this FIP _rejects_ CIDs with unknown code
 We previously considered performing link analysis in a Wasm module to:
 
 1. Take advantage of our existing Wasm gas accounting rather than manually charging for gas based on the CBOR's structure.
-2. Ensure that all implementations used the exact same wasm parsing and validation logic.
+2. Ensure that all implementations used the exact same Wasm parsing and validation logic.
 
 However, this would have increased implementation complexity and introduced additional runtime costs (switching in and out of Wasm isn't free).
 
@@ -287,7 +287,7 @@ TODO
 ## Security Considerations
 <!--All FIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. FIP submissions missing the "Security Considerations" section will be rejected. A FIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
 
-This FIP primarily aims to increase the security of the FVM and guard against malicious and/or buggy wasm actors. However, as with any code:
+This FIP primarily aims to increase the security of the FVM and guard against malicious and/or buggy Wasm actors. However, as with any code:
 
 1. This FIP increases the complexity of the protocol, potentially introducing bugs.
 2. This FIP proposes algorithms to parse potentially user-specified data. If the parsing algorithms and gas costs are not carefully designed/implemented, this could introduce an attack vector.
--- a/FIPS/fip-XXXX.md
+++ b/FIPS/fip-XXXX.md
@@ -262,7 +262,7 @@ We made this choice for a few reasons:
 
 To highlight this point:
 
-1. The DagCBOR spec has been tweaked _many_ times and I have little confidence that it won't change again in the future.
+1. The DagCBOR spec has been tweaked _many_ times and may change again in the future.
 2. The validation rules are [complicated](https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#strictness).
 3. Most DagCBOR implementations _don't_ (or haven't until recently) correctly implemented this spec.
 
--- a/FIPS/fip-XXXX.md
+++ b/FIPS/fip-XXXX.md
@@ -260,6 +260,14 @@ We made this choice for a few reasons:
 1. Performance. Fully validating CBOR would require validating UTF-8 (walking all strings byte by byte), map key order, etc.
 2. Security. The validation logic would become a part of the spec. Any deviation from said validation logic in any client implementation would cause forks. Any bug would either reject valid blocks or accept invalid blocks (betraying the user's expectation that all blocks in the state-tree are fully validated CBOR).
 
+To highlight this point:
+
+1. The DagCBOR spec has been tweaked _many_ times and I have little confidence that it won't change again in the future.
+2. The validation rules are [complicated](https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#strictness).
+3. Most DagCBOR implementations _don't_ (or haven't until recently) correctly implemented this spec.
+
+So, instead, we went for a small but easily verifiable implementation that can be concisely specified with no strange edge-cases and/or exceptions.
+
 ## Backwards Compatibility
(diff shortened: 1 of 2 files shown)


FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,279 @@
...
+    2. If the additional information is in the range 0-23 inclusive, we treat it as the immediate value.
+    3. If the additional information is 24, 25, 26, 27; we read an additional 1, 2, 4, 8 bytes respectively; decode said bytes as a big-endian integer; and treat that as the immediate value. We do not validate that such integers are "minimally encoded" (e.g., 0x1 could be encoded in 8 bytes).
+    4. If the additional information is greater than 27, we treat the CBOR as malformed. Importantly, this means that the FVM will not support [indefinite length][cbor-indefinite] strings, maps, arrays, etc.
+4. If the major type is 0 (integer), 1 (negative integer), or 7 (special), we continue.
+5. If the major type is 2 (byte string) or 3 (string), we seek forward "immediate value" bytes and continue. We do not validate the string's encoding.
+6. If the major type is 4 (array) we add the immediate value to the expected number of fields, and continue.
+7. If the major type is 5 (map) we add two times the immediate value to the expected number of fields, and continue. We do not or restrict the allowed key/value types, nor do we validate the order of keys, nor do we check for duplicates.
```
ZenGround0, line 123: IIRC IPLD CBOR has a hard constraint that all map keys are strings.  Is the idea that any block already linked will be  valid IPLD CBOR so there's no special care needed to ensure that?

ZenGround0 (commented):
Could you explain a bit more of the details and motivation behind the `handle` concept?  What does this level of indirection help us with as opposed to just using cids directly as handles and merging `block_create` with `block_link` and `block_open` with `block_read`? 

Are handles uniquely determined by cids or the same cid could have multiple handles within the same actor? Are handles for the same cid different in different actors?  Do handles contain information about which actor's reachable set they originate from?  What do handles look like in memory/serializaed as method parameters in send:send?

FIPS/fip-XXXX.md
Stebalien: > IIRC IPLD CBOR has a hard constraint that all map keys are strings. Is the idea that any block already linked will be valid IPLD CBOR so there's no special care needed to ensure that?

No, the idea is that the FVM doesn't really care one way or the other. Applications may care, but then its their job to check. The FVM checks the property it cares about.

---

(small rant)

I'd like to eventually move towards a world were we split the IPLD model into:

- A "block" model. I.e., blocks, CIDs, and blocks that link to other blocks. This is the level of bitswap, blockstores, garbage collection, etc.
- A "path" model that simply defines how to path through blocks. You need this for basic pathing, filesystems, etc.
- A "data" model. This is the part that starts caring about map key types, ordering of map keys, etc. You need this if you want to write general-purpose algorithms over abstract data (e.g., graphsync).

One of the biggest challenges in IPLD has been agreeing on how formats map to the data model, especially as not all formats will nicely "fit" into the data model. My goal here is to say that fitting into the data model simply isn't necessary.

Of course, maybe we should use a codec that's not "dag cbor", but I don't think that's really necessary.

Stebalien: I'll add some of this to the design tradeoffs.

Stebalien:
> Could you explain a bit more of the details and motivation behind the handle concept? What does this level of indirection help us with as opposed to just using cids directly as handles and merging block_create with block_link and block_open with block_read?

I've added a link to https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0030.md#ipld-memory-model to the background section. Those syscalls aren't being introduced in this FIP (although I'm happy to discuss them here).

Basically, it comes down to the fact that:

1. We might want to "create" blocks without hashing them (hence returning a handle from `create` instead of returning a CID). This lets us send parameters/return values as IPLD blocks without hashing.
2. We won't know the size of an IPLD block when we first try to `open` it so we don't have a great way to pre-allocate memory for it. The current open/read API mimics the unix file API.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,279 @@
...
+### Link Analysis
+
+#### Native v. Wasm
+
+We previously considered performing link analysis in a Wasm module to:
+
+1. Take advantage of our existing Wasm gas accounting rather than manually charging for gas based on the CBOR's structure.
+2. Ensure that all implementations used the exact same wasm parsing and validation logic.
+
+However, this would have increased implementation complexity and introduced additional runtime costs (switching in and out of Wasm isn't free).
```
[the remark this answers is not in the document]
Stebalien: Unfortunately, we do need to copy and jump every time because we can't trust the user's wasm instance. Basically:

1. Start a new wasm instance for processing the block. We could try to reuse instances here, but that may lead to "weird things". E.g., different costs depending on whether or not memory is already pre-allocated.
2. Jump into the wasm instance to allocate some memory for the block.
3. Jump back into the system to copy the block into the wasm instance's memory.
4. Jump back into the wasm instance to parse the block.
5. Finally, return to the system.
6. And then parse the CIDs emitted by the wasm instance.

But the main part really is complexity:

1. Doing this in wasm saves us from having to define 2 gas charges.
2. Doing this natively saves us from having to manage an additional wasm module, spin up an instance, define the syscalls for crossing back and forth, etc.

Basically, the current approach is both simpler _and_ faster.

anorth (commented):
I'll be happy to give this an editorial 👍 when the abstract is filled in.

I note that test cases will also need to be specified before it can move to last call.

Stebalien:
Ok, this is now ABG&T (all but gas and tests).

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,307 @@
...
+We chose _not_ to perform extensive CBOR validation when creating new DagCBOR blocks and instead do the minimum validation required to extract links. We only guarantee that all written DagCBOR blocks are correctly structured: have valid CBOR major types and can be traversed, field by field.
+
+We made this choice for a few reasons:
+
+1. Performance. Fully validating CBOR would require validating UTF-8 (walking all strings byte by byte), map key order, etc.
+2. Security. The validation logic would become a part of the spec. Any deviation from said validation logic in any client implementation would cause forks. Any bug would either reject valid blocks or accept invalid blocks (betraying the user's expectation that all blocks in the state-tree are fully validated CBOR).
+
+To highlight this point:
+
+1. The DagCBOR spec has been tweaked _many_ times and I have little confidence that it won't change again in the future.
```
anorth, line 265:
```suggestion
1. The DagCBOR spec has been tweaked _many_ times and may change again in the future.
```

anorth (approved):

fridrik01 (commented):
oh had some comments pending for 2 weeks without noticing, sharing now

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,281 @@
...
+This section describes the FVM as it exists today. For more context and history, you may want to read about the [IPLD memory model](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0030.md#ipld-memory-model).
+
+#### IPLD API
+
+The FVM's exposes 4 syscalls for reading/writing IPLD state to actors:
+
+1. `ipld::block_create(codec, data) -> handle` registers a new IPLD block with the FVM, returning a block "handle" (similar to a file descriptor).
+2. `ipld::block_open(cid) -> handle` "opens" the IPLD block referenced by the passed CID, returning a handle to the block.
+3. `ipld::block_read(handle, offset, length) -> data` reads data from an open block into the actor's memory.
+4. `ipld::block_link(handle, multihash_code, multihash_length)` creates a CID for the block referenced by the given block "handle", using the specified multihash code and length (currently limited to blake2b-256).
```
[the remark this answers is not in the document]
Stebalien: _Technically_, but I'm trying to convey the concept here more than the literal API (that's specified in the other FIP linked at the top of the spec). I'll document that.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,281 @@
...
+func ListReachable(codec u64, block []byte) ([]Cid, error)
+```
+
+This function will parse the provided block according to the specified codec (see below) and:
+
+1. If it encounters an "allowed CID":
+    1. If the CID uses the identity hash function (inlines a block), this function will not return the CID directly but will instead recursively parse the inlined block.
+    2. Otherwise, this function will emit the CID.
+2. Any "ignored CIDs" will be skiped and ignroed.
+3. Any other CIDs will cause this function to signal an error.
```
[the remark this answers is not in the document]
Stebalien: https://github.com/filecoin-project/FIPs/pull/763/files#diff-2554e48e24a96725aee3a2de915ff76596325561147c0dc73ac16225e5161c5eR245-R247

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
+
+TODO
+
+## Security Considerations
+<!--All FIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. FIP submissions missing the "Security Considerations" section will be rejected. A FIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
+
+This FIP primarily aims to increase the security of the FVM and guard against malicious and/or buggy wasm actors. However, as with any code:
```
kaitlin-beegle, line 290: nit, but I think the convention is capital "W".  So, Wasm.

kaitlin-beegle (approved):
Approved editorial review.  

Also, great work! This is a really well written FIP.  No real editorial changes, and very simply stated.

FIPS/fip-XXXX.md
Stebalien: Fixed.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,313 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
```
jennijuju, line 2: FIP0071

jennijuju:
(havent fully reviewed yet - tho @Stebalien @fridrik01 I have assigned FIP-0071 to this FIP and feel free to merge with updated FIP number given you have gotten 2 editorial ✅ already, further updates can be made in different PRs

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,313 @@
...
+[block-analysis]: #ipld-block-link-analysis
+
+IPLD blocks may link to (reference by CID) IPLD blocks with a [`SUPPORTED_CODECS`](#supported-ipld-codecs) using either either of the following multihash constructions:
+
+1. A 32-byte Blake2b-256 hash (code 0xb220). E.g., a cid of the form `cidv1-CODEC-blake2b256-32-DIGEST`.
+2. A "identity" hash (code 0x0) with up to a 64 byte digest, "inlining" the referenced block into the CID itself.
+
+We will call these "allowed CIDs".
+
+Additionally, FVM block link analysis will ignore CIDs of sealed and unsealed commitments, as long as the multihash _digest_ is less than 64 bytes:
```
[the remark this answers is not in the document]
Stebalien: It's because:
1.  We store such CIDs in the state already.
2. Sector state isn't "reachable" from actor code regardless.

Basically, it lets actors refer to sectors and pieces directly (by CID) without messing with the FVM complaining about such CIDs not being reachable. I'll update the text.

Stebalien:
Review addressed and FIP# assigned.

jennijuju:
Let’s also add it to the readme table - please merge once that’s in

PR #770, opened 2023-07-30
PR template with some guidance

anorth:
This PR proposed a PR template, which GitHub will auto-fill in the description field of any new PRs ([docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates)). It contains some guidance that I've frequently found myself giving FIP authors after they have already submitted their PR.

@filecoin-project/fips-editors I'd welcome thoughts about anything else that we should say here.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,39 @@
+## Are you drafting a new FIP?
+Please follow these guidelines to help ensure that your proposal is set up for success in the governance process,
+and ease the burden on the FIP editors.
+
+### Start a discussion first
+Changes to the Filecoin core protocol start with open discussion.
+Please open a new discussion topic in the FIPs repository discussion board and outline you motivation and proposal,
+then come back in a few days or weeks to submit your draft FIP.
+
+Don't just paste your FIP draft into a discussion topic.
+The FIP format is suitable for specifying a concrete protocol change for implementation, 
+but your discussion post should describe the background and problem to be solved,
+outline a range of approaches and their tradeoffs, and invite participants to explore other solutions.
+
+Historically, failing to gain some community buy-in before submitting a FIP has often resulted in 
+rejected FIPs and much wasted effort.
+
+### Please use the template
+Please draft your FIP by starting with an appropriate template file from templates directory.
+Do not significantly change the header table format or section headings.
+
+### Don't claim a FIP number
+FIP numbers are allocated by the FIP editors when a draft is ready to be merged into the master branch.
+The editors coordinate to ensure allocations to in-flight drafts don't collide.
+Please do not allocate yourself a FIP number.
+
+Name your file `fip-my_draft_title.md` (replacing the suffix with an abbreviated title)
+and leave "to be assigned" for the FIP number in the header table.
+A FIP editor will either inform you of the appropriate number, or insert it themselves when merging.
+
+### Please add your FIP to the README
+The top-level README.md file contains a table of all FIPs.
+Please add yours there in the same PR that adds the FIP draft itself.
+Use XXXX for the FIP number until it is assigned by an editor.
+
+### Delete this guidance and describe your FIP
+If you've followed the guidance above and are ready to submit your FIP draft,
+please delete this text and replace it with a copy of the _Abstract_ from your draft, 
+and a link to the relevant discussion forum topic.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,15 +1,12 @@
 ## Are you drafting a new FIP?
-Please follow these guidelines to help ensure that your proposal is set up for success in the governance process,
-and ease the burden on the FIP editors.
+Please follow these guidelines to help ensure that your proposal is set up for success in the governance process.
 
-### Start a discussion first
-Changes to the Filecoin core protocol start with open discussion.
-Please open a new discussion topic in the FIPs repository discussion board and outline you motivation and proposal,
-then come back in a few days or weeks to submit your draft FIP.
+### Open a discussion topic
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,39 @@
+## Are you drafting a new FIP?
+Please follow these guidelines to help ensure that your proposal is set up for success in the governance process,
+and ease the burden on the FIP editors.
+
+### Start a discussion first
+Changes to the Filecoin core protocol start with open discussion.
+Please open a new discussion topic in the FIPs repository discussion board and outline you motivation and proposal,
+then come back in a few days or weeks to submit your draft FIP.
+
+Don't just paste your FIP draft into a discussion topic.
```
raulk, line 10: My two cents here.

- Having discussion within a FIP PR and within a discussion thread tends to splinter conversation around a certain topic IMO.
- I'd rather throwaway any process related to _discussions_. Whether it's obliging certain discussions to exist, or managing discussions in a certain manner. The more process exists, the more we stifle fluidity, spontaneity, creativity, and fun. Discussions should be a free flowing space, with a low bar, and templateless.
- It's up to the author to decide how much they want to invest in an idea before they submit it to the community.
	- I expect authors to display enough restraint to not overdevelop an idea and get married to an outcome before presenting it to the community for the first time.
	- I expect authors to display enough emotional regulation if the community shows aversion against their idea, no matter how much time they've invested in it.
- Perhaps we should code author expectations, retain their agency, trust they will do the right thing, and hold them accountable to that standard.
- Some early ideas are truly better submitted as draft FIPs, especifically normative/technical ones (e.g. FVM ones, gas changes, syscall changes, event schema changes). Some authors prefer to submit early ideas structrured as per the standard template (FIP, in this case), so as to minimise future standard development.
- I would personally like to undo the policy that a discussion thread _must_ exist for every FIP.
	- If the author deems prior discussion necessary, they're welcome to create it.
	- If the author is more action-oriented and wants to submit a minimal FIP first, they're welcome to do so, as long as there's sufficient evidence of initial analysis, ponderation, conceptualisation, impact analysis, etc. The FIP can continue maturing as discussion progresses.
		- I'd instate the convention that "general" discussion happens on a comment thread on line 1 (title), OR as review comments (folks can comment on a PR without inlining comments).
	- If a discussion transitions into a FIP, I'd expect the FIP to reference the original discussion, and encode much of the content in the FIP's body.

Note that I personally bias towards encoding including as much history, alternative views, counterarguments, etc. arising in a discussion in the FIP's text to assure survival and avoid platform dependence. GitHub Discussions are a proprietary non-portable platform, whereas Markdown texts are portable and open.

kaitlin-beegle: Hi! Good points, overall.  I don't particularly disagree. 

The original reason for requiring a discussion post (and, as is the case for most hard-and-fast rules we try and maintain) is because of the nature of our soft consensus acceptance protocol.  Without a dedicated space for community discussion, it is difficult to assess whether or not a FIP has actually completed its 'Last Call' period.  It is also a single space we can use to track community conversations around really contentious FIPs; when soft consensus is the norm, we need a single point-of-reference for capturing feedback. 

However, as we try and move away from this idea of ONLY accepting FIPs via a soft consensus community protocol, I'm very open to the idea that discussion posts should not be required. 

My only real concern is that by not requiring a discussion post, we'll see a drastic fall off in open, accessible community discussions around FIPs-in-progress.  Yes, process can be stifling, but this isn't a particularly hard line; in the case where authors have opened quality PRs without discussion posts, we simply ask them to open one. Imo, we could certainly try removing the requirement for discussion posts (after we, hopefully, overhaul certain parts of FIP0001), but I personally would rather police an extra rule than see our current conversation threads completely exit a community forum.

.github/pull_request_template.md
```diff
@@ -0,0 +1,39 @@
+## Are you drafting a new FIP?
+Please follow these guidelines to help ensure that your proposal is set up for success in the governance process,
+and ease the burden on the FIP editors.
```
kaitlin-beegle, line 3: Not sure "...and ease the burden on FIP editors" is necessary. Ensuring that your proposal is successful should be a sufficient reason.

kaitlin-beegle (approved):

kaitlin-beegle:
A small change I'd like to propose is adding a _Consensus Decision_ to the FIPs template.  This was first suggested by @jsoares (see: [HERE](https://github.com/filecoin-project/FIPs/pull/850#discussion_r1363576638)) and I think it's a good idea. 

If there's positive sentiment about doing this, I'll make the change on the file. 

Also, fwiw, the open FIP0001v2 proposal suggests new FIP categories- at this time, I don't think there's a need for FIPs of different times to have different templates.  This could change, but I think one standard template is currently appropriate for FIPs of different types.

.github/pull_request_template.md
anorth: I've removed the text here about any delay, or even sequencing, between discussion and FIP draft.

jsoares (approved):

jennijuju:
>A small change I'd like to propose is adding a Consensus Decision to the FIPs template.
>Done

I prefer to add the consensus decision session AFTER FIP0001v2 is approved. For the time being it serves the same role has the fip status (either passed or not passed by soft consensus)

jennijuju (approved):

anorth:
> I prefer to add the consensus decision session AFTER FIP0001v2 is approved. For the time being it serves the same role has the fip status (either passed or not passed by soft consensus)

Fair, I've removed it so I can land the obvious parts.

PR #772, opened 2023-07-31
Set FIP-0047 status to Superseded

anorth:
The letter of FIP-0001 says that the replacement FIP should be `Final` before this status can be set, but it's misleading for FIP-0047 to be continually advertised as Accepted and ready for implementation. We're never going to implement it. I would be just as happy for this to become `Rejected` or `WeChangedOurMind`.

We can defer merging this status change if desired.


jennijuju:
`Superseded` sgtm - @kaitlin-beegle I suggest we add this status to FIP-0001

jennijuju (approved):

kaitlin-beegle (approved):

PR #784, opened 2023-08-10
[WIP] Improved event syscall API

fridrik01:
TODO


FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Simple Summary
+
+Revise the internal [`emit_event`](https://docs.rs/fvm_sdk/latest/fvm_sdk/sys/event/fn.emit_event.html) syscall by replacing the single CBOR encoded `ActorEvent` parameter with three separate fixed-sized byte arrays. This change enhances gas charge accuracy and eliminates the need for parsing.
+
+## Abstract
+
+The current procedure for emitting an event through the FVM SDK involves sending an ActorEvent structure as input. This structure is encoded into a byte array with CBOR formatting and then sent to the kernel.
+
+Inside the kernel, complete CBOR deserialization is necessary before any validation can occur, which results in an upfront gas charge. However, accurately estimating the gas charge is challenging, as the original size of the `ActorEvent` remains unknown until the CBOR byte array is fully deserialized. Consequently, to ensure sufficient gas, a slightly higher gas fee is imposed, making this process more costly than necessary.
```
Stebalien, line 19: More importantly (also for the motivation below), the current gas charge was estimated based on the cost of emitting EVM events and won't be accurate for arbitrary events emitted from, e.g., the market actor.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+    event_off: *const EventEntry,
+    event_len: u32,
+    key_off: *const u8,
+    key_len: u32,
+    value_off: *const u8,
+    value_len: u32,
+) -> Result<()>;
+```
+
+We also introduce a new public struct in `fvm_shared::sys` where we will store the fixed fields per `Entry` (24 bytes for each entry):
```
Stebalien, line 66: nit: we don't really need to talk about the details of where this lives.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+```rust
+// in fvm_shared::sys
+pub struct EventEntry {
+    pub flags: crate::event::Flags,
```
Stebalien, line 71: nit: I'd just specify this as a `u64`, or leave a comment to that effect.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+    key_len: u32,
+    value_off: *const u8,
+    value_len: u32,
+) -> Result<()>;
+```
+
+We also introduce a new public struct in `fvm_shared::sys` where we will store the fixed fields per `Entry` (24 bytes for each entry):
+
+```rust
+// in fvm_shared::sys
```
Stebalien, line 69: We should specify that this struct is encoded as a "packed tuple in field order", or something like that. I'd consider describing it as `(u64, u64, u32, u32)`.

The rust struct definition is helpful (and should be included) but it doesn't convey the layout.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+pub struct EventEntry {
+    pub flags: crate::event::Flags,
+    pub codec: u64,
+    pub key_len: u32,
+    pub val_len: u32,
+}
+```
+
+In the new `emit_event` syscall instead of taking the whole `ActorEvent` data encoded using CBOR this proposed version splits the `ActorEvent` instance into three different buffers that each refer to a pointer and length in WASM memory where their data has been written and are defined as follows:
+- `event_off/event_len`: Pointer to a buffer of all `EventEntry` for all entries.
```
Stebalien, line 79: note that the event length refers to the number of event entries, not the length of the buffer in bytes. I'd consider calling this an "array" instead of a "buffer" as well.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+}
+```
+
+In the new `emit_event` syscall instead of taking the whole `ActorEvent` data encoded using CBOR this proposed version splits the `ActorEvent` instance into three different buffers that each refer to a pointer and length in WASM memory where their data has been written and are defined as follows:
+- `event_off/event_len`: Pointer to a buffer of all `EventEntry` for all entries. 
+- `key_off/key_len`: Pointer to a buffer of all the concatinated keys for all entries.
+- `value_off/value_len`: Pointer to a buffer of all the concatenated values for all entries.
+
+### Serialization and validation
+The serialization from the syscall to FVM is implemented as follows:
```
Stebalien, line 84: IMO, this level of detail probably isn't necessary. Simply describing the procedure as "concatenating" the keys/values is likely sufficient (and we don't really need to talk about casting as that's language specific).

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+3. For each entry `e`:
+    1. Create an `EventEntry` that includes flags, codec, and the size of the key/value for `e`, then add it to buf.
+    2. Update the counters `total_key_len` and `total_val_len`.
+4. Create a buffer called `keys` with a size of `total_key_len` bytes, and another buffer named `values` with a size of `total_val_len` bytes.
+5. For each entry `e`:
+    1. Add `e.keys` to the `keys` buffer.
+    2. Add `e.values` to the `values` buffer.
+6. Cast the raw memory pointed to by `entries` into a `EventEntry` slice of size equal to the number of entries making sure the pointer is aligned and within bounds.
+
+In the FVM we reconstruct the `ActorEvent` as follows:
```
Stebalien, line 97: In terms of decoding, I _would_ talk about validation (restrictions on the input data) as you do below. But I wouldn't be prescriptive on how to actually decode as long as it's clear what the structure is.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+1. Initial validation:
+    1. Confirm that the total number of entries is fewer than 256.
```
Stebalien, line 100: Let's call this out as a change. Before, we allowed 256 entries total, now we allow 255 to ensure that the length in the resulting cbor-encoded event object fits in a single byte (for gas accounting).

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+    2. Ensure the total length of combined entry data does not exceed 8KiB.
+2. Initialize `key_offset` and `val_offset` counters.
+3. Create a vector `v` of `Entry` with capacity matching the size of the `EventEntry` array.
+4. For each event entry 'ee':
+    1. Validate that the `ee.flag` bits are valid,
+    2. Validate that the `ee.key_len` is at most 32 bytes and is a valid UTF-8.
```
Stebalien, line 106: We changed the key length to be strictly less than 31 bytes for the same reason as above.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+    3. Validate that the `ee.codec` is acceptable. Currently, only `IPLD_RAW` (0x55) is allowed).
+    4. Create a new `key` string by reading from the `keys` buffer, starting from `key_offset` and spanning `ee.key_len` bytes.
+    5. Create a new `value` byte array by reading from the `values` buffer, starting from `val_offset` and spanning `ee.val_len` bytes.
```
Stebalien, line 109: Given that we're repeating the validation logic, I'd include the restrictions on the total max values length (8KiB). This was included in the initial spec, but I wouldn't count on someone reading that.

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+
+In the FVM we reconstruct the `ActorEvent` as follows:
+
+1. Initial validation:
+    1. Confirm that the total number of entries is fewer than 256.
+    2. Ensure the total length of combined entry data does not exceed 8KiB.
+2. Initialize `key_offset` and `val_offset` counters.
+3. Create a vector `v` of `Entry` with capacity matching the size of the `EventEntry` array.
+4. For each event entry 'ee':
+    1. Validate that the `ee.flag` bits are valid,
```
Stebalien, line 105: I'd reference the original FIP (flags section of FIP 0049).

FIPS/fip-new-syscall-api.md
```diff
@@ -0,0 +1,141 @@
...
+    2. Validate that the `ee.key_len` is at most 32 bytes and is a valid UTF-8.
+    3. Validate that the `ee.codec` is acceptable. Currently, only `IPLD_RAW` (0x55) is allowed).
+    4. Create a new `key` string by reading from the `keys` buffer, starting from `key_offset` and spanning `ee.key_len` bytes.
+    5. Create a new `value` byte array by reading from the `values` buffer, starting from `val_offset` and spanning `ee.val_len` bytes.
+    6. Create new a `Entry` using `ee.flag`/`ee.codec`/`key`/`value` and append that to `v`.
+    7. Update `key_offset` and `val_offset`.
+5. Return vector `v`.
+
+### Gas
```
Stebalien, line 114: We should describe this in more detail but with "TODOs" for the numbers. I.e.:

1. Define what "up front" means (before validating, but after asserting that the actor is not in "read-only" mode).
2. Describe the precise logic of this function. Don't try to describe how it changed, just describe how it is now.

You may need to refer to gas charges defined in previous FIPs (memcpy, etc.). If you can't quickly find a reference, just leave a TODO and I'll fill it in.

(I'm also happy to help here given that I wrote this part of the code)

FIPS/fip-new-syscall-api.md
fridrik01: That's already in 1.2 above but I'll change the text so its more clearer as it did not say values explicitly.

FIPS/fip-0068.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Abstract
+
+The current procedure for emitting an event through the FVM SDK involves sending an `ActorEvent` structure as input. This structure is encoded into a byte array with CBOR formatting and then sent to the kernel.
+
+Inside the kernel, complete CBOR deserialization is necessary before any validation can occur, which results in an upfront gas charge. However, accurately estimating the gas charge is challenging, as the original size of the `ActorEvent` remains unknown until the CBOR byte array is fully deserialized. Consequently, to ensure sufficient gas, a slightly higher gas fee is imposed, making this process more costly than necessary.
+
+Moreover, the current gas charge is estimated based on the cost of emitting EVM events and won't be accurate for general events emitted by Wasm actors. Thus, we propose a new approach to calculating the gas cost, which should account for any type of emitted events.
```
Stebalien, line 21: uber nit: I'd move this to the motivation

FIPS/fip-0068.md
```diff
@@ -0,0 +1,186 @@
...
+
+
+The proposed processing fee is:
+
+```
+// cost for processing each entry
+p_gas_per_entry_flat := 1750
+p_gas_per_entry_per_byte := 25
+
+// cost for UTF-8 decoding per byte, set to zero for now
```
Stebalien, line 127: I'd call this out as an explicit TODO that must be resolved before the FIP can reach final call.

Stebalien (commented):
LGTM! I can't approve as a FIP editor, but this looks correct.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+```
+### Proposed change
+We propose changing the `emit_event` syscall shown above to the following:
+
+```rust
+pub fn emit_event(
+    event_off: *const EventEntry,
+    event_len: u32,
+    key_off: *const u8,
+    key_len: u32,
```
raulk, line 62: (peer review)

1. Could we disambiguate the `key_len` here with `key_len` below? Same for `val_len`? Perhaps call these `*_size` or the ones below `*_cnt`?
2. Mind elaborating why the counts belong inside `EventEntry`, instead of as arguments in the syscall?

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+
+In the new `emit_event` syscall instead of taking the whole `ActorEvent` data encoded using CBOR this proposed version splits the `ActorEvent` instance into three different buffers that each refer to a pointer and length in WASM memory where their data has been written and are defined as follows:
+- `event_off/event_len`: Pointer to an array of `EventEntry` where `event_len` is the number of elements in the array.
+- `key_off/key_len`: Pointer to a buffer of size `key_len` bytes of all the keys from each entry are concatenated.
+- `value_off/value_len`: Pointer to a buffer of size `value_len` bytes of all the values from each entry are concatenated.
+
+### Serialization
+The serialization from the syscall to FVM is implemented as follows:
+
+Given a `ActorEvent` with _N_ number of `Entry` elements, declare three buffer:
```
raulk, line 88: (Editorial review) Nits.

```suggestion
Given an `ActorEvent` with _N_ number of `Entry` elements, declare three buffers:
```

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+In the FVM we perform the following validation while deserializing the input:
+
+1. Initial validation:
+    1. Validate that the size of `entries` is 255 or less. Previously 256 entries were accepted but reducing it by 1 ensures that the resulting CBOR encoding can store the length in a single byte.
+    2. Validate that the size of `values` does not exceed 8KiB.
+2. For each event entry `ee` in `entries`:
+    1. Validate that the `ee.flags` bits are valid (see [FIP-0049#New chain types](https://github.com/filecoin-project/FIPs/blob/6c2be9a09a7f03f16aa5f635e4beadeaa9c4fb3b/FIPS/fip-0049.md#new-chain-types) for more details).
+    2. Validate that the `ee.key_len` is 31 bytes or less and is a valid UTF-8. Previously 32 byte sized keys were accepted but reducing it by 1 ensures compact CBOR encoding.
+    3. Validate that the `ee.val_len` is 8Kib or less.
+    4. Validate that the `ee.codec` is acceptable. Currently, only `IPLD_RAW` (0x55) is allowed).
```
raulk, line 110: (Editorial review)

```suggestion
    4. Validate that the `ee.codec` is acceptable. Currently, only `IPLD_RAW` (0x55) is allowed.
```

raulk, line 110: (Peer review)

Decreasing the lengths to ensure compact CBOR encoding in all cases should have no impact since the EVM runtime produces events with max. 5 entries and and 5 total keys. However, we should specify that in the backwards compatibility section.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Gas
```
raulk, line 112: (Peer review)

Could we call out how the gas fee is effectively reduced, as per what's stated in the abstract?

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+## Design Rationale
+
+### Serialize a single buffer instead of using three
+We considered serializing the `ActorEvent` using a single buffer (as with CBOR) for simplicity but decided against that since it required _some_ parsing if we wanted to charge gas for keys/values separately.
+
+## Backwards Compatibility
+
+This FIP is consensus breaking, but should have no other impact on backwards compatibility.
```
raulk, line 174: (Peer review)

See comment above. I think this should explicitly reason about why EVM smart contracts are unaffected.

raulk (approved):
LGTM for an editorial pass after editorial nits are commited. Have peer review comments, but those can be answered even after merge.

anorth:
I notice the title of this PR says FIP-00678, but that FIP number is not available.Please wait for editor approval before they will assign a FIP number.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+    2. Validate that the size of `values` does not exceed 8KiB.
+2. For each event entry `ee` in `entries`:
+    1. Validate that the `ee.flags` bits are valid (see [FIP-0049#New chain types](https://github.com/filecoin-project/FIPs/blob/6c2be9a09a7f03f16aa5f635e4beadeaa9c4fb3b/FIPS/fip-0049.md#new-chain-types) for more details).
+    2. Validate that the `ee.key_len` is 31 bytes or less and is a valid UTF-8. Previously 32 byte sized keys were accepted but reducing it by 1 ensures compact CBOR encoding.
+    3. Validate that the `ee.val_len` is 8Kib or less.
+    4. Validate that the `ee.codec` is acceptable. Currently, only `IPLD_RAW` (0x55) is allowed).
+
+### Gas
+
+This FIP makes an adjustment to the gas cost when emitting an event as originally defined in [FIP-0049#Gas costs](https://github.com/filecoin-project/FIPs/blob/6c2be9a09a7f03f16aa5f635e4beadeaa9c4fb3b/FIPS/fip-0049.md#gas-costs). Instead of having a separate gas fee for validation and processing of an event, we instead charge an upfront gas fee. This fee is applied after checking for read-only but before any other validation, allocation, or deserialization is performed as described in the specification above.
```
kaitlin-beegle, line 114: Could you specify this a bit more? The "upfront" gas fee is actually a single fee charged for both validation and processing, rather than two fees charged separately?

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+
+This FIP is consensus breaking, but should have no other impact on backwards compatibility.
+
+## Test Cases
+
+Provided with implementation.
+
+## Security Considerations
+
+TODO
```
kaitlin-beegle, line 182: Are you working to have this audited for security considerations, or are there no security considerations that you're reasonably aware of?

kaitlin-beegle (changes requested):
Editorially review complete.

FIPS/fip-XXXX.md
```diff
@@ -0,0 +1,190 @@
...
+The proposed processing fee is:
+
+```go
+// cost for processing each entry
+p_gas_per_entry_flat := 1750
+p_gas_per_entry_per_byte := 25
+
+// cost for UTF-8 decoding per byte, set to zero for now
+//
+// TODO: This needs to be benchmarked and tuned before this FIP can proceed to Final status
```
arajasek, line 131: It'd be nice if we had a separate "blockers" or "TODOs" section within the FIP template itself. We could put stuff like this there, that doesn't need to be done before we can start discussing the FIP seriously, but must be done before the FIP can actually land.

arajasek (approved):
LGTM, thank you!

FIPS/fip-XXXX.md
BlocksOnAChain: No audits are planned for this at the moment, we will probably go over this in future FVM audits, as we move further in our FVM Wasm timeline.

FIPS/fip-XXXX.md
fridrik01: There is a TODO below in the gas fee calculation but yes I agree we should confirm that the gas fee is reduced if we claim that in the FIP. Will add a new TODO to call that out explicitly

fridrik01:
Updated PR that should should address all review comments. Created a new [FIP PR](https://github.com/filecoin-project/FIPs/pull/794) that proposes a new TODO section which @arajasek suggested and I think would be useful.

BlocksOnAChain (approved):
this proposal looks good to me, after all the small edits that @fridrik01 made.

fridrik01:
@kaitlin-beegle ascould you do another review to confirm that all requested changes have been addressed?

kaitlin-beegle (approved):
Looks great; editorial review done. Thanks!!

fridrik01:
Rebased to address merge conflict

BlocksOnAChain:
@kaitlin-beegle @luckyparadise @jennijuju - Since this Proposal has now passed the reviews, can we get a FIP number and move it to the list for the NV21 upgrade?
CC: @Stebalien @fridrik01 for visibility.

PR #785, opened 2023-08-11
FIP-0063: additional details requested during core dev call

CluEleSsUK:
I've added some extra details to the FIP that were highlighted during the dev call, in particular:
- specification for implementers unfamiliar with drand
- link to rust crate for performing beacon validation

I've left a slot for the chain info for `quicknet` which should be available on 2023-08-23 after the ceremony is completed

### DIFF
--- a/FIPS/fip-0063.md
+++ b/FIPS/fip-0063.md
@@ -17,7 +17,7 @@ The randomness source for Filecoin, the League of Entropy’s drand network, is 
 
 ## Abstract
 
-The current drand network operates in “chained mode”, i.e., each beacon produced is tied to the previous one. Currently, drand beacons are produced every 30 seconds. The protocol upgrades we have recently developed and deployed as the new League of Entropy drand `fastnet` network change these two network characteristics. Randomness beacons are *not chained* to previous beacons and are *produced at a higher frequency of once every 3s* (that remains a divisor of Filecoin block time).
+The current drand network operates in “chained mode”, i.e., each beacon produced is tied to the previous one. Currently, drand beacons are produced every 30 seconds. The protocol upgrades we have recently developed and deployed as the new League of Entropy drand `quicknet` network change these two network characteristics. Randomness beacons are *not chained* to previous beacons and are *produced at a higher frequency of once every 3s* (that remains a divisor of Filecoin block time).
 
 These new features open up a number of new horizons for drand itself, but also for the applications that use drand. Applications operating on lower or higher frequencies can now use drand by picking the subset of rounds according to their needs for randomness input. For instance, Filecoin can pick one every 10 beacons (since the new network is running at a 3s frequency) to satisfy its need for randomness every 30s for its Leader election. This is supported by the fact that randomness is unchained and therefore not all beacons need to be kept by clients in order to verify every new beacon. In turn, this means a significantly lower memory requirement, as: i) client nodes do not need to keep the entirety of the beacon chain since the genesis beacon and, ii) the verification process is stateless and much simpler.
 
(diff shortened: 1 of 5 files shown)


FIPS/fip-0063.md
```diff
@@ -28,23 +28,42 @@ For further details on these new features and their implementation we refer the
...
 For Go implementations, this can be achieved by using the latest drand client library, and is transparent and handled by specifying the right “scheme” upon instantiation. At present there are no other language implementations supporting this, however they can be provided on demand (or existing open source implementations can be easily modified to be compliant).
 
 Additional minor changes will be needed in implementations:
-- the way beacons are retrieved will need to be adapted in order to fetch every 10th beacon to accommodate for Filecoin epochs
+
 - the new public key, chain hash and scheme will need updated to correctly verify beacon signatures
-- the new catch-up time of 2s per beacon will need to be taken into account versus the current 15s catch-up period for the current frequency of 30s. When fetching only every 10th beacon, it means a slightly larger catch-up time of 20s.
+
+The details of `quicknet` chain info are:
+$insertDetailsHereAfterNetworkFoundation
```
jennijuju, line 56:
```suggestion
TODO: $insertDetailsHereAfterNetworkFoundation
```

FIPS/fip-0063.md
```diff
@@ -28,23 +28,42 @@ For further details on these new features and their implementation we refer the
...
+
+- retrieve every 10th beacon
+
+The way beacons are retrieved must be adapted to fetch every 10th beacon instead of every beacon to accommodate for Filecoin epochs. In previous iterations of filecoin, operators could request drand beacons with monotonically increasing round numbers, as the drand and filecoin epochs were aligned. Upon switching to the new scheme, consumers must ensure they request the right drand epoch over either HTTP or gossipsub.
```
jennijuju, line 60: > every 10th beacon 

might be a dumb q: does every 10th beacon mean we are enforcing Filecoin to only consume drand beacons at height thats a multiplier of 10, i.e: height 100, 150, 180, 5000 and so on?

Asking cuz does that mean, network upgrade epoch couldn't be scheudled at the time  say when the quicknet height is at 75,  otherwise  the network will have to wait for 15 sec (5 quicknet height) before proceeding?

FIPS/fip-0063.md
```diff
@@ -28,23 +28,42 @@ For further details on these new features and their implementation we refer the
...
+
+- account for new catch-up time
+
+The new catch-up time of 2s per beacon will result in a longer catch-up period per Filecoin epoch of 20 seconds rather than the current 15 seconds; any monitoring, alerting, reporting or automation relying on the existing catchup time must be updated with the new period of 20s.
```
jennijuju, line 64: (side note - what kind of the monitoring do we have in place today relay on this, do you know?

FIPS/fip-0063.md
```diff
@@ -134,6 +157,8 @@ The following configurations will need to be adapted:
 - [/build/params_mainnet.go](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/params_mainnet.go#L17)
 - [drand configuration](https://github.com/filecoin-project/lotus/blob/e2f4c70b70241cf0ee0af74d32c98bf48a6fb294/build/drand.go#L45) will need to be amended to include the new `public key` and`chain hash` as well as to add the notion of `scheme` that determines how verification and signatures are done
 
+For rust implementations, the [Nois network](https://nois.network) maintain a crate for performing verification of beacons using the new scheme, [drand-verify](https://github.com/noislabs/drand-verify/)
```
jennijuju, line 160: Ooo this is cool 👍  , thanks for linking

jennijuju (approved):

jennijuju:
added editorial stamp for the update.. tho with some open qs!

FIPS/fip-0063.md
CluEleSsUK: Not quite - the network could start with quicknet height 75 but would then have to consume 85 as the next beacon.

FIPS/fip-0063.md
CluEleSsUK: I have no insight into FIL infra's monitoring for this but @glacierwalrus might. On the drand side, we have a bunch of alerting for the new network (pagerduty is outstanding though I think it's more general on the node being down anyhow)

GlacierWalrus: We don't monitor this directly. I'm not even sure we have any second order alerts/metrics which would be significantly affected.

CluEleSsUK: If i understand correctly, drand going down would cause Filecoin to just start using the old block headers in some fashion right? i.e. the network wouldn't ground to a halt (at least not immediately)

anorth (approved):

PR #794, opened 2023-08-18
Add a TODO section to template

fridrik01:
This PR adds a TODO section to the PR template which lists any unresolved issues or tasks that need to be addressed before giving the FIP a "Final" status but are otherwise not required to move the FIP forward in the review process. 

Examples of these include performing benchmarking to know gas fees, validate claims made in the FIP once the final implementation is ready, etc.

This should make it easier to track unresolved issues, reduce chance of missing them, and allows for easier review (especially for larger FIPs). The TODO section should be empty before moving a FIP to Final status.


anorth:
I think the idea in general is sound. However the TODOs must be resolved before a FIP can move to Last Call, not to Final. Participants need all the information to evaluate a proposal.

anorth (approved):

jennijuju (approved):

jennijuju:
Thank you for  the suggestion @fridrik01 !

PR #796, opened 2023-08-21
Create FIP-0077-add-cost-opportunity-for-new-miner-creation.md

remakeZK:
Create add-cost-opportunity-for-new-miner-creation fip

Discussion: https://github.com/filecoin-project/FIPs/discussions/780


FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+author: "Zac <@remakeZK>"
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-21
+---
+
+## Simple Summary
+
+- Add opportunity cost (in FIL) to each node creation to prevent the miner ID creation spam.
```
jennijuju, line 14:
```suggestion
- Add opportunity cost (in FIL) to each miner actor creation to prevent the miner ID creation spam.
```

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Abstract
+
+By imposing a cost, this proposal helps prevent the spamming of creating miner IDs in order to:
+
+1) lighten the network chain storage burden
```
jennijuju, line 20: could you please elaborate more on/quantify the storage burden empty miner actor creates on the network?

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+2) save wasted resources and future increment of gas that caused by wasted storage
+
+3) avoid the malicious attacks achieved by creating a large number of accounts to cause network congestion and unstable gas fees
```
jennijuju, line 24:
```suggestion
3) avoid the malicious attacks achieved by sending large amount of `CreateMiner` messages with low cost that cause network congestion and unstable gas fees
```

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Problem Motivation
+
+By August 2023, according to the statistics of Filscan, there are 601,131 Miner IDs have been built with only 4,862 actives. The actual ID usage counts for merely 0.8% and the abuse of ID creation trend continues. The actual ID usage only counts 3.7% for the recent two months (125 Miner IDs are active out of 3,254 newly created). 
+
+<img width="865" alt="total_in_use" src="https://github.com/remakeZK/FIPs/assets/92775781/fc8266ab-4224-49ab-a4a7-6450377d4638">
+<img width="913" alt="recent_two_months" src="https://github.com/remakeZK/FIPs/assets/92775781/2b41e3a8-ad5f-481a-bbbb-569fc4533e0b">
+
+  
+Many users randomly create numerous nodes just to get preferred IDs, wasting resources and burdening the network for no purpose. Moreover, those wasted nodes occupy the storage of the network; with time elapsed, the waste resource rate will increase, along with increasing the future gas price because of those unnecessary storages of the network. Furthermore, there is a risk of exposure to potential attacks by creating numerous accounts, causing network congestion.
```
jennijuju, line 34: > occupy the storage of the network

would be great to get more data https://github.com/filecoin-project/FIPs/pull/796/files#r1300851319 so to give a clearer picture of "how big the problem is"

jennijuju, line 34: > causing network congestion. 

do you mean "One can cause network mpool congestion by sending numerous `CreateMiner` messages yet with low cost, and leads to unstable network base fee and potentially blocking more useful/critical/time sensitive network messages (i.e: `SubmitWindowedPoSt`, `ReportConsensusFault`, sector/data onboarding messages,  to land onchain."?

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+<img width="865" alt="total_in_use" src="https://github.com/remakeZK/FIPs/assets/92775781/fc8266ab-4224-49ab-a4a7-6450377d4638">
+<img width="913" alt="recent_two_months" src="https://github.com/remakeZK/FIPs/assets/92775781/2b41e3a8-ad5f-481a-bbbb-569fc4533e0b">
+
+  
+Many users randomly create numerous nodes just to get preferred IDs, wasting resources and burdening the network for no purpose. Moreover, those wasted nodes occupy the storage of the network; with time elapsed, the waste resource rate will increase, along with increasing the future gas price because of those unnecessary storages of the network. Furthermore, there is a risk of exposure to potential attacks by creating numerous accounts, causing network congestion.  
+
+
+## Specification
+
+In creating a new node,
```
jennijuju, line 39: -> Upon creating a new Miner actor?


(node is a very overloaded term in general Lolol  for example, could be easily mistaken as client node and so on.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Currently: there is no cost associated with creating new miners, so users spams many new nodes without limit
```
jennijuju, line 41: > there is no cost associated with creating new miners

-> there is no cost associated with creating new miner actors other than the gas cost for the `CreateMiner` message..?

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+- Proposed:   
+Require spending of a one-time 500FIL to create a new miner ID and deposit the 500FIL into miner account as locked rewards.
```
jennijuju, line 43: https://github.com/filecoin-project/FIPs/discussions/780#discussioncomment-6787087

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Design Rationale
+
+Adding an opportunity cost will prevent such spammy behavior; moreover, depositing fees as locked rewards does not incur actual costs but will occupy funds and bring opportunity costs. Users need to weigh the opportunity cost of locking funds for 180 days per new node. Also, the more nodes created, the more funds need to be locked. This can prevent users from abusing node creation. Additionally, this mechanism helps determine whether each new node will be fully utilized since idle miners will waste locked funds in vain. 
+
+## Backward Compatibility
+
+The charging fee of node creation might increase the entry barrier of new SPs, but 500FIL is a small amount in terms of initiating. Moreover, the proposed first solution of converting fees to 180-day locked rewards adds no real capital cost, only opportunity cost. This encourages on-demand account creation, as users will create accounts based on actual needs, ensuring practical use of each new account.
```
jennijuju, line 51: I think `Product Consideration` is a better place for whats written here.


If there is no retrospective charge for existing miner actors, the proposal likely is backward compatible.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Specification
+
+In creating a new node,
+
+- Currently: there is no cost associated with creating new miners, so users spams many new nodes without limit
+- Proposed:   
+Require spending of a one-time 500FIL to create a new miner ID and deposit the 500FIL into miner account as locked rewards.
+
+## Design Rationale
```
jennijuju, line 45: I would like to see more rational behind why  500 FIL was chosen as the opportunity cost in this section - I.e: how this amount should be efficient enough to prevent the spam.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+Adding an opportunity cost will prevent such spammy behavior; moreover, depositing fees as locked rewards does not incur actual costs but will occupy funds and bring opportunity costs. Users need to weigh the opportunity cost of locking funds for 180 days per new node. Also, the more nodes created, the more funds need to be locked. This can prevent users from abusing node creation. Additionally, this mechanism helps determine whether each new node will be fully utilized since idle miners will waste locked funds in vain.
```
jennijuju, line 47: would suggest replace `node` with `miner actor`

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Security Considerations
+
+There is no security consideration for this proposal; instead, it helps to prevent potential attacks by using a large number of newly created IDs, and the scenarios of network congestion and instability of gas fees that are caused by unnecessary numerous ID creations. 
+
+## Incentive Considerations
+
+Charging for creating new accounts can help prevent abuse, reduce unnecessary storage, and avoid increased gas fees over time caused by such wasteful activities that burden the network with unnecessary storage. 
+
+## Product Considerations
```
jennijuju, line 69: https://github.com/filecoin-project/FIPs/discussions/780#discussioncomment-6787111

jennijuju:
editorial pass ✅ with some suggestions

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
remakeZK: the amount of storage burden looks relatively small now, but it’s a long-term case, with the time elapsed, the accumulated cost would be not ignored. The more important is also preventing attacks/spamming to the network with very cheap cost.

remakeZK: > could you please elaborate more on/quantify the storage burden empty miner actor creates on the network?

the amount of storage burden looks relatively small now, but it’s a long-term case, with the time elapsed, the accumulated cost would be not ignored. The more important is also preventing attacks/spamming to the network with very cheap cost.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
remakeZK: > I would like to see more rationale behind why 500 FIL was chosen as the opportunity cost in this section - I.e: why this amount is efficient enough to prevent the spam.

The 500 FIL amount is adopted from the suggestion in the discussion. That will be great to hear if you have any suggestion regarding the amount.

hunjixin: what about use waterline storage deposit as create miner deposit, means 10t*6fil/t =60fil on mainnet

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,79 @@
...
+author: "Zac <@remakeZK>"
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-21
+---
+
+## Simple Summary
+
+- Add opportunity cost (in FIL) to each miner actor creation to prevent the miner ID creation spam.
```
kaitlin-beegle, line 14: Why is "opportunity cost" a better term than simply saying "fee"?

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Abstract
+
+By imposing a cost, this proposal helps prevent the spamming of creating miner IDs in order to:
+
+1) lighten the network chain storage burden
+
+2) save wasted resources and future increment of gas that caused by wasted storage
```
kaitlin-beegle, line 22:
```suggestion
2) Prevent wasted resources and incremental increases in gas cost caused by wasted storage, and;
```

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,79 @@
...
+
+3) avoid the malicious attacks achieved by sending large amount of `CreateMiner` messages with low cost that congestion and unstable gas fees
```
kaitlin-beegle, line 24:
```suggestion
3) Avoid an attack vector whereby an attacker sends a large amount of low cost `CreateMiner` messages, thus causing network congestion and destabilizing gas fees. 
```

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Problem Motivation
+
+From @Jennijuju mentions: https://github.com/filecoin-project/FIPs/discussions/780#discussioncomment-6787043.  
+this was a real issue occurred in March 2021 where someone spammed the network with thousands of thousands of CreateMiner messages with very cheap cost. It caused a couple issue: 1) base fee increases and congests the mpool 2) stress some nodes from block producing opportunities 3)chain performance (cron jobs overload - which was addressed in nv12 upgrade)
+
+By August 2023, according to the statistics of Filscan, there are 601,131 Miner IDs have been built with only 4,862 actives. The actual ID usage counts for merely 0.8% and the abuse of ID creation trend continues. The actual ID usage only counts 3.7% for the recent two months (125 Miner IDs are active out of 3,254 newly created).
```
kaitlin-beegle, line 31:
```suggestion
It [was documented](https://github.com/filecoin-project/FIPs/discussions/780#discussioncomment-6787043) that in March 2021, someone had spammed the network by sending thousands of CreateMiner messages.  This caused several issues: 
1) Base fee increases associated with mpool congestion
2) Redirect nodes away from producing blocks
3) Lower overall chain performance, specifically related to cron jobs overload

As of August 2023, according to FilScan statistics, there were 601,131 Miner IDs and only 4,862 actively in use. The actual ID usage counts for merely 0.8% of all existing Miner IDs, and the abuse of ID creation trend continues. In the previous two months, only 3.7% of all newly created Miner IDs (125 Miner IDs are active out of 3,254 newly created) were active. 
```

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,79 @@
...
+
+<img width="865" alt="total_in_use" src="https://github.com/remakeZK/FIPs/assets/92775781/fc8266ab-4224-49ab-a4a7-6450377d4638">
+<img width="913" alt="recent_two_months" src="https://github.com/remakeZK/FIPs/assets/92775781/2b41e3a8-ad5f-481a-bbbb-569fc4533e0b">
+
+  
+Many users randomly create numerous miner actors just to get preferred IDs, wasting resources and burdening the network for no purpose. Moreover, those wasted miner actors occupy the storage of the network; with time elapsed, the waste resource rate will increase, along with increasing the future gas price because of those unnecessary storages of the network. Furthermore, there is a risk of exposure to potential attacks by creating numerous accounts, one can cause network mpool congestion by sending numerous `CreateMiner` messages yet with low cost, which leads to unstable network base fee and potentially blocks more useful/critical/time sensitive network messages (i.e: SubmitWindowedPoSt, ReportConsensusFault, sector/data onboarding messages, to land on-chain.
```
kaitlin-beegle, line 37:
```suggestion
Users create multiple Minder IDs in order to get preferred IDs, wasting resources and unnecessarily burdening the network. Unused Miner IDs continue to occupy storage on the network.  As time goes on, it is expected that the rate of wasted resources will increase, as will future gas prices.  

Furthermore, allowing network participants to cheaply and easily create numerous accounts introduces a potential attack vector.  Were participants to spam the network with CreateMiner messages, mpool congestion and unstable network fees could potentially affect more useful, critical, and/or time sensitive messages (i.e: SubmitWindowedPoSt, ReportConsensusFault, sector/data onboarding messages) from landing on chain.
```

kaitlin-beegle (approved):
Editorial review in!

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
remakeZK: Because the amount to each miner actor creation will vest in the locked rewards for 180 days, which means this fund will eventually return to the person. It does not really count as a “fee” since there is no cost but opportunity cost.

remakeZK:
> Thank you @remakeZK for opening the FIP draft! first editorial pass done with some suggestions

@jennijuju , hello:) I wanted to kindly follow up and see if there's anything we can discuss to move this forward. Please let me know, thanks so much!

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
anorth: I concur this value needs a rationale, and somehow tying it to pledge seems like a good option.

anorth (approved):
This is fine as a draft of draft, but I don't think is ready for last call until the pledge amount is determined on some rational basis (which should then be documented here).

anorth:
Please update to use FIP number 0077, and update the table in the README too

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,85 @@
...
+2) Prevent wasted resources and incremental increases in gas cost caused by wasted storage, and;
+
+3) Avoid an attack vector whereby an attacker sends a large amount of low cost `CreateMiner` messages, thus causing network congestion and destabilizing gas fees. 
+
+## Problem Motivation
+
+It [was documented](https://github.com/filecoin-project/FIPs/discussions/780#discussioncomment-6787043) that in March 2021, someone had spammed the network by sending thousands of CreateMiner messages.  This caused several issues: 
+1) Base fee increases associated with mpool congestion
+2) Redirect nodes away from producing blocks
+3) Lower overall chain performance, specifically related to cron jobs overload
```
momack2, line 31: work was done in a past network upgrade (@jennijuju can you remind which one to include here?) to reduce the network burden of these inactive miner_ids - they are no longer a significant impact on chain quality or cron execution times.  Right now including this information here without the info on the resolution deployed implies this is still a significant chain health issue - which it isn't

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
```diff
@@ -0,0 +1,85 @@
...
+
+https://github.com/filecoin-project/builtin-actors/pull/1398
+
+## Security Considerations
+
+There is no security consideration for this proposal; instead, it helps to prevent potential attacks by using a large number of newly created IDs, and the scenarios of network congestion and instability of gas fees that are caused by unnecessary numerous ID creations. 
+
+## Incentive Considerations
+
+Charging for creating new accounts can help prevent abuse, reduce unnecessary storage, and avoid increased gas fees over time caused by such wasteful activities that burden the network with unnecessary storage.
```
momack2, line 76: We should also include that requiring a creation fee will be a moat against new SPs entering the Filecoin ecosystem, or existing SPs who want to scale their deployments horizontally instead of vertically (within the same miner_id). How does the chosen deposit cost compare to the other necessary onboarding costs for a new miner (ex, what percent of the total hardware / pledge / sealing costs to reach the 10 TiB threshold to start earning block rewards is 500 FIL today?)

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
momack2: when will that 500FIL be unlocked? never?

remakeZK: > I concur this value needs a rationale, and somehow tying it to pledge seems like a good option.

if we need a rationale for setting the deposit amount, maybe we could choose using what @hunjixin suggested using the waterline storage deposit as the standard, multiple by the pledge + sealing cost which is 6FIL/TiB, turns to 10t*6Fil/t = 60 FIL as the fee? Will this work?

remakeZK: > when will that 500FIL be unlocked? never?

the 500 FIL amount could change, but this amount will be unlocked within 180 days, just like the locked rewards.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
remakeZK: > We should also include that requiring a creation fee will be a moat against new SPs entering the Filecoin ecosystem, or existing SPs who want to scale their deployments horizontally instead of vertically (within the same miner_id). How does the chosen deposit cost compare to the other necessary onboarding costs for a new miner (ex, what percent of the total hardware / pledge / sealing costs to reach the 10 TiB threshold to start earning block rewards is 500 FIL today?)

Yes, that’s a really good point, already added it!  As for the deposit amount, Alex also mentions that we need a rationale reason for setting it. the 500 FIL was referred to as the suggestion from the discussion board before, so I changed the amount to it. but that’s great to consider tying up the cost of the 10 Tib threshold! I also replied above, actually @hunjixin also suggested this as well. so, if we consider to tying the 10 Tib threshold and the cost of pledge+sealing which is 6FIL/Tib, to make the deposited amount as 60FIL in the locked rewards and allow it to unlock within 180 days, will that work?

steven004: The question here will be why 10Tib as a threshold is a good choice, or even it was, how's now? I would think there is not that reasonable to match 10Tib pledge. 
I think what @momack2 mentioned is we need have a description of the product consideration of how's this design impact. From my point of view, it is almost no impact to SPs, since the SP's requirement of pledge is much larger than 500FIL. They need prepare for it at creation of the minerID, at the same time, the 500Fil is vested linearly in 180days, which could be used as pledge to onboard sectors. But, this design could very effectively block abuse of resources.

anorth: Just to be clear, I don't think the FIP should have any number like 60 FIL that is calculated from the cost of 10 TiB, it should _specify the calculation_ that should be executed every time, so that the cost stays up to date as the underlying economics (e.g. pledge amount) change. The 10 TiB number can refer to the existing protocol constant about minimum consensus raw byte power.

kaitlin-beegle:
Thanks for so proactively incorporating edits, @remakeZK.  This looks great as a draft; I'm going to go ahead and merge. Please note that the draft is still a draft, even once merged.  There are a few outstanding edits, etc., above that will need to be addressed before we can schedule this for Last Call.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
steven004: I suggest considering Rational from two angles:

1. **Adequate Opportunity Cost to Prevent Abuse of CreateMiner:** The cost of 500 FIL is significantly high for maliciously creating numerous Miner IDs, acting as a deterrent. This also applies to Storage Providers (SPs) looking to reserve attractive Miner IDs. The capital requirement makes it challenging to create too many Miner IDs. Additionally, creating more Miner IDs would result in increased locked funds, which would be beneficial for FIL demand. 

2. **Minimal Burden for SPs Who Genuinely Need to Create Miner IDs:** One way is to analyze a specific time frame (e.g., half year from the height at which a miner get its power) to do pledge collateral statistics. How much is the pledge collateral these Miners have within six months? Alternatively, we can directly evaluate the peak QAP of all Miners if it is easier.

If nearly 100% of properly functioning Miners have pledges exceeding 500 FIL within six months, with most far exceeding this threshold, it suggests that 500 FIL is not a burden for a properly functioning Miner.

FIPS/FIP-XX-add-cost-opportunity-for-new-miner-creation.md
steven004: @remakeZK , Is it possible to have statistic data regarding how long the miners will pledge 500+ FIL after their first sector committed to the network, and the proportion rate of miners who never pledge 500+ FIL?  

Or at 95 percentile, how much FIL pledged in half year from an active miler. If it is much larger than 500FIL, then the requirement of 500FIL deposit should be fine, or we may need to cut it down to some level.

PR #788, opened 2023-08-16
Draft of Phasing Out Fil+ and Restoring Deal Quality Multiplier to 1x

dcasem:
Draft of Phasing Out Fil+ and Restoring Deal Quality Multiplier to 1x (Option 1 of Discussion 774)

PR #803, opened 2023-08-23
Draft of Removing Restriction to Access to Multiplier and Allowing Unrestricted Minting of Datacap

dcasem:
Draft of Removing Restriction to Access to Multiplier and Allowing Unrestricted Minting of Datacap (FCM Option)


FIPS/FIL+ FIP.md
```diff
@@ -0,0 +1,86 @@
+
+# FIPXXXX: Phasing Out Fil+ and Restoring Deal Quality Multiplier to 1x
```
anorth, line 2: You appear to have inadvertently included the file from #788 in this PR as well. I assume it's not intended. Can you remove it here (it should stay in #788)?

anorth:
This proposal is technically well specified and implementable. It's kinda light on motivation and analysis of impacts, but much of that is present in the discussion. I do appreciate this brevity vs copy-pasting the discussion post into the FIP. We might want a _little_ more included inline before we could consider this final, but as a first draft this is describing a pretty clear change.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-22
+requires: 
+replaces:
+```
+
+## Simple Summary
+This proposal aims to remove restrictions associated with the 10X multiplier in Fil+ deals and allow any actor to mint Datacap, promoting decentralization and simplifying the Filecoin ecosystem.
```
jennijuju, line 15: The title suggests "Removing Restriction to Access to Multiplier" & "Remove restrictions associated with the 10X multiplier in Fil+ deals"  - what does this mean? 

are you suggesting who may call `AddVerifier` and `AddVerifiedClient`

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Abstract
+The Fil+ program was designed to encourage high-quality storage on the Filecoin network. The current proposal intends to remove complexities that have arisen from its initial conception, thereby ensuring the Filecoin mission is better represented and uncomplicated.
```
jennijuju, line 18: High-quality storage sounds ambiguous to me, as one can also provide "high quality" committed capacity storage on Filecoin.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Change Motivation≠
+The current Fil+ implementation, though well-intentioned, has added complexities that deviate from Filecoin's core mission. Simplifying this system will encourage the onboarding of genuine data and uphold the integrity of the Filecoin blockchain.
```
jennijuju, line 21: Can you elaborate on the complexities and the issue/challenges associated with them?  (They were discussed in the discussed, lets summarize the points that your proposal can resolve. We should ensure FIP readers may get enough context without going through the whole fip discussion)

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Specification
+ **Change in the Datacap Actor's Mint Function**: Alter the DataCap actor's mint() function to remove the restriction that the caller must be the "governor" (the verified registry actor). Allow any actor to mint datacap.
+
+## Design Rationale
+While more comprehensive solutions might be considered in the future, this FIP offers an immediate, simpler approach. 
+
+Allowing any actor to mint datacap is the simplest change to the current protocol that removes the Fil+ process restrictions on datacap issuance. This datacap can then be transferred to the verified registry to create a verified allocation and claim the associated multiplier when onboarding data.
```
jennijuju, line 29:
```suggestion
Allowing any actor to mint datacap is the simplest change to the current protocol that removes the Fil+ process restrictions on datacap issuance. Datacap can then be transferred to the verified registry to create a datacap allocation. Storage Providers can claim the allocation assigned to them and get the associated 10x deal multiplier when onboarding the data piece.
```

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+After this change, the verified registry and datacap actors will be mostly obsolete. Allowing unlimited datacap does not create any complexity, but does make it possible to remove some. Future protocol improvements could remove these actors and support direct specification of a multiplier when onboarding to further simplify the protocol and reduce cost.
```
jennijuju, line 30: as suggested above - specify the complexity it addresses and how plz

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+After this upgrade, Storage Providers (SPs) can engage with any deal, whether Fil+ or not by minting Datacap. Deals will still need the `verified` flag set to true, but clients will be able to freely do this because they can mint the required datacap.
+
+
+## Test Cases
+Test that an account actor that is not a notary can mint datacap
+Test that a non-account actor can mint datacap
+
+## Security Considerations
+This FIP, though providing simplified access, doesn't introduce new security threats.
```
jennijuju, line 43: I think we will need analysis on this -what is the cost for any/group of malicious actors  to mint large amount of the data cap and onboarding CC sectors with the 10x multiplier and get>30% of the network QAP.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Test Cases
+Test that an account actor that is not a notary can mint datacap
+Test that a non-account actor can mint datacap
+
+## Security Considerations
+This FIP, though providing simplified access, doesn't introduce new security threats. 
+
+## Incentive Considerations
+The proposed change effectively removes a significant incentive for the onboarding of data that was previously determined as valuable by the off-chain notaries under the Fil+ mechanism. As a result, it's anticipated that the overall rate of data onboarding may decrease. The extent to which the currently incentivized data under this mechanism is fraudulent remains uncertain.
```
jennijuju, line 46: >As a result, it's anticipated that the overall rate of data onboarding may decrease

Are you suggesting that this FIP, will reduce the network storage utility?

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+With the removal of the restriction, effectively-empty sectors can now benefit from a multiplier. This adjustment allows Storage Providers (SPs) to onboard more Raw Byte Power (RPB) at a potential profit, potentially leading to an increase in such sectors being onboarded.
```
jennijuju, line 48:
```suggestion
With the removal of the restriction, effectively-empty (a.k.a committed capacity) sectors can now benefit from a multiplier. This adjustment allows Storage Providers (SPs) to onboard more Raw Byte Power (RPB) at a potential profit, potentially leading to an increase in such sectors being onboarded.
```

jennijuju: > With the removal of the restriction, effectively-empty sectors can now benefit from a multiplier. This adjustment allows Storage Providers (SPs) to onboard more Raw Byte Power (RPB) at a potential profit, potentially leading to an increase in such sectors being onboarded.

please elaborate what are the incentives that will bring to the network/SP by enabling onboard more CC/RBP and drive towards potential profit.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+Following this amendment, primary motivations for SPs to onboard data will still need to originate from outside the core protocol. This could encompass direct client payments or potential incentive models established within upcoming user-defined smart contracts.
+
+
+## Product Considerations
+Clients and Storage Providers will experience a streamlined interaction with the Filecoin network, which could lead to more user-friendly products and services built on top of Filecoin.
```
jennijuju, line 54: https://github.com/filecoin-project/FIPs/pull/803/files#r1309506511

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+Following this amendment, primary motivations for SPs to onboard data will still need to originate from outside the core protocol. This could encompass direct client payments or potential incentive models established within upcoming user-defined smart contracts.
+
+
+## Product Considerations
+Clients and Storage Providers will experience a streamlined interaction with the Filecoin network, which could lead to more user-friendly products and services built on top of Filecoin. 
+
+Fil+ team and many more customized deal markets can be developed on L2 of Filecoin in either permissioned or permissionless manners in an open and fair market competition. 
+
+The proposed changes present SPs with a more flexible environment, allowing them to select a pledge-to-hardware ratio best aligned with the specific costs and benefits they encounter. This enhanced flexibility means SPs can make more rational decisions based on their individual circumstances.
```
jennijuju, line 58: Could you please add how would the proposed change impact the data client?

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
+```---
+fip: "FIP-XXXX"
+title: "Removing Restriction to Access to Multiplier and Allowing Unrestricted Minting of Datacap"
+author: Fatman13 (@Fatman13), flyworker (@flyworker), stuberman (@stuberman), Eliovp (@Eliovp), dcasem (@dcasem), and The-Wayvy (@The-Wayvy)
+Discussions-to: https://github.com/filecoin-project/FIPs/discussions/774
```
jennijuju, line 5: I would suggest we create a new FIP discussion just for this proposed solution

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
+```---
+fip: "FIP-XXXX"
+title: "Removing Restriction to Access to Multiplier and Allowing Unrestricted Minting of Datacap"
```
kaitlin-beegle, line 3: For my own clarity (apologies if it's been mentioned elsewhere, but it's not clear in the PRs)- can you explain how we should understand this FIP, as it relates to #788? I haven't had time to closely follow the discussion, which is happening across multiple spheres.  I assume you opened this PR as an acceptable alternative to #788?

FIPS/fip-FCM.md
kaitlin-beegle, line 15: [editorial nit...not a big deal].  

I would remove the "promoting decentralization and simplifying the Filecoin ecosystem" piece of this.  It's not necessary.

I would simply state: "This proposal aims to remove restrictions associated with the 10x multiplier offered to Fil+ deals.  Specifically, it does so by allowing any actor to mint Datacap so long as....[insert condition here].

anorth: I don't think any change to AddVerifier/Client is needed to accomplish the authors goal here. Those only serve to funnel the datacap minting process through the verifreg. Am I missing something?

FIPS/fip-FCM.md
anorth: I don't think this FIP has any place to specify how a future FIP should do that. It's enough to point out that it will be possible.

FIPS/fip-FCM.md
kaitlin-beegle: ^^ I think a lot of this language is ambiguous.  

If the simple summary gives a 1-2 sentence overview of the changes you want, the abstract should add a bit more detail about why. Saying that you want to "remove complexities" and "ensure the Filecoin mission is better represented" is a unclear.  Just say **what** the complexities are, and **how** your proposed changes service the Filecoin mission. 

Re: @jennijuju's points above, one definition that you can use for Fil+ is the one in FIP0003: "... maximize the amount of useful storage that Filecoin can and will support." 

I want to help you avoid vagueness as much as possible.  If you have vague language and terms in your FIP draft, it makes it harder to navigate the soft consensus/last call process.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+Allowing any actor to mint datacap is the simplest change to the current protocol that removes the Fil+ process restrictions on datacap issuance. This datacap can then be transferred to the verified registry to create a verified allocation and claim the associated multiplier when onboarding data.
+After this change, the verified registry and datacap actors will be mostly obsolete. Allowing unlimited datacap does not create any complexity, but does make it possible to remove some. Future protocol improvements could remove these actors and support direct specification of a multiplier when onboarding to further simplify the protocol and reduce cost.
+
+## Backwards Compatibility
+This proposal does not change any chain state or method interfaces so is backwards compatible to all clients. A network upgrade is required to deploy the new actor code.
+
+After this upgrade, Storage Providers (SPs) can engage with any deal, whether Fil+ or not by minting Datacap. Deals will still need the `verified` flag set to true, but clients will be able to freely do this because they can mint the required datacap.
```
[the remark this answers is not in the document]
anorth: Good point. This is currently acheived by `AddVerifiedClient` (which should really be called AddDataCapToVerifiedClient). _Instead_ of that, clients (or anyone) will need to mint the datacap by explicitly invoking the Datacap actor's `mint` method. The old methods all still work, there's no backwards incompatibility. But the new flow uses a more direct method (the client can mint their own instead of requiring a verifier to do it for them).

FIPS/fip-FCM.md
anorth: @jennijuju what is the editorial change you are requesting before considering this a valid FIP draft?

FIPS/fip-FCM.md
anorth: @jennijuju they won't be CC sectors. They will have (self-)deals in them because that's how to onboard the datacap.

anorth: This would actually be a strong motivation to follow on with the clean-up work of removing datacap and verifreg if this FIP is accepted.

anorth: The incentive to onboard is just block rewards, the same incentive currently for onboarding 10x sectors today.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+
+## Test Cases
+Test that an account actor that is not a notary can mint datacap
+Test that a non-account actor can mint datacap
+
+## Security Considerations
+This FIP, though providing simplified access, doesn't introduce new security threats. 
+
+## Incentive Considerations
```
kaitlin-beegle, line 45: I still have some hesitation about the doneness of this proposal, and I think it has to do with this section. 

@jennijuju and I each flagged above that we'd like more specific language, specifically where "complexities" and "simplification" are referenced. 

I'd also like to see a bit more documentation about _the assumptions/observations driving this desired change_, and _what it actually means for anyone to be able to mint datacap_. I know what they are, and they've been well discussed by folks engaging on this discussion topic.  Those details should at least be summarized here as well (or below in Product Considerations; doesn't matter too much). 

This FIP should not just direct implementers, but should also justify and explain the policy. It would be great for you to add a few sentences about how those minting datacap will (presumably) be opting for higher collateral rates, and why they may choose to do so.  Why would people choose to mint datacap if its otherwise available to everyone? Or rather, why would someone NOT mint datacap if it's giving them a 10x multiplier? Etc. 

I know you know these things; they should be here too.

kaitlin-beegle (changes requested):

FIPS/fip-FCM.md
The-Wayvy: The motivation for this FIP is less the complexities of FIL+ and more that it’s permissioned.

We talked about this at length in 774.

FIPS/fip-FCM.md
The-Wayvy: Yes, this is an alternative.

FIPS/fip-FCM.md
The-Wayvy: Create a new discussion, miss the deadline, and then have to go through another 6 months of permissioned nonsense.

No.

FIPS/fip-FCM.md
The-Wayvy: The SPs are the real users of the network at the moment because they pay for the vast, vast majority of Filecoin blockspace / transaction fees.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-22
+requires: 
+replaces:
+```
+
+## Simple Summary
+This proposal aims to remove restrictions associated with the 10X multiplier in Fil+ deals and allow any actor to mint Datacap, promoting decentralization and simplifying the Filecoin ecosystem.
```
The-Wayvy, line 15: I would replace 

> promoting decentralization and simplifying the Filecoin ecosystem.

with

> thus making Filecoin permissionless again.

FIPS/fip-FCM.md
The-Wayvy: I would replace

> The current proposal intends to remove complexities that have arisen from its initial conception, thereby ensuring the Filecoin mission is better represented and uncomplicated.

with

> the current proposal intends to remove the permissioned notary system, ensuring that Filecoin remains decentralized and fair.

FIPS/fip-FCM.md
The-Wayvy: I would replace

> The current Fil+ implementation, though well-intentioned, has added complexities that deviate from Filecoin's core mission. Simplifying this system will encourage the onboarding of genuine data and uphold the integrity of the Filecoin blockchain.

with

> The current Fil+ implementation, though well-intentioned, has created an uneven playing field for SPs, permissioned the chain, and encouraged undesirable behaviors like bribery and false accusations. Our proposal makes Filecoin fully permissionless. It will remove these problems and restore confidence in the integrity and neutrality of our consensus algorithm

FIPS/fip-FCM.md
ArthurWang1255: Filecoin is a blockchain, and the users of a blockchain are those who hold and consume Filecoin through gas. The impact to data clients are not directly related to the filecoin core protocol. IMHO, the fertility of RBP and SPs is what needs to be considered in a fip.

FIPS/fip-FCM.md
ArthurWang1255: The security of the network will actually be increased by the fip. In the past, only 1/10 of the hardware + enough DC was needed to do evil. Now it needs 1:1 hardware to be evil.

FIPS/fip-FCM.md
The-Wayvy: Agree with Arthur.

A fair, transparent chain with a lower barrier to entry will attract more SPs.
All else equal that should mean more storage, cheaper storage, and better storage.

The notary system doesn't incentivize useful storage.
It incentivizes bribery and fraud.
The scale and sophistication of the fraud/bribery just increases with the size of the bureaucracy.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+fip: "FIP-XXXX"
+title: "Removing Restriction to Access to Multiplier and Allowing Unrestricted Minting of Datacap"
+author: Fatman13 (@Fatman13), flyworker (@flyworker), stuberman (@stuberman), Eliovp (@Eliovp), dcasem (@dcasem), and The-Wayvy (@The-Wayvy)
+Discussions-to: https://github.com/filecoin-project/FIPs/discussions/774
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-22
+requires: 
+replaces:
```
kaitlin-beegle, line 11: Please specify that this would need to replace FIP0003.

The-Wayvy: Sure

The-Wayvy, line 11: replaces FIP0003

The-Wayvy (changes requested):

FIPS/fip-FCM.md
The-Wayvy, line 11: replaces: FIP0003

FIPS/fip-FCM.md
The-Wayvy, line 15: This proposal aims to remove restrictions on access to the 10X multiplier in Fil+ deals and allow any actor to mint Datacap, thus making Filecoin permissionless again.

FIPS/fip-FCM.md
The-Wayvy, line 18: The Fil+ program was designed to encourage 'useful' storage on the Filecoin network. The current proposal intends to remove the permissioned notary system. This will ensure that Filecoin remains decentralized, transparent and fair. The power to define usefulness is transferred to users.

FIPS/fip-FCM.md
The-Wayvy, line 21: The current Fil+ implementation, though well-intentioned, has created an uneven playing field for SPs, permissioned the chain, and encouraged undesirable behaviors like bribery, fraud and framing competitors. Our proposal makes Filecoin fully permissionless. It will remove these problems and restore confidence in the integrity and neutrality of our consensus algorithm

The-Wayvy (changes requested):

FIPS/fip-FCM.md
momack2: Jennifer is requesting a new discussion so that folks who read this proposal have a place to go comment and discuss it going forward that isn't already crowded with many competing/alternate proposals that are different to the one described here. This enables clearer discussion and debate with a new draft proposal.

FIPS/fip-FCM.md
momack2, line 15:
```suggestion
This proposal aims to allow any actor to mint Datacap and bestow a 10X multiplier to any sector, regardless of whether it stores client data / serves FIL+ deals.
```
Seems like this is a more accurate and descriptive summary for this proposal

FIPS/fip-FCM.md
momack2: Can you please explain how this proposal will "encourage the onboarding of genuine data"? This seems core to your argument, but is unsubstantiated in this proposal.

FIPS/fip-FCM.md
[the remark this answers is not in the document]
momack2: how does this change interface with the new direct data onboarding proposal here (https://github.com/filecoin-project/FIPs/pull/804/files)?

FIPS/fip-FCM.md
momack2: I agree with Jennifer that this section seems incomplete. Security consideration should be given to: 
- could a large CC-sector miner use this change to quickly upgrade all their CC sectors into fake "10x deal" sectors (without actually onboarding any real/useful client data) and use that to gain control of the network?
- how does this change harm the security of **data** on the network? (ex it significantly decreases the confidence of existing Filecoin clients that SPs will continue to store and safeguard their data - since they could instead dump their sectors full of real client data and get the same protocol incentive for storing strings of 0s - which are easier to repair and come with no client expectations)

FIPS/fip-FCM.md
momack2: How does this proposal contend with the attack vector that both FIP36 and FIP56 ran into - that an increased multiplier actually has a *downward* pressure on RBP (and therefore block rewards)?
- Most SPs are **capital-limited**, not hardware-limited, and will therefore onboard 1/10th the number of sectors for the same collateral requirements (causing a 10x decrease in RBP)
- The final version of the FIP56 proposal decreased the multiplier for any non-deals sector to a max of 2x to ensure the network RBP would not contract massively if SPs rationally chose to terminate their existing storage and recommit only 1/Xth

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Security Considerations
+This FIP, though providing simplified access, doesn't introduce new security threats. 
+
+## Incentive Considerations
+The proposed change effectively removes a significant incentive for the onboarding of data that was previously determined as valuable by the off-chain notaries under the Fil+ mechanism. As a result, it's anticipated that the overall rate of data onboarding may decrease. The extent to which the currently incentivized data under this mechanism is fraudulent remains uncertain.
+
+With the removal of the restriction, effectively-empty sectors can now benefit from a multiplier. This adjustment allows Storage Providers (SPs) to onboard more Raw Byte Power (RPB) at a potential profit, potentially leading to an increase in such sectors being onboarded.
+
+Following this amendment, primary motivations for SPs to onboard data will still need to originate from outside the core protocol. This could encompass direct client payments or potential incentive models established within upcoming user-defined smart contracts.
```
momack2, line 50: Can you give a rough estimate on the drop in current inbound client demand for Filecoin that would drop off given the scale required for these "direct client payments", either here or in the product consideration section?

FIPS/fip-FCM.md
momack2: Can you please clarify why you expect more user-friendly product and services to exist without client data storage incentives?

FIPS/fip-FCM.md
momack2: 1. FIL+ is not limited to the notary system. There are 4 different data cap distribution programs within FIL+ - one of which automatically gives _any_ github acct that requests it 64GB of datacap per month. 
2. Data storage clients are a valued and critical constituency within the Filecoin network. If your proposal aims to disenfranchise them in favor of SPs, you need to say so explicitly in your proposal.

momack2 (changes requested):

FIPS/fip-FCM.md
raghavrmadya: I would like to share that there is no concrete evidence of bribery, even though multiple claims have been made about secondary markets for DC. Additionally, all accusations, as they concern abuse of DC, are well founded in evidence here - https://www.notion.so/filecoin/T-T-Dispute-Tracker-d28b93677cb544b48e77b585856601cf

FIPS/fip-FCM.md
raghavrmadya: If we are going to take the current state of the world where SPs have access and influence ( partly because the network spent the initial years focusing on the supply side) and use that as an argument to build the future state, then how do we create a place for clients in this network? If data clients are arguably a negligible constituency, this should be a concern for us and we should find ways to support this constituency.

FIPS/fip-FCM.md
galen-mcandrew: Agree with this edit, it is more accurate than unsubstantiated commentary around decentralization, simplification, or permissionlessness.

FIPS/fip-FCM.md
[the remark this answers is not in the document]
galen-mcandrew: Will this proposal change existing RKH and Notary chain-states? Will clients still be able to choose to go through an existing notary process, such as Glif Autoverification?

FIPS/fip-FCM.md
galen-mcandrew: Currently, DataCap has no 'expiration point' and can only be destroyed by notary and RKH message (per FIP0028) or burned in deals.
Are there additional concerns of actors minting massive amounts of DataCap? What would happen if a bot opened new client addresses and spammed that mint DataCap function, creating EiB of DataCap immediately? 
Should there be a threshold or gate on minting?

FIPS/fip-FCM.md
galen-mcandrew: Currently, we see 10.8EiB with 1.4EiB active Fil+ deals. Based on this proposal, one possible outcome is a continuing decrease in RBP with a massive increase in QAP. For example, what happens if the RBP to active 'DataCap' deals ratio is closer to 1:1? Has there been any modeling on how the block reward calculation would be impacted?

FIPS/fip-FCM.md
anorth: @raghavrmadya thanks for your contributions here, but this PR review should be fairly strictly focussed on editorial review of the content. It's not the place to ask or answer questions like this. Please do raise them in the discussion thread item, though.

FIPS/fip-FCM.md
kaitlin-beegle: Great. Does this mean that we can close the other PR? @The-Wayvy

FIPS/fip-FCM.md
The-Wayvy: Molly, the core of our argument is making Filecoin Permissionless / providing an even playing field for all participants / neutrality and objectivity.

I think by now you get our views.
@momack2

FIPS/fip-FCM.md
15012700225: What's the difference between fil+casting datacap and free casting datacap, especially given the amount of abuse?
What I want to know is why you don’t need to consider security issues when you turn on fil+ casting datacap, but you need to turn on free casting?

Is it more straightforward to switch to staking mining (POS) in one step?

FIPS/fip-FCM.md
The-Wayvy: @kaitlin-beegle 
That's David's decision.

FIPS/fip-FCM.md
The-Wayvy: @galen-mcandrew 

Our commentary on permissionlessness is pure fact.

FIPS/fip-FCM.md
The-Wayvy: @galen-mcandrew 

This is something for CEL to analyze.

FIPS/fip-FCM.md
cryptowhizzard: This is really strange. Fil+ is about onboarding valuable data and onboarding this takes effort, resources and time.

If you want to onboard crap and trash I think people are welcome to do that, just not with 10x.

And why is a permissioned system bad? If the goal is onboarding true and real data there is nothing wrong with it. The argument “no market” makes no sense at all, there is market and there are customers atm. If those customers don’t want have anything to do with company’s who are not suiting them well for their needs it is up to them to change and make themselves competitive.

hyunmoon: Could the decision to weaken Filecoin Plus at this time speed up our move to a future where most storage deals are paid?
If yes, could you please explain why?

hyunmoon: While I am not overly concerned about being permissionless, I do prioritize a steady increase in market capitalization.
I believe that an increase in paid deals is crucial to achieving this objective.

FIPS/fip-FCM.md
The-Wayvy: That is CEL's job.

If you are worried about it then you can do the analysis yourself.

The-Wayvy:
@momack2 

Stop putting words in our mouths

> Can you please explain how this proposal will "encourage the onboarding of genuine data"? This seems core to your argument, but is unsubstantiated in this proposal.

The core of our argument is that Filecoin Plus creates an uneven playing field, and that non-user's judgements about quality are not a solid foundation for a consenus algorithm. 

We covered this in 774. That is where the 'philosophical' part of the debate occurred.

FIPS/fip-FCM.md
The-Wayvy: @cryptowhizzard 

You are a notary and do not want to lose your privilege.
That is why you defend the current system.

@hyunmoon 
You have made large investments in Filecoin Plus-compatible hardware.
That is why you defend the current system.

The-Wayvy: > This is really strange. Fil+ is about onboarding valuable data and onboarding this takes effort, resources and time.

The question is: who should decide what is valuable?

Users will decide what is valuable and what is a good use-case for our technology.

The governance team and notaries are not the ones paying and therefore should not be making decisions about what is quality and useful.

Having non-users dictate 'quality' and 'utililty' is speculative at best and corrupt at worst.

> If you want to onboard crap and trash I think people are welcome to do that, just not with 10x.

Again, data you don't like is not 'crap' 

The usefulness of data = how much users will pay for it

You don't know what Filecoin's PMF will be. 

This is like saying Uniswap is crap because you can't trade 'real assets' like US Treasury Bonds and NASDAQ stocks.

> And why is a permissioned system bad? If the goal is onboarding true and real data there is nothing wrong with it. The argument “no market” makes no sense at all, there is market and there are customers atm. If those customers don’t want have anything to do with company’s who are not suiting them well for their needs it is up to them to change and make themselves competitive.

The whole point of having a sybil resistance mechanism on a network is to enable permissionless participation.
PoSt and PoS are sybil resistance mechanisms.
If you want a permissioned network then there is no need to waste so much capital in PoS and so much computation in PoSt.
We can KYC every storage provider and have them sign contracts with PL.

Literally the entire point of this technology is to enable permissionlessness.

True and Real data does nothing for the price of Filecoin is users aren't paying for it.

Also, if you believe in the market then at the very least all Filecoin Plus deals should be required to go up for public auction.

The-Wayvy: @hyunmoon 

If the governance team wants to play VC they can use their own money to subsidize 'real' users.

Their speculation/preferences are not a legitimate input to a crypto consensus algorithm.

FIPS/fip-FCM.md
The-Wayvy: Without the notary system there would be fewer steps to getting their data onboarded.
 
Storage incentives are fine, but they can't have a human in the loop.

FIPS/fip-FCM.md
The-Wayvy: @kaitlin-beegle 

if you add me as a collaborator I will speed up the editing process.

This was not David's first choice and he just wanted to get the PR out the door, not write a masterpiece.

anorth: @The-Wayvy we have no ability to give you edit access here since the source of the PR is a repository under @dcasem's GitHub account. He should be able to do so, though.

FIPS/fip-FCM.md
The-Wayvy: @momack2 
@galen-mcandrew 

Crypto-Economic Analysis is a job for... Crypto-Econ Lab

They've already done the work.

https://hackmd.io/zp-40inqSny5HMqmTSOKxw?view

FIPS/fip-FCM.md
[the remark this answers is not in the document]
The-Wayvy: @momack2 

That is a legitimate question. 

I will ping Alex.

FIPS/fip-FCM.md
The-Wayvy: 1

> I agree with Jennifer that this section seems incomplete. Security consideration should be given to:

could a large CC-sector miner use this change to quickly upgrade all their CC sectors into fake "10x deal" sectors (without actually onboarding any real/useful client data) and use that to gain control of the network?

Even in the worst case scenario where only one miner converts their RBP to 10x, and the others stick with CC
It would still take that malicious miner over 2 exabytes of RBP + all that new 10x collateral to get 50% of network power.

If you aren't convinced that that's unlikely then I hope you ask CEL.

2

> how does this change harm the security of data on the network? (ex it significantly decreases the confidence of existing Filecoin clients that SPs will continue to store and safeguard their data - since they could instead dump their sectors full of real client data and get the same protocol incentive for storing strings of 0s - which are easier to repair and come with no client expectations)

Real users will pay the price that a service is worth.

We keep telling you that you're in too much of a rush to get these big Web2 clients.

It takes a long time to build consumer applications on a blockchain.

Ex. Ethereum is not used for mortgages or trading TradFi stocks / bonds even after 8 years in production.

The first applications are going to be stupid: like CryptoKitties and Satoshi Dice.

Please be more patient.

3

There is no fake and real. 
No useful and useless.
Only paid and unpaid.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,72 @@
...
+fip: "FIP-XXXX"
+title: "Removing Restriction to Access to Multiplier and Allowing Unrestricted Minting of Datacap"
+author: Fatman13 (@Fatman13), ArthurWang1255 (@ArthurWang1255), flyworker (@flyworker), stuberman (@stuberman), Eliovp (@Eliovp), dcasem (@dcasem), and The-Wayvy (@The-Wayvy)
+Discussions-to: https://github.com/filecoin-project/FIPs/discussions/774
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-22
+requires: 
+replaces:
```
The-Wayvy, line 11: replaces: FIP0003

The-Wayvy (changes requested):

FIPS/fip-FCM.md
hyunmoon: I concur. The project should have commenced with a stipend rather than a subsidy from the network.
I think the situation wouldn't have been this problematic if the program had been managed by Protocol Labs employees.

hyunmoon: I concur. Ideally, the project should have been supported by a single, upfront allocation of funds.
I believe the situation wouldn't have become this problematic had it remained under the management of Protocol Labs employees, rather than being handed over to elected notaries.

Derek-zd:
> @momack2
> 
> It would be nice if you focused on the benefits of this proposal rather than cherry-picking individual phrases in a PR draft.
> 
> Marko summarized our position in 774, and many others made good contributions as well.
> 
> You could help us synthesize all those valid points into a cohesive narrative for why Filecoin Plus 1.0 is being retired.
> 
> > Can you please explain how this proposal will "encourage the onboarding of genuine data"? This seems core to your argument, but is unsubstantiated in this proposal.
> 
> The core of our argument is that Filecoin Plus creates an uneven playing field, and that non-user's judgements about the value of a data-set to Filecoin are guess-work and not an acceptable input to our consenus algorithm.
> 
> But I'll respond to this point: The current system incentives miners to take data that you would consider 'fake' and pass it off as 'real' to get the 10x multiplier. To get rid of the incentive for this charade, just get rid of notaries. Then 100% of user data onboarded will be real. We can be sure of that because no real user would pay to onboard useless data.
> 
> > FIL+ is not limited to the notary system. There are 4 different data cap distribution programs within FIL+ - one of which automatically gives any github acct that requests it 64GB of datacap per month.
> 
> > Data storage clients are a valued and critical constituency within the Filecoin network. If your proposal aims to disenfranchise them in favor of SPs, you need to say so explicitly in your proposal.
> 
> Nobody is disenfranchising users.
> 
> They are free to bid for services on Filecoin.
> 
> That is the whole point.
> 
> We want users themselves to decide what data and what transactions make it onto Filecoin.
> 
> Rather than the governance team and notaries.

oh, man. Why are you so ignorant of the current fil + situation, and why are you so naive。 No user is willing to pay to store the data in Filecoin。SP pays client to store the data in SP.
>We can be sure of that because no real user would pay to onboard useless data.

The-Wayvy:
@Derek-zd 

> oh, man. No user is willing to pay to store the data in Filecoin。SP pays client to store the data in SP.

I know that.

So let's drop the Fil+ charade about 'quality' and focus instead on improving the tech.

FIPS/fip-FCM.md
The-Wayvy: @raghavrmadya 

### It's too early for the use cases that Filecoin Plus is trying to subsidize

I believe the root of our problem is an extreme lack of patience.

We don't even have stablecoins or a DEX on Filecoin.
 
In fact we don't even have good oracles.

Ofcourse we aren't going to have Web2 giants using our tech.

There is still a lot of work to be done on very basic infrastructure.

We do have real users on the FVM and the miners themselves are users because they pay for blockspace.

Some are choosing to 
1.) deny the value of these real users.
2.) try to force certain use-cases and users on the network before Filecoin is ready for them.

Just keep building and have a little faith.

And please keep an open-mind to who the next set of real users and use-cases will be.

It probably will not be what you expect: just like few saw Uniswap or Cryptokitties etc coming.

FIPS/fip-FCM.md
The-Wayvy, line 11: replaces: FIP0003

The-Wayvy (changes requested):

FIPS/fip-FCM.md
jennijuju: @anorth No, im not referring to then future opportunities. 

> Allowing unlimited datacap does not create any complexity, but does make it possible to remove some.

I am asking the complexities that will be removed by this proposal to be included as a a part of the design rationale/product/incentive consideration.

FIPS/fip-FCM.md
jennijuju: I also dont think they are needed - thus im confused with the phrasing of "Removing restriction to access to multiplier" in the title -  to me it implies a change anyone can deploy their own multiplier and gets additional QAP thats other than the `deal  quality multiplier` that comes with datacap.  

That being said, this proposal should also call out the degraded functionality and/or the unnecessity  of  `AddVerifier` and `AddVerifiedClient` methods, with the introduction of "all you can mint" datacap rule.

The-Wayvy: @jennijuju 

You are the expert on implementation.

We appreciate your help.

FIPS/fip-FCM.md
The-Wayvy: @jennijuju 

No

FIPS/fip-FCM.md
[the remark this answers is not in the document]
anorth: @galen-mcandrew no it won't, all the existing workflows will still work fine (and the FIP should note this).

I've responded about interactions with direct onboarding [here](https://github.com/filecoin-project/FIPs/discussions/774#discussioncomment-6918983) (I don't see any).

Fatman13:
Should we keep the PR thread strictly between authors and FIP editors? Everything else should go to the discussion thread.

The-Wayvy:
I committed changes based on everyone's comments.

@anorth @jennijuju @jsoares @raulk @kaitlin-beegle @arajasek 

Can I get a merge?

FIPS/fip-FCM.md
anorth: Regarding the risk of rapid gains in power from upgrading CC, thanks for the napkin analysis but the point of the question is the FIP document must discuss this the presence and likely size of this risk, and any mitigating factors. You might also note that SPs would have to snap-deal to do that upgrade, which introduces a hardware constraint. But again, the point isn't to convince anyone here, the point is that the document needs to contain a reasonable discussion of the risk.

anorth: Re existing client data, I don't really think it's a security consideration, but it's definitely an incentive consideration. The change in incentives to SPs currently storing verified data is real. I tend to think the practical impact is low since "dumping" the data requires paying termination fees and re-sealing. But again, the FIP document must lay this out.

FIPS/fip-FCM.md
anorth: This substitution risk similar to FIP56 is important to explain in the FIP document. You can note that it's similar to risk of such behaviour if FIL+ was significantly more efficient and so datacap easier to get (even if all "good" data). It won't change the behaviour of SPs currently getting FIL+ much, which account for most onboarding today. It could change behaviour of others, but it's rather hard to say in which direction: balancing the substitution idea is the idea that for SPs who can't/won't get FIL+, onboarding _anything at all_ becomes profitable, an improvement to today.

This risk comes down to the ability of SPs to chose their hardware/data capital balance. I tend to this this could be good for SP business models (heterogeneous environments and setups can optimise for their comparative advantage) but it does introduce the _risk_ of this substitution having an impact on RPB. The FIP must discuss this.

It is not sufficient to say anything is someone else's job. This FIP document must describe all the relevant considerations. You can take some things from CEL's existing analysis. If more is needed you can ask them to help you, and they probably will. But no-one is obligated to do the work for you.

FIPS/fip-FCM.md
momack2: there already exists pathways to onboard data without FIL+ incentives or using the notary program - however we have not seen any user-friendly products or services building around that data onboarding flow (that I know of - maybe Telnyx offers one that is publicly accessible?). I don't think you have evidence to support your claim that better UX will just happen without economic incentives (ex, multiple data onboarding solutions - offering much better UX) have been created around the FIL+ data onboarding flow including Spade, Singularity, web3.storage, BDE, Motion, and many tools created directly by storage providers.

FIPS/fip-FCM.md
momack2: If this FIP doesn't discuss data security of client data here, then it needs to be explicitly discussed in the product section - both for new data onboarded (how secure is that data vs status quo) and data that's already been onboarded into the network. IMHO is appropriate to include in the security section because ensuring storage is not suddenly dropped or storage quality reduced is a security concern for achieving the core purpose of the Filecoin network - but as long as it's included clearly it could be in either section

FIPS/fip-FCM.md
Fatman13: How about the following edits regarding substitution risk? @anorth 

There may be different outcomes with regard to substitution risk, which refers to a CC SP who may expire its RBP and use the released collateral to achieve same amount of QAP power as before but with less hardware. In the scenario where we don't have [direct data on-boarding](https://github.com/filecoin-project/FIPs/discussions/730), the network growth would still be limited by the chain throughput of PublishStorageDeal messages which in turn minimizing substitution risk and effectively maintaining the current network status quo of RBP falling generally. In the scenario where we do have [direct data on-boarding](https://github.com/filecoin-project/FIPs/discussions/730), substitution risk would depend on how fast different SPs could react to the new FIP, recalibrate their hardware deployment to maximize their data on-boarding efficiency (hardware/data capital balance) and new chain throughput of direct data on-boarding message. Substitution risk in second scenario is also mitigated by the fact that CC SPs are expiring sectors anyway given the history CC expiring trend. Substitution in such case may even help to retain the SPs to stay in the network.

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,82 @@
...
+created: 2023-08-22
+requires: 
+replaces: FIP-0003
+```
+
+## Simple Summary
+This proposal removes restrictions on access to the 10X multiplier in Fil+ deals, allowing any actor to mint Datacap.
+
+## Abstract
+The Fil+ program intended to compensate miners for storing non-random data on Filecoin. It was hoped that these datasets would increase the utility of the network and thereby attract paying users. Fil+ did succeed in accelerating the development of data-onboarding tools. However, the notary system used to execute Fil+ has not brought paying data clients to Filecoin, and is riddled with perverse incentives. This proposal removes the permissioned, off-chain pathways to the 10x multiplier, and makes it available to all actors who stake the required collateral.
```
momack2, line 18: actually - there are paying customers utilizing FIL+ right now to achieve the cost-reductions necessary to make adopting Filecoin (instead of other storage networks) worth the transition costs. Agree we're still early days here - but this is an increasing trend that Fil+ has enabled.

It has also increased network utility -- there are now PiBs of valuable datasets (such as the recently rescued [Meta BioML dataset used for training protein folding language models](https://medium.com/@labdao/how-to-save-20-terabytes-of-scientific-public-goods-23151d24c6b2)) stored on Filecoin that are available for compute providers, DeSci builders, FVM smart contracts, and more to build their businesses on. There are [nearly 100M NFTs](https://nft.storage/stats/) stored on Filecoin (26B market cap). 

Maybe reframe this as: "Fil+ has accelerated the development of data-onboarding tools, increased network utility, and is in the early days of attracting and converting paying data clients to Filecoin -- however the notary system used by 3/4 Fil+ programs today is inefficient, un-objective, and riddled with perverse incentives."

FIPS/fip-FCM.md
```diff
@@ -0,0 +1,82 @@
...
+
+By removing notaries we will instantly eliminate all fraud, bribery and other harmful behavior in Filecoin. Under this proposal, miners can get the 10x multiplier as long as they post the required collateral, regardless of what kind of data they store, so there will be no reason for a miner to pretend their data is something that it's not. We'll also have complete certainty that all storage deals involve a real user and useful storage since no miner would go through the trouble of storing non-random data in a sector unless they were paid to do so, and no user would pay for storage unless they believed in its utility.   
+
+A more transparent, predictable and fair Filecoin will attract new miners, driving down the price of storage.
```
momack2, line 25: Agree this would likely attract new miners - but since a much smaller portion of miners will be engaging in data storage and there will be no incentives to subsidize data storage - I don't see how this would drive down the price of storage. I think more likely we'll see more miners join the network with zero intention to store data.

The-Wayvy: I do think that in the short-run most miners will join the network looking to just mine block rewards.

My _guess_ is that storage of user data will become the norm _evenutally_ because 

1.) Margins on CC mining will keep going down (ever more mining power + an ever smaller block reward) 
2.) We'll build applications that users are actually willing to pay for. 

IMO we're trying to get there too fast.

FIPS/fip-FCM.md
The-Wayvy: I briefly discussed this in a PR just now.

FIPS/fip-FCM.md
The-Wayvy: @momack2 

We would like to cut the multiplier to keep RBP from further contracting.

One variation on 803 would be to give everyone access to the multiplier, but also gradually cut the multiplier down to one.

The-Wayvy: Resolved

FIPS/fip-FCM.md
The-Wayvy: I changed it to: 

> Removing the notary system will streamline data onboarding for users as well as lower costs and uncertainty for miners.

Miners are still free to use all those data onboarding tools that you mentioned.

Under this proposal, the only economic incentive for miners to improve data-onboarding UX would be to get more paying users.

We think that's a good thing.

I added it to incentive considerations.

FIPS/fip-FCM.md
kaitlin-beegle: No, these are the same things. Words like "permissioned", "decentralized", and "fair" don't clearly frame your proposal. You need to say 1) what concrete change you're making, and 2) for what specific, concrete reason.  Adding vague words- especially ones like "decentralized", which can be interpreted and defined in many different ways- makes both governance and implementation harder to navigate. 

So, for example: 

"The Fil+ program was designed to encourage high-quality storage on the Filecoin network.  However, [insert specific problem you're trying to address].  To address this, we propose a change to the maintenance of datacap, such that datacap can be accessed by any network participant.  Doing so will enable us to solve [insert aforementioned problem] because [insert reason]." 

In other words, you're not making an argument; you're specifying a proposal.  Does that make sense?

FIPS/fip-FCM.md
The-Wayvy: I edited the FIP to mention that drops in demand are hard to quantify because
1.) We don't know what % of existing data on Filecoin has a real user behind it.
2.) We don't know what users are actually willing to pay for Filecoin services since they've been getting storage for free on Filecoin Plus.

FIPS/fip-FCM.md
The-Wayvy: I addressed this in security considerations.

FIPS/fip-FCM.md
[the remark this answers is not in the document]
The-Wayvy: I addressed direct data onboarding in "Specification"

FIPS/fip-FCM.md
The-Wayvy: I addressed how 803 will disincentivize the onboarding of data with no paying users and fabricated datasets in "Product Considerations"

FIPS/fip-FCM.md
anorth: I agree with @jennijuju and @momack2 here and have created https://github.com/filecoin-project/FIPs/discussions/844. This should not be taken as any rewinding of process, but a venue for discussion of this concrete proposal to aid a quicker governance process.

anorth:
I'm assigning FIP number 0078 to this proposal. @The-Wayvy please update the FIP document and the README.md to reflect this.

anorth (approved):
I don't see any reason not to merge this draft FIP. Merging the draft is not equivalent to declaring this FIP ready for last call. Merging the draft gives us a linkable document to refer to, and the ability to PR additional changes into it and track history.

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Specification
+ **Change in the Datacap Actor's Mint Function**: Alter the DataCap actor's mint() function to remove the restriction that the caller must be the "governor" (the verified registry actor). Allow any actor to mint datacap.
+
+ **Snap Deals**
+When an SP only wants to commit capacity, but do so with a high multiplier, the verified allocation they claim must specify a data CID, which must of course match the committed data. (Before direct onboarding, they would do a full-on deal with the built-in market; after direct onboarding, they can just directly commit the data). This would have the effect of giving the sector verified deal weight, and making it not appear to be CC, and therefore not eligible for subsequent snap deal with client data.
+
+We will work around this by detecting explicit commitments to the "Zero" unsealed CID, which is the same unsealed CID that CC sectors use. Even though this unsealed CID is not stored on chain, we can set a flag in sector on-chain info to indicate that the sector data is zero, despite having verified deal weight, and that it's therefore a valid target for snap deals.
```
jsoares, line 33:
```suggestion
 **Snap Deals**: When an SP only wants to commit capacity, but do so with a high multiplier, the verified allocation they claim must specify a data CID, which must match the committed data. (Before direct onboarding, they would do a full-on deal with the built-in market; after direct onboarding, they can just directly commit the data). This would give the sector verified deal weight, making it not appear to be CC, and therefore ineligible for subsequent snap deal with client data. We will work around this by detecting explicit commitments to the "Zero" unsealed CID, which is the same unsealed CID that CC sectors use. Even though this unsealed CID is not stored on chain, we can set a flag in the sector's on-chain info to indicate that the sector data is zero, despite having verified deal weight, and that it is a valid target for snap deals.
```

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Design Rationale
+While more comprehensive solutions might be considered in the future, this FIP offers an immediate, simpler approach. 
+
+Allowing any actor to mint datacap is the simplest change to the current protocol that ends human gating of datacap issuance. Datacap can then be transferred to the verified registry to create a datacap allocation. Storage Providers can claim the allocation assigned to them and get the associated 10x deal multiplier when onboarding the data piece.
+After this change, the verified registry and datacap actors will be mostly obsolete. Allowing unlimited datacap does not create any complexity. Future protocol improvements could remove these actors and support direct specification of a multiplier when onboarding to further simplify the protocol and reduce cost.
```
jsoares, line 39:
```suggestion

After this change, the verified registry and datacap actors will be mostly obsolete. Allowing unlimited datacap does not create any complexity. Future protocol improvements could remove these actors and support direct specification of a multiplier when onboarding, in order to further simplify the protocol and reduce cost.
```

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Backwards Compatibility
+This proposal does not change any chain state or method interfaces so is backwards compatible to all clients. All existing workflows will continue to work, and this change is also compatible with direct onboarding. A network upgrade is required to deploy the new actor code.
```
jsoares, line 42:
```suggestion
This proposal does not change any chain state or method interfaces. It is, therefore, backwards compatible with all clients. All existing workflows will continue to work, and this change is also compatible with direct onboarding. A network upgrade is required to deploy the new actor code.
```

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+After this upgrade, Storage Providers (SPs) can engage with any deal, whether Fil+ or not by minting Datacap. Deals will still need the `verified` flag set to true, but clients will be able to freely do this because they can mint the required datacap.
```
jsoares, line 44:
```suggestion
After this upgrade, Storage Providers (SPs) can engage with any deal, whether Fil+ or not, by minting Datacap. Deals will still need the `verified` flag set to true, but clients will be able to do this freely because they can mint the required datacap.
```

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Test Cases
+Test that an account actor that is not a notary can mint datacap
+Test that a non-account actor can mint datacap
```
jsoares, line 48:
```suggestion
- Test that an account actor that is not a notary can mint datacap
- Test that a non-account actor can mint datacap
```

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+Substitution risk is also mitigated by the fact that the expiration of CC sectors is a well-established trend. Substitution may even help to retain some hardware that would otherwise completely drop off the network.
+
+In order to get 20% of QAP, an attacker who only has CC would need about 600 petabytes of RBP and about 40,000,000 FIL. To convert 600 PB of CC to 10x sectors via snap deals would take over 10,000 sealers working non-stop for a week. This is a worst-case analysis that assumes the attacker is the only one to substitute CC for 10x sectors; in reality, this is extremely unlikely.
+
+The effects of this change on RBP, QAP and network security were also analyzed by CEL [in this report](https://hackmd.io/zp-40inqSny5HMqmTSOKxw?view)
+
+
+## Incentive Considerations
+The proposed change removes Fil+'s incentive for the onboarding of non-random data. As a result, it's anticipated that the onboarding rate of non-random data may decrease. It is still uncertain how much of the data currently incentivized by Fil+ is fake.
```
jsoares, line 64:
```suggestion
The proposed change removes Fil+'s incentive for the onboarding of non-random data. As a result, it's anticipated that the onboarding rate of non-random data may decrease. It remains unclear how much of the data currently incentivized by Fil+ is not in compliance with the program's goals.
```

_Fake_, like _useful_, remains vaguely and contentiously defined.

FIPS/fip-0078.md
```diff
@@ -0,0 +1,96 @@
...
+
+Because all sectors will have the same multiplier, we expect miners to charge more for storing user data. So users whose data is currently stored via Fil+ deals will probably have to pay more for storage after the sector containing their data expires. We suggest existing users discuss pricing changes with their SPs or the intermediary who handled deal-making for them. It is unlikely that user data will be dumped before sector expiry since a miner would have to pay a termination fee and re-seal a new sector. 
+
+Data onboarded after the transition to free multipliers will not face the risk of price hikes or dumping.  
+
+It is hard to estimate how much user demand Filecoin will lose in the short-run, (both demand for data already stored on the network, and data that potentially could have been onboarded in the future). There are two reasons for this: first, we don't know how much of the data stored via Fil+ deals has an actual user behind it, second, because Fil+ distorted price discovery, we don't know what users are willing to pay for Filecoin storage and what miners are willing to accept.
```
jsoares, line 74:
```suggestion
It is hard to estimate how much user demand Filecoin will lose in the short run, (both demand for data already stored on the network and for data that potentially could have been onboarded in the future). There are two reasons for this: first, we don't know how much of the data stored via Fil+ deals has an actual user behind it; second, because Fil+ distorted price discovery, we don't know what users are willing to pay for Filecoin storage and what miners are willing to accept.  
```

jsoares (approved):
I second @anorth. Leaving some minor editorial suggestions for the sake of clarity and readability.

anorth:
The merge button on GitHub is disabled because @momack2 and @kaitlin-beegle used "request changes" for their reviews. GitHub workflow requires them to re-review to lift the request for change (which makes sense and is useful in many code review contexts). I believe a some reasonable changes have been made in response, at least enough to make this PR reasonable as a draft FIP. I will "dismiss review" and then merge in a couple of days. Of course, more improvements can be requested in the associated discussion thread if needed before moving to Last Call.

kaitlin-beegle (approved):

kaitlin-beegle:
> The merge button on GitHub is disabled because @momack2 and @kaitlin-beegle used "request changes" for their reviews. GitHub workflow requires them to re-review to lift the request for change (which makes sense and is useful in many code review contexts). I believe a some reasonable changes have been made in response, at least enough to make this PR reasonable as a draft FIP. I will "dismiss review" and then merge in a couple of days. Of course, more improvements can be requested in the associated discussion thread if needed before moving to Last Call.

Thanks for the flag, Alex. It was not my intention to hold this up, and I agree that substantial improvements have been made.  I've approved this for merging.

PR #804, opened 2023-08-23
First draft of FIP-0076: Direct data onboarding

anorth:
Adds new `ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34) methods to the miner actor
which support committing to data and claiming verified allocations without requiring a built-in market deal.
Adds a sector→deal mapping in the built-in market actor and deprecates the storage of deal IDs in the miner actor.
Removes the legacy `PreCommitSector` (method 6) and `PreCommitSectorBatch` (method 25)
in favour of the existing `PreCommitSectorBatch2` (method 28) which requires
the sector unsealed CID to be specified while pre-committing.
Removes the legacy and unused `ProveReplicaUpdates2` (method 29) (giving that name to the new method 34).

Existing onboarding flows that use the built-in market actor remain fully supported, but optional.

Discussion: https://github.com/filecoin-project/FIPs/discussions/730

### DIFF
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -415,17 +415,19 @@ may not be extended to other actors in the future (because termination can be in
 The existing `ProveCommitSector` (method 7) and `ProveCommitAggregate` (method 26) remain,
 and continue to activate deals in the built-in market actor and claim verified allocations using
 the current algorithm.
+These methods remain available as a grace period for participants to migrate their workflows.
 These methods no longer writes deal IDs into per-sector chain state.
 
-These methods cannot be used for direct data onboarding and wil always incur the cost of a built-in market deal.
+These methods cannot be used for direct data onboarding and will always incur the cost of a built-in market deal.
 
 #### ProveReplicaUpdates
 
 The existing `ProveReplicaUpdates` (method 27) remains and continues to activate deals and claim verified allocations
 using the current algorithm.
+This method remains available as a grace period for participants to migrate their workflows.
 This method no longer writes deal IDs into per-sector chain state.
 
-This method cannot be used for direct data onboarding and wil always incur the cost of a built-in market deal.
+This method cannot be used for direct data onboarding and will always incur the cost of a built-in market deal.
 
 #### Deprecation of legacy methods
 
@@ -590,7 +592,7 @@ However, the specification of piece manifests renders this unnecessary.
 Method 29 is not used by the widely used "Lotus miner" software and has never been invoked on mainnet.
 It is deprecated in order to reduce the number of actively-supported methods.
 The name `ProveReplicaUpdates2` is taken for the new method 34 in order to align the `2` suffix with the
-similarly-behaving `ProveReplicaUpdates2`.
+similarly-behaving `PreCommitSectorBatch2`.
 
 ### Sector deal information moved to market actor
 
@@ -656,7 +658,7 @@ After this proposal:
 - _verified_ data _with_ on-chain payments will be the most expensive (but still a little cheaper than current).
 
 The variations in cost account for the variations in computational work necessary to provide incremental functionality.
-These variation might increase the relative profitability of onboarding data that is unverified
+These variations might increase the relative profitability of onboarding data that is unverified
 and/or requires no on-chain payments, and hence the uptake of these.
 When `SectorContentChanged` notifications support user-programmed actors, they may be able to implement
 on-chain payments more efficiently than the built-in storage market actor.
@@ -762,7 +764,7 @@ but requires the storage provider to maintain this metadata during the onboardin
 
 ### Source of truth for data commitments and verified claims
 
-This proposal decouples the both ability to commit to data, and the allocation of DataCap,
+This proposal decouples both the ability to commit to data, and the allocation of DataCap,
 from on-chain deals with the built-in market actor.
 The motivation for this is to avoid the high costs of the built-in market actor when they are not necessary.
 
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -1,5 +1,5 @@
 ---
-fip: "<to be assigned>"
+fip: "0076"
 title: Direct data onboarding
 author: "Alex North (@anorth), @zenground0"
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/730
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -540,6 +540,28 @@ struct OnMinerSectorsTerminateParams {
 The sector is removed from the `ProviderSectors` mapping, if present.
 Any deals mapped to that sector are marked as terminated, and subsequent processing deferred to cron.
 
+### Migration
+
+The built-in market actor's `ProviderSectors` mapping is initialised from the existing deal state
+and miner actor state per-sector deal IDs. 
+
+- For each deal state object in the market actor state that has a terminated epoch set to `-1`:
+  - find the corresponding deal proposal object and extract the provider's actor ID;
+  - in the provider's miner state, find the ID of the sector with the corresponding deal ID in sector metadata;
+    - if such a sector cannot be found, assert that the deal's end epoch has passed and use sector ID `0` [1];
+  - set the new deal state object's sector number to the sector ID found;
+  - add the deal ID to the `ProviderSectors` mapping for the provider's actor ID and sector number.
+- For each deal state object in the market actor state that has a terminated epoch set to any other value:
+  - set the deal state object's sector number to `0`. 
+
+[1] It may be impossible to find the sector for a deal that has completed successfully 
+but not yet been cleaned up in market actor state, 
+if the corresponding sector has since expired and been compacted out of state.
+
+The result includes a value in the `ProviderSectors` mapping for each activated and not yet terminated or expired deal.
+The built-in market actor's implementation of deal expiration clean-up must be robust to the provider sector mapping
+missing a value for a terminated or expired deal.
+
 ## Design Rationale
 
 This proposal makes deals optional in order to reduce the cost of data onboarding when deal payments are not required.
@@ -760,6 +782,24 @@ The built-in market actor may have a partial view of both piece commitments and 
 but this is no longer a reliable or complete source of information about either.
 Clients, explorers and analytic tools should inspect the source of truth for each type of information,
 and avoid reliance on the presence of deals with built-in market actor.
+
+### Reduced scope of on-chain visibility into terms for sector data
+
+This proposal has potential to reduce the scope of on-chain visibility into the data committed to sectors.
+At present, _all_ sector data is represented by a record stored on-chain in the built-in market actor,
+which includes a client identifier and deal terms,
+and all verified data is represented by a similar record stored on-chain in the verified registry actor.
+After this proposal, it becomes possible for data to be committed to a sector without a corresponding record
+in either the built-in market actor or the verified registry actor.
+
+The commitment to such data still appears in the chain message history, visible from outside the chain,
+but is not verifiable from an actor (unless a future change adds visibility into change history).
+The distinction between provably-empty sectors and those containing some data remains in chain state,
+but no client or terms for such data is necessarily present.
+
+Note that even today there is no guarantee that deal metadata is meaningful.
+A storage provider can commit arbitrary data through the built-in market without a meaningful client or terms.
+This proposal makes that behaviour cheaper by not storing the meaningless metadata on chain.
 
 ## Implementation
 
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -135,11 +135,12 @@ Per-sector deal IDs may be removed from state in a future migration.
 The miner actor exports new methods for sector activation:
 `ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34).
 These new methods both support either batched or aggregated proofs, though aggregated proofs for
-replica updates are not yet possible.
+replica updates are not yet possible (see https://github.com/filecoin-project/FIPs/discussions/752).
 
 ##### ProveCommitSectors2
 
-This method rejects sectors with deal IDs specified at pre-commit.
+This method rejects sectors with deal IDs specified at pre-commit to avoid operator confusion
+between the two distinct ways of specifying deals (one of which would have to be ignored).
 Such sectors must be activated with the existing `ProveCommitSector` or `ProveCommitAggregate` methods.
 This method does not fetch deal information from the built-in market actor.
 Instead, the pieces of data that comprise a sector are declared by the sender as a _manifest_.
@@ -239,6 +240,13 @@ struct FailCode {
     Code: ExitCode,
 }
 ```
+
+##### FVM syscalls
+
+The gas cost for the `batch_verify_seals` syscall is changed to 42M gas per proof.
+This value was previously a placeholder because the syscall was only ever invoked from 
+call paths originating in the cron actor, which has unlimited gas budget.
+The new value has been determined empirically following a similar scheme to other gas prices.
 
 ##### ProveReplicaUpdates2
 
@@ -366,7 +374,7 @@ Each data piece can have a single verified claim and multiple notifications.
 Any of these items might fail, but only limited ability to handle individual failures in a group is practical.
 
 **The activation of each sector is independent**.
-A failed sctor activation will not cause a top-level method to abort, unless all activations fail.
+A failed sector activation will not cause a top-level method to abort, unless all activations fail.
 An SP can specify `RequireActivationSuccess=true` to instead require every sector activation to succeed,
 aborting the operation if one fails.
 
@@ -380,6 +388,8 @@ they can instead resubmit the failed sector with only the remaining valid claims
 Notifications are sent strictly _after_ activation, and only for successfully activated sectors.
 If a notification call returns a non-zero exit code, sector activation will be committed regardless;
 notifications are sent on a _best-effort_ basis.
+If a notification to the built-in storage market actor fails, the associated deal will not be started,
+but the sector can activate anyway.
 An SP can specify `RequireNotificationSuccess=true` to instead require every notification to succeed,
 aborting the entire operation if one fails.
 There is no ability to roll back the activation of only some sectors in response to failed notifications.
@@ -750,9 +760,6 @@ The built-in market actor may have a partial view of both piece commitments and 
 but this is no longer a reliable or complete source of information about either.
 Clients, explorers and analytic tools should inspect the source of truth for each type of information,
 and avoid reliance on the presence of deals with built-in market actor.
-
-## TODO
-- Specify gas cost of synchronous PoRep verification and any consequent change to batch balancer parameters.
 
 ## Implementation
 
--- a/FIPS/fip-direct-onboarding.md
+++ b/FIPS/fip-direct-onboarding.md
@@ -11,12 +11,13 @@ created: 2023-08-24
 
 ## Simple Summary
 
-Adds new prove-commit and replica-update methods to the miner actor which support committing to data
-and claiming verified allocations without requiring a deal with the built-in market actor.
+Adds new `ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34) methods to the miner actor
+which support committing to data and claiming verified allocations without requiring a built-in market deal.
 Adds a sector→deal mapping in the built-in market actor and deprecates the storage of deal IDs in the miner actor.
-Removes the legacy Miner::PreCommitSector (method 6) and Miner::PreCommitSectorBatch (method 25),
-in favour of the existing Miner::PreCommitSectorBatch2 (method 28) which requires
+Removes the legacy `PreCommitSector` (method 6) and `PreCommitSectorBatch` (method 25)
+in favour of the existing `PreCommitSectorBatch2` (method 28) which requires
 the sector unsealed CID to be specified while pre-committing.
+Removes the legacy and unused `ProveReplicaUpdates2` (method 29) (giving that name to the new method 34).
 
 Existing onboarding flows that use the built-in market actor remain fully supported, but optional.
 
@@ -35,7 +36,8 @@ The verified registry actor already records deal-like information describing Dat
 This proposal adds new onboarding methods which support direct commitment of data into sectors,
 without necessary reference to any deal.
 This provides gas-cheap data onboarding for many use cases.
-It also introduces a new scheme for deal activation which can support deals brokered by user-programmed smart contracts.
+It also introduces a new scheme for deal activation which can support deals (and other data-application transactions)
+brokered by user-programmed smart contracts.
 This new scheme is initially limited to the built-in market actor, but can be extended to other actors in the future.
 
 ## Change Motivation
@@ -65,7 +67,7 @@ It cannot support the expansion in usage or functionality that network participa
 This proposal aims to:
 
 - Support data onboarding, including Fil+ verified data, with no smart-contract intermediary;
-- Provide a new scheme for data activation which can support
+- Provide a new scheme for data activation which can later support
   user-programmed smart contracts to function as data storage applications (including as markets);
 - Continue supporting existing onboarding methods to give participants time to migrate their workflows.
 
@@ -103,14 +105,11 @@ Existing onboarding methods are retained and remain fully supported, but optiona
 ### Storage miner actor
 
 No changes to miner state schemas are necessary,
-but the deal IDs stored on with each sector's metadata are redundant.
-The sector→deal association will be stored in the built-in market actor instead, and
-any information that remains in the miner actor state will be incomplete.
+but the deal IDs stored on with each sector's metadata are no longer used.
+Any deal-related information that remains in the miner actor state will be incomplete so should not be used.
+A sector→deal association will be stored in the built-in market actor instead (see below).
 
-The interpretation of some fields changes slightly.
-No migration is necessary.
-
-TODO: remove PreCommitSector and PreCommitSectorBatch methods
+The interpretation of some per-sector metadata fields changes slightly.
 
 #### State
 
@@ -124,7 +123,7 @@ Pre-commit deal IDs may be removed in a future FIP when deprecating the old meth
 ##### SectorOnChainInfo
 
 The per-sector `DealIDs` field is deprecated and to be ignored.
-The miner actor will not write to it.
+The miner actor will not write new deals to it.
 
 The `DealWeight` field is interpreted to carry the weight of _any non-zero, unverified data_ in a sector,
 not only the weight of deals made via the built-in market actor.
@@ -136,29 +135,33 @@ Per-sector deal IDs may be removed from state in a future migration.
 #### Sector activation
 
 The miner actor exports new methods for sector activation:
-`ProveCommitSectors2` (method 33) and `ProveReplicaUpdates3` (method 34).
+`ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34).
 These new methods both support either batched or aggregated proofs, though aggregated proofs for
 replica updates are not yet possible.
 
 ##### ProveCommitSectors2
 
-This method rejects sectors that deal IDs specified at pre-commit.
-Such sectors must be activated with the existing `ProveCommitSector` or `ProveCommitAggegate` methods.
+This method rejects sectors with deal IDs specified at pre-commit.
+Such sectors must be activated with the existing `ProveCommitSector` or `ProveCommitAggregate` methods.
 This method does not fetch deal information from the built-in market actor.
-Instead, the pieces of data that comprise a sector are declared as a _manifest_.
+Instead, the pieces of data that comprise a sector are declared by the sender as a _manifest_.
 
 Each piece in a manifest _may_ declare a verified data allocation ID that it satisfies.
 The method will attempt to claim that allocation directly from the verified registry actor and, if successful,
 calculate quality-adjusted power according to the piece’s size.
-If unsuccessful, the sector will not be activated.
+If unsuccessful, the containing sector will not be activated.
 
-Each piece in a manifest _may_ specify the address of an actor and a notification payload,
+Each piece in a manifest _may_ specify zero or more addresses of an actor and notification payloads,
 to be notified synchronously when the sector is activated.
 After successful activation, the method will invoke the `SectorContentChanged` method
-on the target actor with the piece CID and payload.
-This functions as a notification that the piece has been committed, e.g. to a marketplace.
+on the target actor(s) with the piece CID and payload.
+This functions as a synchronous notification that the piece has been committed, e.g. to a marketplace.
 `SectorContentChanged` will be invoked just once per recipient actor, with a message body
 describing all pieces to be notified to that actor.
+
+The miner actor will reject attempts to notify any actor other than the built-in storage market actor (`f05`).
+This restriction may be lifted in a future FIP once the security considerations of calls into
+untrusted code are better understood and mitigated.
 
 In execution of `ProveCommitSectors2`, the miner actor:
 
@@ -170,7 +173,7 @@ In execution of `ProveCommitSectors2`, the miner actor:
 6. notifies any actors specified in the piece manifests.
 
 ```
-struct ProveCommit2Params {
+struct ProveCommitSectors2Params {
     // Activation manifest for each sector being proven.
     SectorActivations: []SectorActivationManifest,
     // Proofs for each sector, parallel to activation manifests.
@@ -219,40 +222,30 @@ struct DataActivationNotification {
     Payload: []byte,
 }
 
-struct ProveCommit2Return {
-    // Sector activation results, parallel to input sector activation manifests.
-    Sectors: []SectorActivationReturn,
+// Note transparent serialization of single-element struct.
+struct ProveCommitSectors2Return {
+    // Success/fail of each input in order.
+    ActivationResults: BatchReturn
 }
 
-struct SectorActivationReturn {
-    // Whether the sector was activated.
-    Activated: bool,
-    // Power of the activated sector (or zero).
-    Power: StoragePower,
-    // Piece activation results, parallel to input piece activation manifests.
-    Pieces: []]PieceActivationReturn,
+// This BatchReturn type is an existing structure used for batched results in other built-in actors.
+struct BatchReturn {
+    // Total successes in batch
+    SuccessCount: u32,
+    // Failure code and index for each failure in batch
+    FailCodes: []FailCode,
 }
 
-struct PieceActivationReturn {
-    // Whether a verified allocation was successfully claimed by the piece.
-    Claimed: bool,
-    // Results from notifications of piece activation, parallel to input notification requests.
-    Notifications: []DataActivationNotificationReturn,
-}
-
-struct DataActivationNotificationReturn {
-    // Exit code from the notified actor.
+struct FailCode {
+    Idx: u32,
     Code: ExitCode,
-    // Return value from the notified actor.
-    Data: RawBytes,
 }
-
-
 ```
 
-##### ProveReplicaUpdates3
+##### ProveReplicaUpdates2
 
-As for `ProveCommitSectors2`, this method does not fetch deal information from the built-in market actor.
+The name `ProveReplicaUpdates2` is taken from the existing and now-deprecated method 29.
+Like `ProveCommitSectors2`, this method does not fetch deal information from the built-in market actor.
 Instead, the pieces of data that comprise the updated sector content are declared as a _manifest_.
 This manifest is similar to the manifest for `ProveCommitSectors2`, with differences being to specify
 the existing sector state to be updated rather than a new one.
@@ -260,7 +253,7 @@ the existing sector state to be updated rather than a new one.
 The specification and semantics of pieces, including verified allocation IDs and notifications,
 is the same as for `ProveCommitSectors2`.
 
-In execution of `ProveReplicaUpdates3`, the miner actor:
+In execution of `ProveReplicaUpdates2`, the miner actor:
 
 1. validates that the existing sectors are eligible for update;
 2. computes an unsealed CID from the piece manifests;
@@ -271,7 +264,7 @@ In execution of `ProveReplicaUpdates3`, the miner actor:
 6. notifies any actors specified in the piece manifests.
 
 ```
-struct ProveReplicaUpdates3Params {
+struct ProveReplicaUpdates2Params {
     SectorUpdates: []SectorUpdateManifest,
     // Proofs for each sector, parallel to activation manifests.
     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
@@ -301,7 +294,7 @@ pub struct SectorUpdateManifest {
     Pieces: []PieceActivationManifest,
 }
 
-type ProveReplicaUpdates3Return = ProveCommit2Return;
+type ProveReplicaUpdates2Return = ProveCommit2Return;
 ```
 
 #### SectorContentChanged
@@ -309,12 +302,8 @@ type ProveReplicaUpdates3Return = ProveCommit2Return;
 When a piece manifest specifies one or notification receivers,
 the storage miner invokes these receivers after activating the sector or replica update.
 The receiving actor must accept the `SectorContentChanged` method number and parameter schema.
-`SectorContentChanged` is an FRC-0046 method, intended to be implemented by other actors.
+`SectorContentChanged` is an FRC-0046 method, intended to be implemented by user-programmed actors.
 The miner actor will invoke each receiver address only once, with a batch of notification payloads.
-
-The miner actor will reject attempts to notify any actor other than the built-in storage market actor (`f05`).
-This restriction may be lifted in a future FIP once the security considerations of calls into
-untrusted code are better understood and mitigated.
 
 ```
 // Notification of change committed to one or more sectors.
@@ -323,7 +312,7 @@ untrusted code are better understood and mitigated.
 // Note transparent serialization of single-element struct.
 struct SectorContentChangedParams {
     // Distinct sectors with changed content.
-    Sectors: []]SectorChanges,
+    Sectors: []SectorChanges,
 }
 
 // Description of changes to one sector's content.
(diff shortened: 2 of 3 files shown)


FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+
+## Product Considerations
+<!--All FIPs must contain a section that discusses the product implications/considerations relative to the proposed change. Include information that might be important for product discussion. A discussion on how the proposed change will enable better storage-related goods and services to be developed on Filecoin. FIP submissions missing the "Product Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Product Considerations discussion deemed sufficient by the reviewers.-->
+
+## Implementation
+<!--The implementations must be completed before any core FIP is given status "Final", but it need not be completed before the FIP is accepted. While there is merit to the approach of reaching consensus on the specification and rationale before writing code, the principle of "rough consensus and running code" is still useful when it comes to resolving many discussions of API details.-->
+
+## TODO
+- Determine whether we will migrate to remove SectorOnChainInfo.Deals, or merely deprecate its use.
+- Determine whether to deprecate ProveReplicaUpdates2 (and claim its name for PRU3)
```
jennijuju, line 529: Can we add a todo for "gas cost benchmarking and impact"?

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+created: 2023-08-24
+---
+
+## Simple Summary
+
+Adds new prove-commit and replica-update methods to the miner actor which support committing to data
+and claiming verified allocations without requiring a deal with the built-in market actor.
+Adds a sector→deal mapping in the built-in market actor and deprecates the storage of deal IDs in the miner actor.
+Removes the legacy Miner::PreCommitSector (method 6) and Miner::PreCommitSectorBatch (method 25),
+in favour of the existing Miner::PreCommitSectorBatch2 (method 28) which requires
```
ZenGround0, line 18: Worth mentioning we're replicating the method formerly known as prove replica updates 2 (29) and giving that name to the replica update method being added (34)

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+
+### Storage miner actor
+
+No changes to miner state schemas are necessary,
+but the deal IDs stored on with each sector's metadata are redundant.
+The sector→deal association will be stored in the built-in market actor instead, and
+any information that remains in the miner actor state will be incomplete.
+
+The interpretation of some fields changes slightly.
+No migration is necessary.
```
ZenGround0, line 111: nit: I would delete this line.  In terms of writing new miner state this is true but all sector infos must be read to make the sector->deal index.  So this FIP requires paying for the most expensive part of any migration.

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+not only the weight of deals made via the built-in market actor.
+Similarly, the `VerifiedDealWeight` field carries the weight of _any verified data_, regardless of
+whether a "deal" was made via the built-in market actor.
+
+Per-sector deal IDs may be removed from state in a future migration.
+
+#### Sector activation
+
+The miner actor exports new methods for sector activation:
+`ProveCommitSectors2` (method 33) and `ProveReplicaUpdates3` (method 34).
```
ZenGround0, line 139: Now ProveReplicaUpdates2

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+// Note transparent serialization of single-element struct.
+struct SectorContentChangedReturn {
+    // A result for each sector that was notified, in the same order.
+    Sectors: []]SectorReturn,
+}
+
+// Note transparent serialization of single-element struct.
+struct SectorReturn {
+    // A result for each piece for the sector that was notified, in the same order.
+    Added: []]PieceReturn,
```
ZenGround0, line 364: typo: []PieceReturn

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+}
+
+struct PieceReturn {
```
ZenGround0, line 367: We've changed this to

```
pub struct PieceReturn {
    // Indicates whether the receiver accepted the notification.
    // The caller is free to ignore this, but may chose to abort and roll back.
    pub accepted: bool,
}
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+}
+
+struct DataActivationNotification {
+    // Actor to be notified.
+    Address: Address,
+    // Data to send in the notification.
+    Payload: []byte,
+}
+
+struct ProveCommit2Return {
```
ZenGround0, line 222: We've changed this to 
```
pub struct ProveCommitSectors2Return {
    pub activation_results: BatchReturn,
}
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+An SP can specify `RequireNotificationSuccess=true` to instead require every notification to succeed,
+aborting the entire operation if one fails.
+There is no ability to roll back the activation of only some sectors in response to failed notifications.
+
+This is a change in sequencing and semantics from the existing onboarding methods flow,
+where deal activation happens first and failure leads to an individual sector failing,
+and verified claims happen second and must all succeed.
+
+#### ProveCommitSector
+
```
ZenGround0, line 404: nit: good to mention ProveCommitAggregate here too

ZenGround0: Probably also ProveReplicaUpdate

ZenGround0 (approved):
Probably planning to cover this in incomplete Design Rationale but make sure there is some discussion of the internal changes to sector termination and its role in motivating sector number => deal ids index in market actor

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,532 @@
...
+    // Proofs for each sector, parallel to activation manifests.
+    // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+    SectorProofs: [][]byte,
+    // Aggregate proof for all sectors.
+    // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+    AggregateProof: []byte,
+    // The proof type for all sector update proofs, individually or before aggregation.
+    UpdateProofsType: RegisteredUpdateProof,
+    // The proof type for the aggregate proof (ignored if no aggregate proof).
+    AggregateProofType: RegisteredAggregateProof,
```
anorth, line 285: @cryptonemo can you verify that this is the correct information that will need to be provided by an SP to submit an aggregated replica update proof?

jennijuju:
@anorth  -  should the `./idea` be committed?

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+
+## Simple Summary
+
+Adds new `ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34) methods to the miner actor
+which support committing to data and claiming verified allocations without requiring a built-in market deal.
+Adds a sector→deal mapping in the built-in market actor and deprecates the storage of deal IDs in the miner actor.
+Removes the legacy `PreCommitSector` (method 6) and `PreCommitSectorBatch` (method 25)
+in favour of the existing `PreCommitSectorBatch2` (method 28) which requires
+the sector unsealed CID to be specified while pre-committing.
+Removes the legacy and unused `ProveReplicaUpdates2` (method 29) (giving that name to the new method 34).
```
jennijuju, line 20: Peer: Ideally I would prefer a new method number to be assigned, and deprecate the old message/drop the code an upgrade after. Imagine a user sends legacy PRU2 message right before the upgrade, the message will fail upon block inclusion if it happens right after the network upgrade. It is a failure we could avoid for users with minimum cost imho.

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/730
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-24
+---
+
+## Simple Summary
+
+Adds new `ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34) methods to the miner actor
```
jennijuju, line 14:
```suggestion
Add new `ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34) methods to the miner actor
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+which support committing to data and claiming verified allocations without requiring a built-in market deal.
```
jennijuju, line 15:
```suggestion
which support committing data to sectors and claiming datacap allocations without requiring a built-in market (f05) deal.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+Adds a sector→deal mapping in the built-in market actor and deprecates the storage of deal IDs in the miner actor.
```
jennijuju, line 16:
```suggestion
Add a provider->sector→deal mapping in the built-in market actor and deprecated the storage of deal IDs in the sector onchain info of the miner actor.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+  The cost of transferring DataCap is still incurred, but is only a fraction of the cost of publishing a deal.
+  The cost of sector pre-commitment is reduced slightly by avoiding interaction with the built-in market actor
+  The cost of sector activation is reduced by avoiding the cost of deal activation.
+- For direct onboarding with a verified claim and a built-in storage market deal, the cost of `PreCommitSectorBatch`
+  is reduced slightly by avoiding interaction with the built-in market actor.
+
+As an illustrative example, an analysis of the gas cost of selected onboarding messages
+with a batch size of 16 in epoch 2844239 indicates:
+
+| Message              | Observed w/ Fil+ | Direct with Fil+ | Observed ex Fil+ | Direct ex Fil+   |
```
jennijuju, line 653: is ex FIL+ for both CC sectors and sectors only has regular deals?

FIPS/fip-direct-onboarding.md
cryptonemo: This looks correct to me, thanks!  Still working on that FIP by the way ... :hourglass_flowing_sand:

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+}
+```
+
+#### BatchActivateDeals
+
+When an SP activates a piece with the existing activation methods,
+deals are activated with the `BatchActivateDeals` (method 6).
+This method returns the necessary verified allocation IDs to the miner actor.
+The `BatchActivateDeals` method parameters are expanded to include the sector ID for each deal.
+When activating a deal, the market actors writes it into the new `ProviderSectors` mapping.
```
arajasek, line 479:
```suggestion
When activating a deal, the market actor writes it into the new `ProviderSectors` mapping.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+
+## Security Considerations
+
+This proposal has no impact on consensus or blockchain security.
+
+## Incentive Considerations
+
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
+
+- gas impacts
```
arajasek, line 609: TODO

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+
+## Implementation
+
+Implementation of the protocol changes is being developed in the `integration/direct-onboarding`
+[branch of the built-in actors](https://github.com/filecoin-project/builtin-actors/tree/integration/direct-onboarding)
+repository.
+
+## TODO
+
+- Determine whether we will migrate to remove SectorOnChainInfo.Deals, or merely deprecate its use.
```
arajasek, line 735: Primary motivation to do so: we're migrating all sectors this upgrade anyway, now's a good time to cleanup.

Primary motivation not to do so: it's a change (albeit a small one) that will save very little state size. Is it worth it?

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+in favour of the existing `PreCommitSectorBatch2` (method 28) which requires
+the sector unsealed CID to be specified while pre-committing.
+Removes the legacy and unused `ProveReplicaUpdates2` (method 29) (giving that name to the new method 34).
+
+Existing onboarding flows that use the built-in market actor remain fully supported, but optional.
+
+## Abstract
+
+The only mechanism today by which storage providers (SPs) are permitted to commit data in sectors is
+to make a deal with the built-in storage market actor.
```
jennijuju, line 27:
```suggestion
to publish a deal made with data client to the built-in storage market actor.
```

FIPS/fip-direct-onboarding.md
anorth: See design rationale. It's just not going to happen - it has never been used.

FIPS/fip-direct-onboarding.md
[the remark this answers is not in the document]
anorth: Can you suggest some names? I've been unable to find something appropriately concise yet more descriptive of the difference.

anorth: I will decline these suggestions. I have used a consistent declarative rather than imperative posture and I don't think it detracts from understanding the content.

FIPS/fip-direct-onboarding.md
anorth: It's not for CC since they never interact with the market or verifreg anyway. This proposal will make no significant difference to CC gas cost.

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+- at pre-commit, an SP must specify a sector’s data commitment (unsealed CID),
+  but does not need to specify the structure of that data nor any deals or verified allocations.
+- at prove-commit or replica-update, an SP specifies the pieces of data comprising a sector,
+  and for each piece may nominate:
+    - a verified allocation to claim, and/or
+    - an actor to notify of the commitment (e.g. to activate a deal).
+
+The miner actor verifies that the pieces of data claimed by the SP correspond to the data commitment proven.
+The verified registry is involved only if verified data is being claimed.
+The built-in market actor is involved only if an on-chain-settled deal is required (i.e. if non-zero payment).
```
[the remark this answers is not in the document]
anorth: Changed to "need only be involved". An SP is not really using direct onboarding if they do a deal – I think direct _means_ no deal. But the new methods also support doing deals, and I can see how a reader might think direct == new methods rather than direct == no deal. Such definition isn't relevant to the protocol though.

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+Removes the legacy and unused `ProveReplicaUpdates2` (method 29) (giving that name to the new method 34).
+
+Existing onboarding flows that use the built-in market actor remain fully supported, but optional.
+
+## Abstract
+
+The only mechanism today by which storage providers (SPs) are permitted to commit data in sectors is
+to make a deal with the built-in storage market actor.
+The built-in market actor is expensive in gas, and offers only basic functionality.
+Requiring its use raises costs and limits the utility and value of sectors,
```
jennijuju, line 29:
```suggestion
Requiring its use raises unnecessary costs for those who don't use f05 market functionalities and limits the utility and value of sectors,
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+and the applications that can be built on Filecoin.
+
+This cost and inconvenience is unnecessary in the common cases of verified deals with no client payments,
+and storage arrangements that do not require any on-chain deal.
+The verified registry actor already records deal-like information describing DataCap allocation terms.
```
jennijuju, line 34: This sentence on it now makes sense, tho im not really sure what it tries to say within the context of this paragraph.  Could the follow be a bit clearer?
```suggestion
The other thing f05 deal offers is on-chain deal term, and for verified pieces, the verified registry actor already records deal-like information describing DataCap allocation terms.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+
+This proposal adds new onboarding methods which support direct commitment of data into sectors,
+without necessary reference to any deal.
+This provides gas-cheap data onboarding for many use cases.
+It also introduces a new scheme for deal activation which can support deals (and other data-application transactions)
```
jennijuju, line 39: do we want to keep use the term `deal`? o should we start to move towards something like  `data (piece) activation`?

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+brokered by user-programmed smart contracts.
```
jennijuju, line 40:
```suggestion
brokered by user-programmed actors.
```

FIPS/fip-direct-onboarding.md
jennijuju, line 39:
```suggestion
It also introduces a new scheme in storage miner actor for deal activation which can support deals (and other data-application transactions)
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+This new scheme is initially limited to the built-in market actor, but can be extended to other actors in the future.
+
+## Change Motivation
+
+The built-in market actor is expensive to use but provides very little utility to majority data onboarding use cases.
+A vast majority of deals today are simple FIL+ verified deals with no client payment.
```
jennijuju, line 46:
```suggestion
A vast majority of deals today are simple Fil+ verified deals with no client payment.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+Since the verified registry actor already records all relevant information about DataCap allocation terms
+(the client and provider, piece commitment, and duration),
+the built-in market actor's replication of this data is unnecessary.
+The built-in market plays no necessary role in allocating or accounting for QA power.
+Publishing deals is the largest on-chain cost of QA power onboarding,
```
jennijuju, line 51: isn't PSD also costy for regular deals too?

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+both verified and unverified data could be onboarded at a significant reduction in gas cost.
+
+The built-in market actor is also very limited.
+It supports only a single, simple deal schema and policies, and does not support a range of desirable market features.
+User-programmed smart contracts are restricted in the functionality that they can provide by the necessity
+of using the built-in market actor as an intermediary for any data commitments.
+They are also restricted by the lack of any hooks into the built-in miner actor to be notified of sector data
+commitments.
+
+Requiring the built-in market actor raises costs and limits the utility and value of sectors,
```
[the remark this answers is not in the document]
jennijuju: ➕

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+and the applications that can be built on Filecoin.
+It cannot support the expansion in usage or functionality that network participants hope for.
+
+This proposal aims to:
+
+- Support data onboarding, including Fil+ verified data, with no smart-contract intermediary;
```
jennijuju, line 69:
```suggestion
- Support data onboarding, including Fil+ verified data, with no built-in storage market or user deployed actor intermediary;
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+- Provide a new scheme for data activation which can later support
```
jennijuju, line 70:
```suggestion
- Provide a new scheme for data activation triggered by storage miner actors which can later support
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+  user-programmed smart contracts to function as data storage applications (including as markets);
+- Continue supporting existing onboarding methods to give participants time to migrate their workflows.
```
jennijuju, line 72:
```suggestion
- Continue supporting existing onboarding methods to keep existing f05 storage markets flow and functionality remain supported (for clients who would like to have onchain payments) and give participants time to migrate their workflows.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+
+## Specification
+
+### Overview
+
+Direct data onboarding comprises changes to the built-in miner and market actors
+to provide new onboarding methods that do not require a built-in market deal.
+These new methods accept additional parameters from the storage provider which specify
+the pieces of data activated in the sector, any verified allocations being claimed,
+and the address of any actors to notify about the successful activation.
```
jennijuju, line 82:
```suggestion
and the address of any actors to notify about the successful data storage activation.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+Deals are possible, but not necessary.
```
jennijuju, line 83:
```suggestion
Builtin storage market (f05) deals are possible, but not necessary.
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+
+In the direct onboarding flow:
+
+- _[for verified data only]_ a client makes a verified allocation directly with the verified registry
```
jennijuju, line 87:
```suggestion
- _[for verified data only]_ a verified client makes a datacap allocation for a specific piece of date to be stored with a designated storage provider directly with the verified registry
```

FIPS/fip-direct-onboarding.md
```diff
@@ -0,0 +1,740 @@
...
+  by transferring DataCap tokens to it (this is already possible today),
```
jennijuju, line 88: Peer: should we mention verified client can also delegate this operation a trusted third party, like SPs, or DP, where trusted third party may allocate datacap on behalf of the client by traversing datacap to verifreg via `transfer_from`?

FIPS/fip-direct-onboarding.md
anorth: already done below, this was stray. But now filling out other incentive considerations.

FIPS/fip-direct-onboarding.md
anorth: Have decided not to remove sector deals immediately.

FIPS/fip-direct-onboarding.md
anorth: The scheme right now is for the built-in market actor, so I think we should keep using that term here. The code doesn't though.

FIPS/fip-direct-onboarding.md
anorth: I get what you're saying here, but the FVM doesn't currently permit user-programmed actors, and 😭 may never. They use the term "smart contract" to refer to a thing running inside another (built-in) actor, like EVM. That is realistically the way this will first be usable.

FIPS/fip-direct-onboarding.md
anorth: They're forced to use collateral, so we can't say that yet.

FIPS/fip-direct-onboarding.md
anorth: Parties can have on-chain payments without using the old methods. The old methods remain only for transition.

FIPS/fip-direct-onboarding.md
anorth: If you don't think it's necessary to specify that a deal is for specific data and a designated provider also in the line below, I don't think it's necessary to specify here either. How allocations work is already specified in other FIPs and not changing here.

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+aborting the entire operation if one fails.
+There is no ability to roll back the activation of only some sectors in response to failed notifications.
+
+This is a change in sequencing and semantics from the existing onboarding methods flow,
+where deal activation happens first and failure leads to an individual sector failing,
+and verified claims happen second and must all succeed.
+
+#### Sector termination
+
+When a sector is terminated, the built-in market actor is notified of the sector termination
```
jennijuju, line 393: What would happen when a sector has datacap pieces onboarding without a f05 deal?

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+whether a "deal" was made via the built-in market actor.
+
+Per-sector deal IDs may be removed from state in a future migration.
+
+#### Sector activation
+
+The miner actor exports new methods for sector activation:
+`ProveCommitSectors2` (method 33) and `ProveReplicaUpdates2` (method 34).
+These new methods both support either batched or aggregated proofs, though aggregated proofs for
+replica updates are not yet possible.
```
jennijuju, line 138:
```suggestion
replica updates are not yet possible (@cryptonemo has proposed a way to enable this called SuperSnap, more details can be found [here](https://github.com/filecoin-project/FIPs/discussions/752)).
```

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+
+##### ProveCommitSectors2
+
+This method rejects sectors with deal IDs specified at pre-commit.
```
jennijuju, line 142: > This method rejects sectors with deal IDs specified at pre-commit.

could you please add a quick note on why we decided to take this approach? i.e: to avoid any special cases towards f05 market?

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+struct PieceReturn {
+    // Indicates whether the receiver accepted the notification.
+    // The caller (miner) is free to ignore this, but may chose to abort and roll back.
+    Accepted: bool,
+}
+```
+
+#### Failure handling
+
+Each batched sector activation or update comprises multiple sectors, each with multiple data pieces.
```
jennijuju, line 364: whats the reason behind this design choice - I understand that if a piece failed in the same sector its better to fail the whole sector for space considerations. However, it is not clear to me why sector A's failure should impact sector B

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+Each data piece can have a single verified claim and multiple notifications.
+Any of these items might fail, but only limited ability to handle individual failures in a group is practical.
+
+**The activation of each sector is independent**.
+A failed sctor activation will not cause a top-level method to abort, unless all activations fail.
+An SP can specify `RequireActivationSuccess=true` to instead require every sector activation to succeed,
```
jennijuju, line 370: When would an SP ever wants to do so (whats the use case)?

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+If a claim fails for one piece in a sector, no claims will be made for any piece in the sector
+and the sector will not be activated.
+The caller cannot choose to have activation proceed despite an invalid claim;
+they can instead resubmit the failed sector with only the remaining valid claims.
+
+**Sector activation does not include notifications**.
+Notifications are sent strictly _after_ activation, and only for successfully activated sectors.
+If a notification call returns a non-zero exit code, sector activation will be committed regardless;
+notifications are sent on a _best-effort_ basis.
+An SP can specify `RequireNotificationSuccess=true` to instead require every notification to succeed,
```
jennijuju, line 383: If >1 notification was specified to be sent and 1 or more failed, will all notification fail or..? how does it work? 

Peer: if the notification causes the message to fail due to ErrOutOfGas or such, I'd guess the sector activation will be reverted too? or if the actor detects the gas usage will just exit the notify loop so to ensure the message can still be successfully sent for sector activation? 

Peer: I have some design concern here for implementations in the future, it is possible some use actors will have expensive notification functions (i.e: lead to message to fail due to error, and ideally implementations should be able to simulate them before committing  to the piece onboarding and before adding the piece to the sectors.  (cc @nonsense @dirkmc

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+aborting the entire operation if one fails.
+There is no ability to roll back the activation of only some sectors in response to failed notifications.
+
+This is a change in sequencing and semantics from the existing onboarding methods flow,
+where deal activation happens first and failure leads to an individual sector failing,
+and verified claims happen second and must all succeed.
```
jennijuju, line 389: could you please specify if a sector contains both f05 deal pieces and non f05 pieces, when notification failed for f05 pieces yet the sector is successfully activated - what would the deal `pub activated: ChainEpoch,` value be?
 and how would `sector_start_epoch`  & `last_updated_epoch` be updated accordingly? (if it is specified in the later session, please kindly send a premark link, thanks.

FIPS/fip-0076.md
```diff
@@ -0,0 +1,765 @@
...
+- The source of truth for verified data claims is the verified registry actor state.
+- The built-in market actor state is a source of truth only for on-chain deal accounting.
+
+The built-in market actor may have a partial view of both piece commitments and verified claims,
+but this is no longer a reliable or complete source of information about either.
+Clients, explorers and analytic tools should inspect the source of truth for each type of information,
+and avoid reliance on the presence of deals with built-in market actor.
+
+## TODO
+- Specify gas cost of synchronous PoRep verification and any consequent change to batch balancer parameters.
```
arajasek, line 755: @anorth If you have no objection, the first TODO here can be resolved with a gas cost of 42M. See [here](https://github.com/filecoin-project/ref-fvm/issues/1852#issuecomment-1714320502) for more.

FIPS/fip-0076.md
anorth: > the built-in market actor is notified of the sector termination if the sector has non-zero deal weight or verified deal weight

Therefore it will be notified. From the scheme of the data we keep, the miner can't tell if verified deal weight is _also_ deal weight, or isn't. We could change this in the future, but would need to migrate things (adding verified deal weight to deal weight, so they will no longer sum to total weight).

FIPS/fip-0076.md
anorth: Because they expect it to work and want a hard abort if it doesn't. No I don't think I can specify a concrete use case, allowing atomic batches (when not _requiring_ them) is just our standard approach.

FIPS/fip-0076.md
anorth: If >=1 notification was specified then a failure will result in "aborting the entire operation". Aborting the entire operation means no sectors are activated. No there is no handling for out-of-gas or similar. That's why notifications are restricted to the built-in market actor to start. Your concerns about notifying untrusted code are valid, but there's no attempt to solve them here.

FIPS/fip-0076.md
anorth: If the notification fails _the deal is not activated_. So there is no activated epoch, sector_start_epoch, or last_updated_epoch.

FIPS/fip-direct-onboarding.md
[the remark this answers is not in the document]
arajasek: Yeah, I didn't have any good ideas either :(

FIPS/fip-0076.md
```diff
@@ -0,0 +1,772 @@
...
+This deal collateral is intended to provide an assurance mechanism to benefit clients,
+and represents a cost and risk to storage providers.
+Because the deal collateral is calculated on raw byte power, 
+it represents a proportionally 10x larger assurance and cost/risk for unverified data than verified data,
+as compared with the QA power and rewards attributable to the data.
+This proposal reduces the cost of onboarding data by removing the need for deal collateral.
+Parties can still use the built-in market actor for arrangements where deal collateral is desired.
+Future user-programmed actors will also be able to implement more flexible collateral schemes.
+
+## Product Considerations
```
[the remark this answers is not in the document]
anorth: Done

FIPS/fip-0076.md
```diff
@@ -0,0 +1,772 @@
...
+The reason for this change is that verified allocations and deals are independent: a client can allocate DataCap
+without necessarily involving the built-in market actor, and save the SP significant gas costs by doing so.
+
+## Backwards Compatibility
+
+This proposal deprecated the miner methods `PreCommitSector` (method 6), `PreCommitSectorBatch` (method 25),
+and `ProveReplicaUpdates2` (method 29).
+All three have alternatives already available on mainnet that should be used instead.
+
+This proposal requires a state migration to the market actor to add the new `ProviderSectors` mapping,
```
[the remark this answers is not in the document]
anorth: Done

FIPS/fip-direct-onboarding.md
jennijuju: > Parties can have on-chain payments without using the old methods

I understand - however, I think it is too soon to assume all network users want to migrate to the new flow. if we can have a quick mention on if one would like keep their exact existing flow - how should they do so might be useful. i.e: PSD -> PreCommitBatch2 -> ProveCommitSectors/ProveCommitAggregation; or PSD -> PreCommitBatch2 -> ProveCommitSectors2 with RequireNotificationSuccess=true && RequireActivationSuccess==true

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+
+#### Failure handling
+
+Each batched sector activation or update comprises multiple sectors, each with multiple data pieces.
+Each data piece can have a single verified claim and multiple notifications.
+Any of these items might fail, but only limited ability to handle individual failures in a group is practical.
+
+**The activation of each sector is independent**.
+A failed sector activation will not cause a top-level method to abort, unless all activations fail.
+An SP can specify `RequireActivationSuccess=true` to instead require every sector activation to succeed,
```
jennijuju, line 378: is there any reason why we are still using go spec actor convention even its already deprecated and we are only implementing in rust? it would took less time to look up code reference if we just using `require_notification_success` here (and same for the rest of the fip

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+    // Activation manifest for each sector being proven.
+    SectorActivations: []SectorActivationManifest,
+    // Proofs for each sector, parallel to activation manifests.
+    // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+    SectorProofs: [][]byte,
+    // Aggregate proof for all sectors.
+    // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+    AggregateProof: []byte,
+    // Whether to abort if any sector activation fails.
+    RequireActivationSuccess: bool,
```
jennijuju, line 185:
```suggestion
    RequireActivationSuccess: bool,
```

could we rename this to  `RequireAllActivationSuccess` for clarity?

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+    // Whether to abort if any notification returns a non-zero exit code.
+    RequireNotificationSuccess: bool,
```
jennijuju, line 187:
```suggestion
    RequireNotificationSuccess: bool,
```

could we rename this to  `RequireAllNotificationSuccess ` for clarity?

FIPS/fip-0076.md
jennijuju: moving to https://github.com/filecoin-project/FIPs/discussions/730#discussioncomment-7040329 (doesn't block merge but would like alignment before last call / acceptance

FIPS/fip-0076.md
jennijuju: https://github.com/filecoin-project/FIPs/discussions/730#discussioncomment-7040451

jennijuju: this doesn't block the first merge - but would like to have more product discussion before moving towards last call

FIPS/fip-0076.md
anorth: I think the current name captures the semantics clearly.

FIPS/fip-0076.md
anorth: I think the current name captures the semantics clearly already

jennijuju (approved):

FIPS/fip-direct-onboarding.md
jsoares: Resolved in https://github.com/filecoin-project/FIPs/blob/987b3f8c2b36d0ea886ddaae8a4d986ad911c3a8/FIPS/fip-0076.md?plain=1#L691

FIPS/fip-0076.md
jsoares: Agreed it's clear as is.

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+
+This proposal avoids the storage of deal-related information on chain between sector pre-commitment and activation.
+Rather than the looking up piece CIDs and allocation IDs in the built-in storage market actor during activation,
+an SP must provide them explicitly as parameters.
+This movement of data and computation off-chain enables gas savings,
+but requires the storage provider to maintain this metadata during the onboarding process.
+
+### Source of truth for data commitments and verified claims
+
+This proposal decouples the both ability to commit to data, and the allocation of DataCap,
```
jsoares, line 765:
```suggestion
This proposal decouples both the ability to commit to data and the allocation of DataCap
```

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+
+This proposal changes the _relative_ cost of onboarding data with different features, which is fairly uniform at present.
+After this proposal:
+- _unverified_ data _without_ on-chain payments will be the cheapest to onboard;
+- _verified_ data _without_ on-chain payments will be slightly more expensive;
+- _unverified_ data _with_ on-chain payments will be significantly more expensive;
+- _verified_ data _with_ on-chain payments will be the most expensive (but still a little cheaper than current).
+
+The variations in cost account for the variations in computational work necessary to provide incremental functionality.
+These variation might increase the relative profitability of onboarding data that is unverified
```
jsoares, line 659:
```suggestion
These variations might increase the relative profitability of onboarding data that is unverified
```

jsoares (approved):
I have nothing to point out at this stage. The proposal is well supported and very well described, addressing all key concerns raised in discussion and review. Looks ready to merge. Excellent work.

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+may not be extended to other actors in the future (because termination can be invoked by cron).
+
+#### ProveCommitSector
+
+The existing `ProveCommitSector` (method 7) and `ProveCommitAggregate` (method 26) remain,
+and continue to activate deals in the built-in market actor and claim verified allocations using
+the current algorithm.
+These methods no longer writes deal IDs into per-sector chain state.
+
+These methods cannot be used for direct data onboarding and wil always incur the cost of a built-in market deal.
```
kaitlin-beegle, line 420:
```suggestion
These methods cannot be used for direct data onboarding and will always incur the cost of a built-in market deal.
```

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+the deals contained in a sector (if any).
+
+The miner actor no longer carries reliable information about which sectors carry deals.
+
+Note that this notification on termination is a privilege of the built-in market actor that
+may not be extended to other actors in the future (because termination can be invoked by cron).
+
+#### ProveCommitSector
+
+The existing `ProveCommitSector` (method 7) and `ProveCommitAggregate` (method 26) remain,
```
kaitlin-beegle, line 415: Just so I understand- the only group of people who would still be using these methods and submitting sectors via the builtin market actor are those who are (theoretically) working with clients on-chain? Am I misunderstanding this workflow? 

Please note this is just a curiosity; it's unrelated to my editorial review of this draft.

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+
+These methods cannot be used for direct data onboarding and wil always incur the cost of a built-in market deal.
+
+#### ProveReplicaUpdates
+
+The existing `ProveReplicaUpdates` (method 27) remains and continues to activate deals and claim verified allocations
+using the current algorithm.
+This method no longer writes deal IDs into per-sector chain state.
+
+This method cannot be used for direct data onboarding and wil always incur the cost of a built-in market deal.
```
kaitlin-beegle, line 428:
```suggestion
This method cannot be used for direct data onboarding and will always incur the cost of a built-in market deal.
```

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+}
+```
+
+The sector is removed from the `ProviderSectors` mapping, if present.
+Any deals mapped to that sector are marked as terminated, and subsequent processing deferred to cron.
+
+### Migration
+
+The built-in market actor's `ProviderSectors` mapping is initialised from the existing deal state
+and miner actor state per-sector deal IDs.
```
kaitlin-beegle, line 546: Is this jargon that I'm unfamiliar with? Either the hyphen should be removed, or it should read as "state-per-sector".

FIPS/fip-0076.md
```diff
@@ -0,0 +1,812 @@
...
+
+### CommD not required at replica update
+
+FIP-0041 introduced a `ProveReplicaUpdates2` (method 29) in order to add an explicit CommD parameter
+with similar reasoning to that motivating `PreCommitSectorBatch2`.
+However, the specification of piece manifests renders this unnecessary.
+Method 29 is not used by the widely used "Lotus miner" software and has never been invoked on mainnet.
+It is deprecated in order to reduce the number of actively-supported methods.
+The name `ProveReplicaUpdates2` is taken for the new method 34 in order to align the `2` suffix with the
+similarly-behaving `ProveReplicaUpdates2`.
```
kaitlin-beegle, line 593:
```suggestion
similarly-behaving `PreCommitSectorBatch2`.
```

kaitlin-beegle (approved):
Editorial review in.  A few (very small) nits.  Otherwise, great proposal.  Sorry for the delay on a detailed review.

FIPS/fip-0076.md
jsoares: I believe you're supposed to read it as _the per-sector deal IDs in the miner actor state_.

FIPS/fip-0076.md
anorth: Initially, everyone will be using these methods because it will take time for operations to migrate to the new flow. They remain as part of a two-stage migration to avoid requiring all operations to update in synchrony.

In the long run, no-one should use them and I think we should remove them. Parties can still use the built-in storage market actor with the new methods too.

PR #758, opened 2023-07-26
FRC: Piece Multihash CID

aschmahmann:
Defines an alternative CID representation for Filecoin Piece Commitments (i.e. CommP)

[Rendered Version](https://github.com/aschmahmann/FIPs/blob/frc/piece-multihash/FRCs/piece-multihash.md)

Discussion in https://github.com/filecoin-project/FIPs/discussions/759

PR #808, opened 2023-08-24
frc(0069): change v2 piece multihashes to enable arbitrarily sized data

aschmahmann:
This is an alternative building off of #758 that enables piece CIDs that refer to arbitrarily sized data rather than just data of size `N= 2^i * 127/128` where `N` and `i` are integers.

This is option 3 from the FRC Discussion https://github.com/filecoin-project/FIPs/discussions/759#discussioncomment-6554646.

## What this gives us

This means that data of basically any size (there is a maximum of around 2^63) can be represented by a piece CID. As a result:
1. It is possible to reference user data bytes using a piece CID alone (i.e. there's no magic sniffing of zeros or external length metadata required to validate the data)
   - For example, if a 100 MiB CAR file is referenced in a piece aggregate that data could be referenced and verified by CID exactly without having to **also** parse the CAR file to figure out how many zeros are supposed to be at the end of the piece aggregate.
   - For example, if a 6 MiB jpeg was embedded as raw bytes inside of a piece aggregate that jpeg could be referenced loaded and verified directly.
2. It would be possible to create IPLD data structures like UnixFS directories and files that reference piece CID objects
   - This allows existing IPLD tooling to enable referring to groups of piece CIDs and/or mixing piece CID data with other data in ways that have not been previously doable. While the prior proposal would allow embedding a piece CID + the data size (for data that needed zero padding at the end in addition to the fr32 padding) that would not compose with existing formats like UnixFS
   - Note: This capability is not immediately unlocked across the ecosystem because most IPFS implementations are unable to deal with large blocks, however this is an area of active interest as it enables IPFS implementations such as [iroh](https://github.com/n0-computer/iroh) that are heavily dependent on large Blake3 blocks to be compatible with other implementations.
4. Some very common IPFS tooling such as the [IPFS Gateway API](https://specs.ipfs.tech/http-gateways/) (and `ipfs:// URIs) could be used to refer to and load assets (e.g. jpegs, videos, tar files, ...) by their piece CID directly.
   - Note: While their are lots of reasons why piece CIDs are not necessarily as good at representing flat files as using UnixFS with small blocks, a single large blake3 block, the BitTorrent-v2 file format or others there are also reasons why CommP could be useful and this capability is unlocked here.
5. When more generic IPFS tooling for large blocks (e.g. Blake3, SHA2, etc.) comes along (e.g. extending the CAR format to handle large blake3 and sha2 blocks) it will be possible to ship verifiable large CommP blocks without needing separate handling for all the zero padding at the end.

## What it costs us

1. Users may have to understand whether the piece CID they're getting is for data that has been padded (i.e. if someone took chain deal information with padded-size + digest and created a CID out of it) or if it's for data before the padding has been applied.
   - To some extent they need to know this kind of thing anyway and understand both how padding works and handle the processing of CAR files and understand the Filecoin-specific null terminated endings that are implied in the storage formats used in tools like Boost and lotus
2. The CIDs are larger (since the padding information has to be included in the multihash)
3. The implementation is slightly more complicated
4. In this particular iteration (that obeys the unsigned-varint spec) the maximum piece size is smaller (although still in the EiB range so plenty large)

Note: This is true in the current version as well, but by having the multihash do the fr32 padding for us it means we cannot have a v2 piece CID that leverages data in fr32 but doesn't require the padding (while I'm not familiar with the internals here it seems to be indicated by https://spec.filecoin.io/#section-systems.filecoin_files.piece.data-representation that this is possible). Given that AFAIK all data today uses fr32 padding and I've heard of no plans for not requiring the padding this doesn't seem like a loss.

## Outstanding questions

There are some minor questions around formatting tagged in the PR for threaded discussion, but mostly the question is whether there are issues with this feature or if it's sufficiently desirable to move ahead.

Some people who might be interested in review:

@Stebalien + @gozala (you both asked for this 😄),  @ribasushi, @Kubuxu, @willscott, @masih (this seems relevant for your CommPact proposal)

### DIFF
--- a/FRCs/frc-0069.md
+++ b/FRCs/frc-0069.md
@@ -77,7 +77,11 @@ A tuple of (v1 Piece CID, Piece size) can be converted into a valid v2 Piece CID
 ## Test Cases
 
 Take data of size 127*4 bytes where the first 127 bytes are 0, the next 127 are 1, the next 127 are 2, and the last 127 are 3. The v1 piece CID of this data would be `baga6ea4seaqes3nobte6ezpp4wqan2age2s5yxcatzotcvobhgcmv5wi2xh5mbi`. The multihash-based piece CID would be `bafkzcibcaaces3nobte6ezpp4wqan2age2s5yxcatzotcvobhgcmv5wi2xh5mbi`. With a base16 multibase this would be `f01559120220004496dae0cc9e265efe5a006e80626a5dc5c409e5d3155c13984caf6c8d5cfd605` or equivalently (`(multibase = f) | (CIDv1 prefix = 0x01) | (Raw codec = 0x55) | (fr32-sha2-256-trunc254-padded-binary-tree multihash encoded varint = 0x9120) | (length of digest = 0x22) | (amount of data padding = 0x00) | (tree height = 0x04) | (underlying hash digest = 496...605)`)
+Take payload of size 0 bytes. There MUST be `127` bytes of padding. The v2 piece CID MUST be `bafkzcibcp4bdomn3tgwgrh3g532zopskstnbrd2n3sxfqbze7rxt7vqn7veigmy`.
 
+Take payload of 127 bytes where all bytes are 0. There MUST be `0` bytes of padding. The hight MUST be `2`. The v2 piece CID MUST be `bafkzcibcaabdomn3tgwgrh3g532zopskstnbrd2n3sxfqbze7rxt7vqn7veigmy`.
+
+Take payload of 128 bytes where all bytes are 0. There MUST be `126` bytes of padding. The height MUST be `3`. The v2 piece CID MUST be `bafkzcibcpybwiktap34inmaex4wbs6cghlq5i2j2yd2bb2zndn5ep7ralzphkdy`
 Given the piece CID v1 of the empty 32 GiB piece (i.e. 32 * 2^30 * 127/128 bytes of zeros)`baga6ea4seaqao7s73y24kcutaosvacpdjgfe5pw76ooefnyqw4ynr3d2y6x2mpq` the corresponding mutlihash piece CID would be `bafkzcibcaapao7s73y24kcutaosvacpdjgfe5pw76ooefnyqw4ynr3d2y6x2mpq`
 
 Given the piece CID v1 of the empty 64 GiB piece (i.e. 64 * 2^30 * 127/128 bytes of zeros)`baga6ea4seaqomqafu276g53zko4k23xzh4h4uecjwicbmvhsuqi7o4bhthhm4aq` the corresponding piece mutlihash piece CID would be `bafkzcibcaap6mqafu276g53zko4k23xzh4h4uecjwicbmvhsuqi7o4bhthhm4aq`
--- a/FRCs/frc-0069.md
+++ b/FRCs/frc-0069.md
@@ -43,12 +43,11 @@ The digest for the multihash is a variable number of bytes.
 It can be roughly described as `uvarint padding | uint8 height | 32 byte root data` where `|` means concatenation.
 
 - The first bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
-    - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115` (TODO: Is this necessary or would even 32 bytes be ok?)
+    - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115` 
     - For example, if the data is of size 254 (i.e. 2^i * 127/128, where i is a positive integer >= 7) then this will be 0
     - For example, if the data is of size 256 then the padding is `508-256=252`
     - Note: because the unsigned-varint spec currently has a maximum representable size of 2^63-1 (and 9 bytes to represent the varint) this puts a cap on the maximum size of data representable by this multihash as well
-    - Note: because this is padding data it must be less than the size of the underlying data
-       - (TODO: this is not true for data less than 64 bytes if padding up to 127)
+    - Note: because this is padding data it must be less than the size of the underlying data (with the exception of data less than 127 bytes which is always padded up to 127 bytes)
 - The next byte defines the height of the tree
     - For example if the first byte is 0 the tree is a single 32-byte leaf
     - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB. 
--- a/FRCs/frc-0069.md
+++ b/FRCs/frc-0069.md
@@ -12,7 +12,7 @@ created: 2023-07-26
 
 ## Simple Summary
 
-Introduces an alternative CID representation for the FR32 padded sha256-trunc254-padded binary merkle trees used in Filecoin Piece Commitments (i.e. CommP). In it we use the [Raw codec](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L41) and a new [sha2-256-trunc254-padded-binary-tree multihash (or piece multihash)](https://link.tld) rather than the [Fil-commitment-unsealed codec](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L517) and the [sha2-256-trunc254-padded multihash](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L149).
+Introduces an alternative CID representation for the FR32 padded sha256-trunc254-padded binary merkle trees used in Filecoin Piece Commitments (i.e. CommP). In it we use the [Raw codec](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L41) and a new [sha2-256-trunc254-padded-binary-tree multihash (or piece multihash)](https://github.com/multiformats/multicodec/pull/331) rather than the [Fil-commitment-unsealed codec](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L517) and the [sha2-256-trunc254-padded multihash](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L149).
 
 
 ## Abstract
@@ -24,7 +24,7 @@ For example, in much of the relevant portions of the Filecoin spec and lotus' Go
 
 This makes it much more natural to work with new concepts like [Deal Aggregates](https://pkg.go.dev/github.com/filecoin-project/go-data-segment@v0.0.0-20230605095649-5d01fdd3e4a1/datasegment#NewAggregate), as proposed in [FRC-0058](https://github.com/filecoin-project/FIPs/blob/7e499523c9c7ed2c48c6a36967f7f011cee1fefd/FRCs/frc-0058.md).
 
-To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://link.tld) which combines the root hash with the tree height and the amount of padding of the data with zeros such that the result is a full and balanced binary tree after the fr32 padding is applied. If this multihash were to be used to reference the full piece as understood by the Filecoin on-chain consensus mechanism, this would involve the special case where the padding is zero is used.
+To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://github.com/multiformats/multicodec/pull/331) which combines the root hash with the tree height and the amount of padding of the data with zeros such that the result is a full and balanced binary tree after the fr32 padding is applied. If this multihash were to be used to reference the full piece as understood by the Filecoin on-chain consensus mechanism, this would involve the special case where a padding of zero is used.
 
 ## Specification
 
@@ -36,18 +36,25 @@ A CIDv1 requires a:
 
 The core component introduce in this specification is a new multihash type fr32-sha2-256-trunc254-padded-binary-tree multihash.
 
-The multihash code for this type is 0x1011 as identifier in the [multicodec code table](https://link.tld).
+The multihash code for this type is 0x1011 as identified in the [multicodec code table](https://github.com/multiformats/multicodec/pull/331).
 
 The digest for the multihash is a variable number of bytes.
-- The first byte defines the height of the tree
-    - For example if the first byte is 0 the tree is a single 32-byte leaf
-    - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB.
-- The next bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
+
+It can be roughly described as `uvarint padding | uint8 height | 32 byte root data` where `|` means concatenation.
+
+- The first bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
     - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115` (TODO: Is this necessary or would even 32 bytes be ok?)
     - For example, if the data is of size 254 (i.e. 2^i * 127/128, where i is a positive integer >= 7) then this will be 0
     - For example, if the data is of size 256 then the padding is `508-256=252`
     - Note: because the unsigned-varint spec currently has a maximum representable size of 2^63-1 (and 9 bytes to represent the varint) this puts a cap on the maximum size of data representable by this multihash as well
-    - Note: because this is padding data it must be less than the size of the underlying data 
+    - Note: because this is padding data it must be less than the size of the underlying data
+       - (TODO: this is not true for data less than 64 bytes if padding up to 127)
+- The next byte defines the height of the tree
+    - For example if the first byte is 0 the tree is a single 32-byte leaf
+    - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB. 
+- The last 32 bytes are the value at the root of the binary tree
+
+Note: This structure is such that the last 33 bytes of the multihash are `uint 8 height | 32 byte root data` which is the data that the proofs underlying Filecoin consensus can attest to (i.e. the proofs don't know how much of the padding is padding vs zeros that are part of the user data).
 
 ### Raw codec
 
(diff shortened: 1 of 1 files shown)


FRCs/frc-0069.md
```diff
@@ -38,9 +38,16 @@ The core component introduce in this specification is a new multihash type fr32-
...
-
-Note that the data processed by this hash function must be of size `N = 2^i * 127/128` where `i` is any positive integer >=7 and <=255. This means that the minimum hashable amount of data is 127 bytes.
+The digest for the multihash is a variable number of bytes.
+- The first byte defines the height of the tree
+    - For example if the first byte is 0 the tree is a single 32-byte leaf
+    - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB.
+- The next bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
+    - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115` (TODO: Is this necessary or would even 32 bytes be ok?)
```
aschmahmann, line 46: I'm not sure whether the minimum size here should be 31 bytes or 127 bytes. My understanding is that fr32 padding is defined only on data > 127 bytes and so we should pad up to 127, but I'm not confident that that's correct.

cc @Kubuxu

FRCs/frc-0069.md
```diff
@@ -38,9 +38,16 @@ The core component introduce in this specification is a new multihash type fr32-
...
 The multihash code for this type is 0x1011 as identifier in the [multicodec code table](https://link.tld).
 
-The digest for the multihash is 33 bytes. The first byte defines the height of the tree. For example if the first byte is 0 the tree is a single 32-byte leaf. Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB.
-
-Note that the data processed by this hash function must be of size `N = 2^i * 127/128` where `i` is any positive integer >=7 and <=255. This means that the minimum hashable amount of data is 127 bytes.
+The digest for the multihash is a variable number of bytes.
+- The first byte defines the height of the tree
+    - For example if the first byte is 0 the tree is a single 32-byte leaf
+    - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB.
+- The next bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
```
aschmahmann, line 45: While it would be more natural to instead of encoding `height | padding` to just encode `data size` a nice feature of this approach is that it decreases the size of the CIDs of all full trees (e.g. CC sectors, prepadded data in existing places on the chain where the PieceInfo abstraction is used) and makes them more uniform (i.e. they all have padding of zero). This approach was recommended by Mikeal Rogers.

I don't feel strongly about this though if people would prefer something else like encoding the size, or using a different more compact representation of the size information.

FRCs/frc-0069.md
```diff
@@ -38,9 +38,16 @@ The core component introduce in this specification is a new multihash type fr32-
...
+    - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115` (TODO: Is this necessary or would even 32 bytes be ok?)
+    - For example, if the data is of size 254 (i.e. 2^i * 127/128, where i is a positive integer >= 7) then this will be 0
+    - For example, if the data is of size 256 then the padding is `508-256=252`
+    - Note: because the unsigned-varint spec currently has a maximum representable size of 2^63-1 (and 9 bytes to represent the varint) this puts a cap on the maximum size of data representable by this multihash as well
```
aschmahmann, line 49: While we don't strictly need to follow the unsigned-varint spec here and have a maximum size it seems like a pretty reasonable thing to do. Any disagreement?

Gozala (approved):
Looks good to me, however I'm not sure if the variable size CIDs this would lead to would be worth it as I suspect it may break code in the wild (I've seen bunch of eth contracts in nft space that hard code expected CID size)

FRCs/frc-0069.md
```diff
@@ -24,7 +24,7 @@ For example, in much of the relevant portions of the Filecoin spec and lotus' Go
 
 This makes it much more natural to work with new concepts like [Deal Aggregates](https://pkg.go.dev/github.com/filecoin-project/go-data-segment@v0.0.0-20230605095649-5d01fdd3e4a1/datasegment#NewAggregate), as proposed in [FRC-0058](https://github.com/filecoin-project/FIPs/blob/7e499523c9c7ed2c48c6a36967f7f011cee1fefd/FRCs/frc-0058.md).
 
-To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://link.tld) which combines the root hash with the tree height (which in the full and balanced binary trees used in Piece commitments is equivalent to size).
+To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://link.tld) which combines the root hash with the tree height and the amount of padding of the data with zeros such that the result is a full and balanced binary tree after the fr32 padding is applied. For Piece commitments the amount of added padding is zero.
```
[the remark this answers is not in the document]
aschmahmann: I was trying to draw some distinction between:
1. Piece commitment -> the thing Filecoin consensus proves that you have (i.e. all the data including the padding)
2. Hashing the data using the specific set of machinery used in piece commitments (e.g. leaves are fr32 padded and it's a balanced binary tree)

Maybe this wasn't the best distinction. Any idea how to best state that if you try and take the digests used in Filecoin commitments/consensus (e.g. the v1 piece CIDs used in https://pkg.go.dev/github.com/filecoin-project/go-state-types/abi#PieceInfo) then they have no padding after conversion to v2 piece CIDs since they've already been padded (absent any off-chain information telling you how they might've been padded).

FRCs/frc-0069.md
aschmahmann: As I understand that 65 (and probably this could've been 64) is saying:
1. 128 bytes is the smallest tree I'll deal with
2. I don't want to answer the question about what happen with the next smallest size tree (64 bytes)
3. So 65 (really 64 I think) is the smallest since it will (after fr32 padding) exceed 64 bytes and have to round up to 128

The question is if we want to loosen this to something else like:
1. If <127 bytes pad the data up to 127 (i.e. a single byte is allowed but it means it comes with 126 zeros)
   - i.e. Stick with 128 as the minimum tree size and define what happens below 64 bytes
2. If <`Smaller` (don't think smaller than 31/32 is possible here) then pad up to `Smaller` 
   - i.e. Choose a different minimum tree size

IIUC the "rule" for the tree construction is that each leaf node must be an fr32 element which can be guaranteed if the last two bits of each 32 byte node are zero. If so then even a single byte can be padded out to 32 bytes and used as a node.

anorth:
@ribasushi detailed technical discussion like this is better in the discussion thread. It will get lost when this PR is merged or closed. And this still might be a reasonable FRC even if flawed – it's not a requirement that anyone use it.

FRCs/frc-0069.md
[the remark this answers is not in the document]
rvagg:
```suggestion
To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://link.tld) which combines the root hash with the tree height and the amount of padding of the data with zeros such that the result is a full and balanced binary tree after the fr32 padding is applied. If this multihash were to be used to reference the full piece as understood by the on-chain consensus mechanism, this would involve the special case where the padding is zero is used.
```

Seems that you're trying to tread a careful path here because lower down in the original form of this doc you have:

> this is not a FIP for changing how CommPs are stored or used on chain, but rather a more user-friendly way of referencing Pieces which can be used in various non-chain interactions (e.g. piece-based retrieval, smart contract interactions, etc.).

I wonder if there could be some more work nuancing these points a bit more. ISTM that this could be an alternate on-chain CID too, there could potentially be a world in which consensus involves optionally understanding both forms of CID; maybe even with a migration toward a "v1". This point goes for both the original FRC text and this updated one I think.

FRCs/frc-0069.md
rvagg: My original understanding when we were working out CommP-as-a-CID was that 256 was the minimum, it was always explained to me as "2 extra bits per 254"; but we now live in this 126+1 world, which seems to be convenient from a calculation perspective, but does that have any implications for lower bound? So some clarity about what's actually required would be neat.

rvagg: (the `sha2-256-trunc254` bit came from this btw, the extra `00` hole at the end of each sha2-256 digest / chunk)

FRCs/frc-0069.md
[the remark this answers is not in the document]
aschmahmann: I think this could be used within consensus if we wanted to. However, I figured that's a discussion for another time. I didn't want to end up in a situation where the process to get this FRC agreed upon and landed slowed down even more due to concerns about the complexity of a migration, what the state size will look like, ...

If you'd like to add text in there about this I'm fine with that, I just left it out to try and reduce potential confusion.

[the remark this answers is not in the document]
aschmahmann:
```suggestion
To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://link.tld) which combines the root hash with the tree height and the amount of padding of the data with zeros such that the result is a full and balanced binary tree after the fr32 padding is applied. If this multihash were to be used to reference the full piece as understood by the Filecoin on-chain consensus mechanism, this would involve the special case where the padding is zero is used.
```

@rvagg I liked your suggestion, I just modified it slightly.

aschmahmann:
Note: Updated the FIP to put the padding before the height. This was a result of an alignment conversation at FilDev Iceland to come to an agreement with a number of interested parties about what the multihash should look like (e.g. #758, this one or something else).

Thank you to everyone who participated. My recollection of participants (please ping me if I missed someone) were:
- @ribasushi (Spade, Singularity and working on integrations with others like .storage)
- @Stebalien (FVM)
- @magik6k (lotus)
- @masih (Boost + IPNI)
- @taoshengshi (Ken Labs - no direct impact at the moment)

FRCs/frc-0069.md
```diff
@@ -24,7 +24,7 @@ For example, in much of the relevant portions of the Filecoin spec and lotus' Go
 
 This makes it much more natural to work with new concepts like [Deal Aggregates](https://pkg.go.dev/github.com/filecoin-project/go-data-segment@v0.0.0-20230605095649-5d01fdd3e4a1/datasegment#NewAggregate), as proposed in [FRC-0058](https://github.com/filecoin-project/FIPs/blob/7e499523c9c7ed2c48c6a36967f7f011cee1fefd/FRCs/frc-0058.md).
 
-To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://link.tld) which combines the root hash with the tree height (which in the full and balanced binary trees used in Piece commitments is equivalent to size).
+To resolve this we introduce a new multihash type [fr32-sha2-256-trunc254-padded-binary-tree multihash](https://github.com/multiformats/multicodec/pull/331) which combines the root hash with the tree height and the amount of padding of the data with zeros such that the result is a full and balanced binary tree after the fr32 padding is applied. If this multihash were to be used to reference the full piece as understood by the Filecoin on-chain consensus mechanism, this would involve the special case where a padding of zero is used.
```
Gozala, line 27: I'm failing to understand the last sentence, what is that special case ? I think it would be a good idea if this was clear to someone like me who's not well versed in filecoin on-chain consensus mechanism or we simple have not mentioned it.

FRCs/frc-0069.md
Gozala: I don't think we deal with such large payloads so we're probably ok, but probably good to confirm with @ribasushi if that could be an issue elsewhere

Gozala (approved):
This looks good to me.

FRCs/frc-0069.md
```diff
@@ -63,11 +77,17 @@ A tuple of (v1 Piece CID, Piece size) can be converted into a valid v2 Piece CID
...
+
+Take data of size 127*8 bytes where the first where the first 127 bytes are 0, the next 127 are 1, the next 127 are 2, the next 127 are 3 and the remaining `127*4` bytes are 0. There is no padding so the v1 piece CID would be `baga6ea4seaqn42av3szurbbscwuu3zjssvfwbpsvbjf6y3tukvlgl2nf5rha6pa` and the v2 would be `bafkzcibcauan42av3szurbbscwuu3zjssvfwbpsvbjf6y3tukvlgl2nf5rha6pa`.
 
-Given the piece CID v1 of the empty 32 GiB piece `baga6ea4seaqao7s73y24kcutaosvacpdjgfe5pw76ooefnyqw4ynr3d2y6x2mpq` the corresponding mutlihash piece CID would be `bafkzcibbdydx4x66gxcqveyduviaty2jrjhl5x7ttrbloefxgdmoy6whv6td4`
+Take data of size 128*4 bytes where the first where the first 127 bytes are 0, the next 127 are 1, the next 127 are 2, the next 127 are 3 and the remaining 127+4 bytes are 0. There is `377` bytes of padding needed and so the v1 piece CID is `baga6ea4seaqn42av3szurbbscwuu3zjssvfwbpsvbjf6y3tukvlgl2nf5rha6pa` (notice it's the same as above) and the v2 is `bafkzcibd7ebalxticxolgseegik2stpfgkkuwyf6kufex3doorkvmzpjuxwe4dz4`.
```
Gozala, line 88: I'm not sure I understand why padding here would be 377 as opposed to 504. Tree height is 5 so we need to feed `(2 ** 5) * 32 / 128 * 127 = 1016` bytes into Fr32 expander. That is to say I'd expect padding to be `1016 - 128 * 4 = 503`.

Maybe I'm misunderstanding spec but `128 * 4 + 377` is `7` quads and we need one more here to form binary tree.

ribasushi: I am also confused... it should be neither 377 nor 504: it should be 512

```
Payload:              512 bytes
Fr32 expansion:       640 bytes ( I think )
256bit node bintree: 1024 bytes
bintree - payload:    512 bytes
```

Gozala: So if payload was 512 bytes after 0 padding to 635 and Fr32 expansion we'll get 640 bytes. From that is 5 quads and nearest pow2 is 8 which will form 1024 bytes indeed.

And yes 1024 - 512 = 512, however that is not 0 padding pre Fr32 which is why my understanding of this spec is.

Gozala: After some explanations in side channel I understand now that @ribasushi's version is number 0 added by both Fr32 padding and 0 padding together while mine was number of added by 0 padding.

I have minor preference for my interpretation, even if it's more complicated to calculate I find it bit more intuitive to understand. That said I'm ok with going with @ribasushi's interpretation as long as spec reflects that

anorth (commented):
There are still a couple of TODOs in the text, but I'll be happy to approve it after they are removed. Alternatively please embrace the new TODO section recently added to the FIP template and pull them up to be clearly identified as outstanding issues.

aschmahmann:
@ribasushi @Gozala if you could take another look that would be great. Additionally, if you have time to add test fixtures for data under 127 bytes that would be appreciated.

anorth (approved):

FRCs/frc-0069.md
aschmahmann: Perhaps a fun fact: Switching from a v1 piece CID to a v2 one (i.e. assuming zero padding) results in a piece CID that is exactly the same size.
- v1: `<0x01><codec = 3 bytes as varint><multihash = 2 bytes as varint><mh-digest-len = 1 byte><digest = 32 bytes> = 39 bytes`
- v2: `<0x01><codec = 1 byte as varint><multihash = 2 bytes as varint><mh-digest-len=1 byte><padding=0x00><height = 1 byte><root-hash = 32 bytes> = 39 bytes`

FRCs/frc-0069.md
```diff
@@ -36,11 +36,24 @@ A CIDv1 requires a:
...
+The multihash code for this type is 0x1011 as identified in the [multicodec code table](https://github.com/multiformats/multicodec/pull/331).
 
-The digest for the multihash is 33 bytes. The first byte defines the height of the tree. For example if the first byte is 0 the tree is a single 32-byte leaf. Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB.
+The digest for the multihash is a variable number of bytes.
 
-Note that the data processed by this hash function must be of size `N = 2^i * 127/128` where `i` is any positive integer >=7 and <=255. This means that the minimum hashable amount of data is 127 bytes.
+It can be roughly described as `uvarint padding | uint8 height | 32 byte root data` where `|` means concatenation.
+
+- The first bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
+    - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115`
```
ribasushi, line 46:
```suggestion
Due to the fr32 expansion mechanism, **if the input data size is < 127 bytes then it is padded to a size of 127 bytes**. For example, if the data is of size 12 the padding will be `127-12 = 115` 

```

ribasushi, line 46: - it's not padding, padding implies byte-work, fr32 is not that...
- 

```suggestion
- The first bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of `0x00` bytes needed to pad the underlying data such that after FR32 expansion it will be a fully balanced binary tree. Due to the fr32 expansion mechanism, **if the input data size is < 127 bytes then it is padded to a size of 127 bytes**. That is, a CID with a certain padding value of `P` and a certain height (see next section) `H`, encodes an initial payload size of `( 1 << (H-2) ) * 127 - P`
    - For example, if the data is of size 12 the padding will be `127-12 = 115` 

```

FRCs/frc-0069.md
```diff
@@ -36,11 +36,24 @@ A CIDv1 requires a:
...
+It can be roughly described as `uvarint padding | uint8 height | 32 byte root data` where `|` means concatenation.
+
+- The first bytes are a [uvarint](https://github.com/multiformats/unsigned-varint) of the number of bytes needed to pad the underlying data such that after FR32 padding it will be a full binary tree
+    - If the data is < 127 bytes then it is padded to 127 bytes. For example, if the data is of size 12 the padding will be `127-12 = 115` 
+    - For example, if the data is of size 254 (i.e. 2^i * 127/128, where i is a positive integer >= 7) then this will be 0
+    - For example, if the data is of size 256 then the padding is `508-256=252`
+    - Note: because the unsigned-varint spec currently has a maximum representable size of 2^63-1 (and 9 bytes to represent the varint) this puts a cap on the maximum size of data representable by this multihash as well
+    - Note: because this is padding data it must be less than the size of the underlying data (with the exception of data less than 127 bytes which is always padded up to 127 bytes)
```
ribasushi, line 50: Given the `( 1 << (H-2) ) * 127 - P` formula, is this sentence even needed...?

FRCs/frc-0069.md
```diff
@@ -36,11 +36,24 @@ A CIDv1 requires a:
...
+- The next byte defines the height of the tree
+    - For example if the first byte is 0 the tree is a single 32-byte leaf
```
ribasushi, line 52:
```suggestion
    - For example if the first byte is 0 the tree is a single 32-byte node

```

FRCs/frc-0069.md
```diff
@@ -36,11 +36,24 @@ A CIDv1 requires a:
...
+    - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since the leaves must be 32 bytes, represents a piece of size 32*2^30 bytes = 32GiB.
```
ribasushi, line 53:
```suggestion
    - Similarly, if the first byte is 30 then the tree is 30 levels deep which, since all nodes used by this hash are 32 bytes long, represents a piece of size 32*2^30 bytes = 32GiB. 

```

FRCs/frc-0069.md
```diff
@@ -63,11 +76,17 @@ A tuple of (v1 Piece CID, Piece size) can be converted into a valid v2 Piece CID
```
ribasushi, line 76: Making a note about line 69: `being able to convert between v1 and v2 Piece CIDs`: one **can not** do that. You can downgrade a v2 to V1+Size, but as written it implies one can up-cast a V1 to a V2, like you can do a CidV0 to CidV1, which... nope.

ribasushi (changes requested):
Made a bunch of comments... I am sorry. Some are suggestions, some are "must have". It should be obvious which is which.

FRCs/frc-0069.md
Gozala: @ribasushi I think you meant "hash payload" and not "hashed payload" right ?

FRCs/frc-0069.md
```diff
@@ -63,11 +76,17 @@ A tuple of (v1 Piece CID, Piece size) can be converted into a valid v2 Piece CID
...
+Take data of size 127*4 bytes where the first 127 bytes are 0, the next 127 are 1, the next 127 are 2, and the last 127 are 3. The v1 piece CID of this data would be `baga6ea4seaqes3nobte6ezpp4wqan2age2s5yxcatzotcvobhgcmv5wi2xh5mbi`. The multihash-based piece CID would be `bafkzcibcaaces3nobte6ezpp4wqan2age2s5yxcatzotcvobhgcmv5wi2xh5mbi`. With a base16 multibase this would be `f01559120220004496dae0cc9e265efe5a006e80626a5dc5c409e5d3155c13984caf6c8d5cfd605` or equivalently (`(multibase = f) | (CIDv1 prefix = 0x01) | (Raw codec = 0x55) | (fr32-sha2-256-trunc254-padded-binary-tree multihash encoded varint = 0x9120) | (length of digest = 0x22) | (amount of data padding = 0x00) | (tree height = 0x04) | (underlying hash digest = 496...605)`)
+
```
Gozala, line 80:
```suggestion
Take payload of size 0 bytes. There MUST be `127` bytes of padding. The v2 piece CID MUST be `bafkzcibcp4bdomn3tgwgrh3g532zopskstnbrd2n3sxfqbze7rxt7vqn7veigmy`.

Take payload of 127 bytes where all bytes are 0. There MUST be `0` bytes of padding. The hight MUST be `2`. The v2 piece CID MUST be `bafkzcibcaabdomn3tgwgrh3g532zopskstnbrd2n3sxfqbze7rxt7vqn7veigmy`.

Take payload of 128 bytes where all bytes are 0. There MUST be `126` bytes of padding. The height MUST be `3`. The v2 piece CID MUST be `bafkzcibcpybwiktap34inmaex4wbs6cghlq5i2j2yd2bb2zndn5ep7ralzphkdy`
```

Gozala (approved):

FRCs/frc-0069.md
ribasushi: > @ribasushi I think you meant "hash payload" and not "hashed payload" right ?

I meant `hashed payload`: the data (payload) that is to be hashed. We can switch to `hashed data`, it sounds clunkier to me...

> struct can be losslessly expressed as a PieceCIDv2 with the padding value set to 0 

`s/losslessly/fully/` ?

FRCs/frc-0069.md
ribasushi: If your PieceCid describes something larger than 63 bits ( 8 EiB ), then the "actual payload size" becomes meaningless anyway, and you are in world with padding of `0` and height up to 127, which is about 1<<132 or... many many **many** [HellaBytes](https://simple.wikipedia.org/wiki/Hellabyte)

FRCs/frc-0069.md
Gozala: For what it's worth I think formula should be `(2 ^ (H-2)) * 127 - P` I know it's the same, but this version does not require you to mentally do the translation.

Gozala:
I am not sure where to put this feedback exactly so just posting it here. I think it would be really great if this spec covered more details of the Piece multihash / CID internals, specifically it would be great if it had sub sections for

1. Payload 0-padding and an the requirement imposed by invariants of fr32-expansion
2. Fr32 expansion and some rational for it (I actually don't know what the motivation is)
3. Merkleization that explaining how binary tree is build

From there it is a lot easier to explain that padding, tree height and root are encoded in the multihash in a specific way

aschmahmann:
> Payload 0-padding and an the requirement imposed by invariants of fr32-expansion

Not sure if some of @ribasushi's suggestions cover this, but suggestions welcome (maybe better after this PR lands so that it's clear that we're done on the mechanics).

> Fr32 expansion and some rational for it (I actually don't know what the motivation is)
> Merkleization that explaining how binary tree is build

Both of these really should be covered already, unfortunately the Filecoin spec (https://specs.filecoin.io/) does not do a great job here. It gives a bunch of the details but not in a particularly nice way. Not sure if this belongs in this FRC vs wherever Filecoin specs should live these days, but agreed that the lack of good info here hurts implementers.

FRCs/frc-0069.md
```diff
@@ -12,7 +12,7 @@ created: 2023-07-26
...
+Introduces an alternative CID representation for the FR32 padded sha256-trunc254-padded binary merkle trees used in Filecoin Piece Commitments (i.e. CommP). In it we use the [Raw codec](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L41) and a new [sha2-256-trunc254-padded-binary-tree multihash (or piece multihash)](https://github.com/multiformats/multicodec/pull/331) rather than the [Fil-commitment-unsealed codec](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L517) and the [sha2-256-trunc254-padded multihash](https://github.com/multiformats/multicodec/blob/566eaf857a9d20573d3910221db7b34d98e8a0fc/table.csv#L149).
```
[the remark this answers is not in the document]
Gozala: I'm not eager to change lib names but ok if it's a must. publishing packages with last name triggered npm's spam filter so I had to email etc... I suspect same will happen with this name also.

FRCs/frc-0069.md
```diff
@@ -12,7 +12,7 @@ created: 2023-07-26
 
 ## Simple Summary
```
kaitlin-beegle, line 13: Regardless of which version of the Simple Summary is merged by the author, can we add a few extra words about why this alternative CID representation is being introduced? What functionality or improvement does it support?

FRCs/frc-0069.md
```diff
@@ -24,7 +24,7 @@ For example, in much of the relevant portions of the Filecoin spec and lotus' Go
 
 This makes it much more natural to work with new concepts like [Deal Aggregates](https://pkg.go.dev/github.com/filecoin-project/go-data-segment@v0.0.0-20230605095649-5d01fdd3e4a1/datasegment#NewAggregate), as proposed in [FRC-0058](https://github.com/filecoin-project/FIPs/blob/7e499523c9c7ed2c48c6a36967f7f011cee1fefd/FRCs/frc-0058.md).
```
kaitlin-beegle, line 25:
```suggestion
The proposed V2 multihash makes it easier to work with new concepts like [Deal Aggregates](https://pkg.go.dev/github.com/filecoin-project/go-data-segment@v0.0.0-20230605095649-5d01fdd3e4a1/datasegment#NewAggregate), as proposed in [FRC-0058](https://github.com/filecoin-project/FIPs/blob/7e499523c9c7ed2c48c6a36967f7f011cee1fefd/FRCs/frc-0058.md).
```

kaitlin-beegle: Nit, but the use of the word "concepts" in the above sentence could also benefit from specification. Is it possible to do so? Perhaps "unstructured", "variably sized" or "large" data, etc., whatever is/may be accurate.

kaitlin-beegle (approved):

PR #819, opened 2023-08-31
Initial draft for on chain commd

ZenGround0:
Depends on direct data on boarding draft: #730 

Ready for review 
cc @Fatman13

### DIFF
--- a/FIPS/fip-on-chain-commd.md
+++ b/FIPS/fip-on-chain-commd.md
@@ -1,0 +1,119 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
+title: On Chain CommD
+author: <a list of the author's or authors' name(s) and/or username(s), or name(s) and email(s), e.g. (use with the parentheses or triangular brackets): @zenground0, Alex North (@anorth), ...
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/496, https://github.com/filecoin-project/FIPs/discussions/760
+status: Draft
+type: Technical
+category: Core
+created: 2023-08-30
+requires: Draft: https://github.com/filecoin-project/FIPs/pull/804
+---
+
+<!--You can leave these HTML comments in your merged FIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new FIPs. Note that a FIP number will be assigned by an editor. When opening a pull request to submit your FIP, please use an abbreviated title in the filename, `fip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+Allow optional storage of CommD within a new field of the sector info.  Adding CommD is specified by a field in the parameters of the new ProveCommit and ProveReplicaUpdate methods proposed in [Direct data onboarding](https://github.com/filecoin-project/FIPs/pull/804).
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+Today there is no way to directly prove that sectors contain particular content after activation.  This is a useful pattern for smart contracts operating over data.  Putting CommD on chain unlocks  this feature as CommD can be accessed by miner actor code and serve as an onchain witness of sector content for inclusion proofs. 
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+The filecoin development community has a demonstrated interest in enabling innovation in smart contracts that interact with verifiable storage.  It is the FVM's main differentiator when compared with other blockchain runtimes.  Two different groups have arrived at similar conclusions on a particular useful direction for the miner actor in relation to these goals:  myself [here](https://github.com/filecoin-project/FIPs/discussions/496) and venus dev team [here](https://github.com/filecoin-project/FIPs/discussions/760).  The first step in this direction is to allow SPs to keep CommD on chain in the sector info.
+
+Adding CommD to sector info requires a sector info migration which on its own is a costly in terms of network development and risk.  Similarly it will require changes to parameters being introduced in ongoing [direct data onboarding (DDO) work](https://github.com/filecoin-project/FIPs/pull/804). Doing this work at the same time as DDO (which also needs a full sector info migration) makes the upgrade risk and development cost overhead much smaller.
+
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+### Add *CommD to SectorInfo 
+
+Add an optional Cid field to the sector info on chain: 
+
+```rust
+pub struct SectorOnChainInfo {
+    pub sector_number: SectorNumber,
+   ...
+    pub unsealed_cid: Option<Cid>,  // NEW
+   ...
+    // Flag for QA power mechanism introduced in fip 0045
+    pub simple_qa_power: bool,
+}
+```
+
+The migration to achieve this will simply set `UnsealedCid = nil`.  Only new sectors will be able to set CommD.
+
+### Add flag for setting CommD to new DDO proving methods
+
+```go
+struct SectorActivationManifest {
+    // Sector to be activated.
+    Sector: SectorNumber,
+    // Pieces comprising the sector content, in order.
+    Pieces: []PieceActivationManifest,
+    KeepCommD: bool, // NEW
+}
+```
+
+
+New ProveCommit and PRU messages will then parse this new bool flag and store CommD if specified.
+If CommD is the zero data CommD then the system will always avoid storing even if KeepCommD is true.
+
+## Design Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+
+### Implementing CommD on chain
+Since CommD is a property of an individual sector if it is onchain it belongs on the sector info.
+
+Specifying inclusion as part of the DDO sector activation manifest allows for user control over which specific sectors can store CommD.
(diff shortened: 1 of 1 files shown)


FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+
+```rust
+pub struct SectorOnChainInfo {
+    pub sector_number: SectorNumber,
+   ...
+    pub unsealed_cid: Option<Cid>,  // NEW
+   ...
+    // Flag for QA power mechanism introduced in fip 0045
+    pub simple_qa_power: bool,
+}
```
hunjixin, line 46:
```rs
pub struct SectorOnChainInfo {
    pub sector_number: SectorNumber,
   ...
    pub piece_cids: Option<Cid>,  // NEW  Amt[string]
   ...
    // Flag for QA power mechanism introduced in fip 0045
    pub simple_qa_power: bool,
}
```

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+
+```go
+struct SectorActivationManifest {
+    // Sector to be activated.
+    Sector: SectorNumber,
+    // Pieces comprising the sector content, in order.
+    Pieces: []PieceActivationManifest,
+    KeepCommD: bool, // NEW
+}
+```
```
hunjixin, line 61:
```rs
struct PieceActivationManifest {
    PieceCid Cid     
    PieceSize u64  
    Keep: bool, // NEW  SP tell network to store or not
}
struct SectorActivationManifest {
    // Sector to be activated.
    Sector: SectorNumber,
    // Pieces comprising the sector content, in order.
    Pieces: []PieceActivationManifest,
}
```

hunjixin: @ZenGround0  Commd is hard to use, prefer to keep piece cids of data. however, considering the cost of on-chain storage, network can allow the SP to decide whether or not to store the piece of data. with this approach,  the storage fee can regulate storage pressure and user demands.. network must do some verify to ensure pieces correctly. 
1. get unseadcid from precommit info
2. calc commd from ``` []PieceActivationManifest```
3. check commd with precommit info
4.  keep pieces that SP want

hunjixin (approved):

FIPS/fip-on-chain-commd.md
ZenGround0: Thanks for reviewing @hunjixin.  Take a look at my arguments against using CommP here: https://github.com/filecoin-project/FIPs/discussions/760#discussioncomment-6597028.  I agree CommP is more directly useful.  I strongly disagree it belongs in the miner actor.

Kubuxu: I agree with @ZenGround0 position.

jennijuju: Agree with @ZenGround0 arguments - the design needs to be scalable.

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
+title: On Chain CommD
+author: <a list of the author's or authors' name(s) and/or username(s), or name(s) and email(s), e.g. (use with the parentheses or triangular brackets): @zenground0, Alex North (@anorth), ...
```
jennijuju, line 4: removed the text hint?

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/496, https://github.com/filecoin-project/FIPs/discussions/760
```
jennijuju, line 5: can we point to one primary discussion rather than two? otherwise it could be messy to leave and track peer review feedback imho

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+The filecoin development community has a demonstrated interest in enabling innovation in smart contracts that interact with verifiable storage.  It is the FVM's main differentiator when compared with other blockchain runtimes.  Two different groups have arrived at similar conclusions on a particular useful direction for the miner actor in relation to these goals:  myself [here](https://github.com/filecoin-project/FIPs/discussions/496) and venus dev team [here](https://github.com/filecoin-project/FIPs/discussions/760).  The first step in this direction is to allow SPs to keep CommD on chain in the sector info.
```
jennijuju, line 26: >  It is the FVM's main differentiator when compared with other blockchain runtimes.

im not sure why verifiable storage is a runtime feature rather than a system feature..?

jennijuju: > Two different groups have arrived at similar conclusions on a particular useful direction for the miner actor in relation to these goals:  myself [here](https://github.com/filecoin-project/FIPs/discussions/496) and venus dev team [here](https://github.com/filecoin-project/FIPs/discussions/760). 

 could you please summarize what does the current Filecoin protocol cant do that motivates this change instead? 
(we wanna know why an FIP is proposed based on the problem and solution, more than who has the cool idea.)

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+
+Adding CommD to sector info requires a sector info migration which on its own is a costly in terms of network development and risk.  Similarly it will require changes to parameters being introduced in ongoing [direct data onboarding (DDO) work](https://github.com/filecoin-project/FIPs/pull/804). Doing this work at the same time as DDO (which also needs a full sector info migration) makes the upgrade risk and development cost overhead much smaller.
```
jennijuju, line 28: This seems like it fits in design rationale  or an implementation suggestion.

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+    // Sector to be activated.
+    Sector: SectorNumber,
+    // Pieces comprising the sector content, in order.
+    Pieces: []PieceActivationManifest,
+    KeepCommD: bool, // NEW
+}
+```
+
+
+New ProveCommit and PRU messages will then parse this new bool flag and store CommD if specified.
```
jennijuju, line 64: Could you please use the FULL  method name here?  (im not the best person to call this out cuz I use acronym excessively lmao

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+If CommD is the zero data CommD then the system will always avoid storing even if KeepCommD is true.
```
jennijuju, line 65: peer review: have we considered reject message early that has KeepCommD = true yet is zero data commD? Im always a bit put off with the "set this flag to do a thing, but the thing wont necessarily happen even you set the right flag"

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Design Rationale
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+
+### Implementing CommD on chain
+Since CommD is a property of an individual sector if it is onchain it belongs on the sector info.
+
+Specifying inclusion as part of the DDO sector activation manifest allows for user control over which specific sectors can store CommD.
```
jennijuju, line 73: im not sure DDO sector is a term that everyone could understand.

FIPS/fip-on-chain-commd.md
```diff
@@ -0,0 +1,119 @@
...
+Sector infos are the datatype that dominate space used by the filecoin state and therefore blockchain syncing storage costs. Adding a potentially large field to this data type therefore has a risk of significantly increasing the data storage needed to sync the filecoin chain.  To understand this risk I measured sector info stats and reported findings [here](https://github.com/filecoin-project/FIPs/discussions/795).
+
+The results are encouraging.  CommR, which is the same size as CommD, takes up about 47% of all of sector info space.  Looking at the number of sectors that have deals vs total live sectors we see that all deal bearing sectors (living and dead) are < 14 percent of all currently live sectors.  So in the worst (for space usage) case we would see a 6-7% increase in space usage of state if we retroactively added CommD on chain for all existing non-CC sectors.  In practice the state impact will be lower. Only a fraction of users will opt into the new data onboarding flow adding CommD on chain and it will only apply to new deals.
```
jennijuju, line 103: Peer review: thanks for calling out the state size impact here. As you mentioned CommD is going to the same size as CommR. so if the data onboarded doubled, tripled in the next couple months, how would that impact filecoin state and snapshot sizes?

FIPS/fip-on-chain-commd.md
Fatman13: Hello, @ZenGround0,

> Ready for review
> cc @Fatman13

Thank you for opening this up!

Some high-level points discussed within Venus team wrt to your comment [here](https://github.com/filecoin-project/FIPs/discussions/760#discussioncomment-6597028). We probably should have brought up this much earlier but wasn't sure about how to go about it until the opt-in model proposed in this draft...

> 1. Too Expensive. 

The team thinks that cost could be paid by imposing a gas model to whoever opt-in to save commP, so that each sector is encouraged to have one or two pieceCid rather than to have a long list of pieceCid like in Web3 dot storage. 

> 2. CommP is not much more useful than CommD.

The team thinks that CommP could be directly used in a L2 contract to initiate storage deal disputes with a better user experience than CommD. Using CommD to dispute a deal might involve extensive on-chain analytics and history lookbacks which client may or may not capable of doing?

> 3. Leads to unscalable usage.

I think this maybe where the opt-in gas model coming into play, which may allow more diversed storage markets to be built, ie a higher premium for easier ux.

### High Level Rationale

Overall, apart from the exact technical route to achieve the stated goal of the FIP, the following are also worth considering...

- What really sets Filecoin apart from other storage chains? Given the current design route we are taking, could other chain or project, say eth.storage, achieve the same?
- Given that we want the minimal on-chain footprint, can we still provide features and have the invisible hand of gas to let market to reach equilibrium?
- A better UX for client?

anorth:
@ZenGround0 is currently on extended vacation, though we expect them to return.

With @jsoares in agreement, I am closing this as stale for now. Of course it can be re-opened in the future.

PR #838, opened 2023-09-27
FIP-0082: SuperSnaps -- Adding Snarkpack for SnapsDeals

cryptonemo:


cryptonemo:
cc: @anorth and @Kubuxu

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+|       3-8        |     48-128     | 86,500,000 |
+|       9-128      |    144-2048    | 92,500,000 |
+|     129-256      |   2064-4096    | 94,000,000 |
+|     257-512      |   4112-8192    | 99,500,000 |
+
+![Plot of observed and computed gas costs](../resources/fip-00XX/supersnap_gas_64gib.png)
+
+#### Batch Gas Charge
+
+Currently, **GasUsed * BaseFee** is burned for every message. We can achieve the requirements described in [Incentive Considerations](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#incentive-considerations) by charging an additional proportional gas cost to an aggregated batch of proofs, and by balancing their gas costs with a minimum gas fee. The gas charge for each aggregate Sector Update proof is computed in the same manner as that for aggregate seal proofs; see FIP-0013's [Batch Gas Charge](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#batch-gas-charge) for further discussion.
```
anorth, line 107: This part of FIP-0013 is written in a confusing way, but let's do better here. Please don't use the word "gas" for anything that is not consuming the scarce resource of blockchain computation. Just call anything else a "fee". I think the above should be "...charging an additional proportional fee to an aggregated batch of proofs, and by balancing their gas costs with a minimum fee. The additional charge for each ...."?

anorth, line 107: Please clarify whether the proof unit that is charged is replica update proof or Groth16 proof.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+
+```rust
+func PayBatchGasCharge(numProofsBatched, BaseFee) {
```
anorth, line 110: See also [FIP-0024](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0024.md) which changed numbers here.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+    // Cryptoecon Params (need to be updated if verification benchmarks change)
+    BatchDiscount = 1/20 unitless
+    BatchBalancer = 2 nanoFIL
+    SingleProofGasUsage = **TODO: should this be 36316136?**
```
anorth, line 114: Where did this number come from? I believe you want to use the cost of a single replica update proof

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+#### Proofs format 
+
+This FIP uses the same aggregate proof format introduced in FIP-0013's [Proofs Format](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#proofs-format).
+
+## Design Rationale
+
+The existing `ProveReplicaUpdates` method will not become redundant,
+since aggregation of smaller batches may not be efficient in terms of
+gas cost (proofs too big or too expensive to verify).  The method is
+left intact to support smooth operation through the upgrade period.
```
anorth, line 270: FIP-0076 ProveReplicaUpdates2 will support both batch and aggregation, so PRU will become redundant. I think all you need to say here is that it will keep supporting batch mode for small batches.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Failure handling
+
+Aborting on any precondition failure is chosen for simplicity.
+Submitting an invalid prove commitment should never happen for
+correctly-functioning miners.  Aborting on failure will provide a
+clear indication that something is wrong, which might be overlooked by
+an operator otherwise.
```
anorth, line 278: FIP-0076 is more friendly. Abort on aggregate proof validation failure, yes, but there are many other things involved in actual sector activation which this FIP has not talked about, and doesn't need to opine on re failure handling.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+A miner may submit multiple batches in a single epoch to grow faster.
+
+## Backwards Compatibility
+
+This proposal introduces a new exported miner actor method, and thus
+changes the exported method API.  While addition of a method may seem
+logically backwards compatible, it is difficult to retain the precise
+behaviour of an invocation to the (unallocated) method number before
+the method existed.  Thus, such changes must be delivered through a
+major version upgrade to the actors.
```
anorth, line 299: FIP-0076 again. But since it changes actor behaviour, it need a network upgrade. But the interface to ProveReplicaUpdates2 remains compatible.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+Aggregated proof verification allows for more sector update commitments to be proven in
+less time which will reduce processing time and therefore gas cost per prove
+replica update. Verification time and size of an aggregated proof scales logarithmically
+with the number of proofs being included.
+
+## Specification
+
+### Actor changes
+
+Add a new method `ProveReplicaUpdatesAggregated` which supports a miner
```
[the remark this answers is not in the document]
jennijuju: I won't suggest to add on a DDO dependency here as these two are two orthogonal proposals. Instead, I think it is worth considering to add a new method here similar to what 0076 is proposing - so whichever FIP lands first or both of them landing it will be compatible.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+
+The miner `ProveReplicaUpdates` method only supports committing a single sector update at
+a time.  This proposal adds a way for miners to post multiple `ProveReplicaUpdates`s at once in an aggregated fashion
+using a `ProveReplicaUpdatesAggregated` method. This method amortizes some of the costs
+across multiple sector data updates, and similar to [FIP-0013 (ProvedCommitSectorAggregated)](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md), aggregate proof verification runtime is incredibly fast due to SnarkPack. This proposal allows many sectors to have data updates in them activated at once when the aggregated message is posted to the chain.
+
+
+## Change Motivation
+
+The miner `ProveReplicaUpdates` method only supports committing a single sector at
```
jennijuju, line 33:
```suggestion
The miner `ProveReplicaUpdates` method only supports updating a single CC sector to data sector at
```

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+   AggregateProofType:          RegisteredAggregateProof
+   VerifiedAllocationClaims:    []{} // Must be empty
+   RequireActivationSuccess:    bool
+   RequireNotificationSuccess:  bool
+}
+```
+
+#### Scale and limits
+
+The minimum of number of Sector Update proofs that may be aggregated is 3, and the maximum is 512.
```
jennijuju, line 63: how much gas would it cost for a prove aggregate message that does 512 sectors? curious if it still falls under a single block limit

jennijuju: nvm found it below

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,345 @@
...
+
+Add a new method `ProveReplicaUpdatesAggregated` which supports a miner
+prove-committing a number of sector updates all at once.  The parameters for this
+method are:
+
+```rust
+struct ProveReplicaUpdateAggregateParams {
+   Sectors:                     []UpdateManifest
+   AggregateProof:              []byte // Before aggregation, concatenate proofs
+   UpdateProofType:             RegisteredUpdateProof
```
jennijuju, line 53: do we need a new proof type?

jennijuju (approved):

jennijuju:
one nit from an editor perspective + agree with anorth's title suggestion - otherwise the the draft lgtm

FIPS/fip-00xx.md
DrPeterVanNostrand: `36316136` is FVM's gas price for verifying a single replica update proof ([link](https://github.com/filecoin-project/ref-fvm/blob/4f2c73e96346cfa7364a51e133ffa906807df7c1/fvm/src/gas/price_list.rs#L193))

LaurenSpiegel:
@AxCortesCubero , any comments?

AxCortesCubero:
> @AxCortesCubero , any comments?

@LaurenSpiegel I'm sorry, can you clarify what is the question for me? Do you mean my thoughts on the Batch Balancer-like discussion above?

I would agree it seems the same motivation that there was at the time of implementing the batch balancer for Pre and Prove Commit seems to apply to this case: introduces a batching tool that makes gas usage for batches more efficient, but would need something like Batch Balancer to ensure this batching is only enabled when there is demand and it is needed. So a batch balancer seems reasonable (would have to carefully choose the parameters for a given goal).

What would be even better would be more permanent solutions like Explicit proof fees @Kubuxu has been leading, but this project has been somewhat paused. The best case scenario would be to finish carrying through the explicit fees idea, including this case.

FIPS/fip-00xx.md
anorth: I think the below is still wrong because it is not accounting for FIP-0025. E.g. the actor code uses a BatchBalancer = 5.

FIPS/fip-00xx.md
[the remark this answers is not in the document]
anorth: If this FIP lands first, the methods will not be compatible. They have fundamentally different data flows.

If you want to propose a concrete method in this FIP that could/would be made redundant by FIP-0076 and thus be omitted if they land together, ok. But the method specified in this FIP has to be correct for a non-FIP-0076 world, and it's not at the moment. 

It will be a total pain to implement both.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,343 @@
...
+   AggregateProofType:          RegisteredAggregateProof
+   VerifiedAllocationClaims:    []{} // Must be empty
+   RequireActivationSuccess:    bool
+   RequireNotificationSuccess:  bool
+}
```
anorth, line 56: This is a FIP-0076-like parameter schema, it's wrong for adding a new method now that follows existing workflows but is agregated. E.g. it needs to include Deal IDs, has no UpdateManifest (which was failed to be specified here), and doesn't have claims or any flags about requiring success. A new standalone method would instead look much more like the current ProveReplicaUpdate.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,343 @@
...
+category: Core
+created: 2023-09-26
+spec-sections: 
+  - section-systems.filecoin_mining.sector.lifecycle
+
+---
+
+## Simple Summary
+
+Add a method for a miner to submit several sector prove replica updates messages in a single one.
```
kaitlin-beegle, line 17:
```suggestion
Add a method for a miner to submit several sector prove replica updates in a single message.
```

anorth:
Assigning FIP-0082 to this. Please update the FIP metadata and filename, and add a link to this document in the root README.

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+
+## Change Motivation
+
+Adding aggregate proof verification to the method `ProveReplicaUpdates2` amortizes some of the costs associated with
+multiple sector updates, in the same way that [FIP-0013](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md)'s
+`ProveCommitAggregate` method has done for sealing proofs. While `ProveReplicaUpdate` is not one of the most frequent message
+on chain, it is believed to be too expensive to use in its current form.
+
+Aggregate proof verification allows for more sector update commitments to be proven in
+less time which will reduce the processing time, and thus the gas cost, per proven
```
jsoares, line 38:
```suggestion
less time, which will reduce the processing time and, thus, the gas cost, per proven
```

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+    // Whether to abort if any sector update activation fails.
+    pub require_activation_success: bool,
+    // Whether to abort if any notification returns a non-zero exit code.
+    pub require_notification_success: bool,
+}
+```
+
+#### Scale and limits
+
+The minimum of number of Sector Update proofs that may be aggregated is 3, and the maximum is 512.
```
jsoares, line 79:
```suggestion
The minimum number of sector update proofs that may be aggregated is 3, and the maximum is 512.
```

This is somewhat irrelevant, but the text capitalises "sector update" inconsistently. There's really no reason for it to be capitalised, so I'm going to suggest standardising on lowercase (see suggestions below...)

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+
+Aggregate proofs must be generated for a power of two number of Groth16 proofs; because each Sector Update proof contains sixteen Groth16 proofs, aggregated Sector Update proofs are padded to contain the nearest power of two number of Groth16 proofs. See the "Proof scheme changes" section in [FIP-0013](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#proof-scheme-changes) for a discussion on aggregate proof padding.
+
+The gas cost of aggregate proof verification grows both linearly and step-wise in the number of Sector Update proofs aggregated (counted prior to proof padding). The linear growth occurring for each Sector Update proof aggregated and the step-wise growth occurs when the number of Groth16 proofs aggregated exceeds a power of two.
+
+The gas table below associates ranges of aggregated Sector Update proofs with their total step-wise gas cost. The total gas cost for verifying an aggregate proof of `n` Sector Update proofs is the sum of `n`'s linear and step-wise gas costs: `n * gas_per_proof + gas_table(n)`.
```
jsoares, line 89:
```suggestion
Aggregate proofs must be generated for a power of two number of Groth16 proofs; because each sector update proof contains sixteen Groth16 proofs, aggregated sector update proofs are padded to contain the nearest power of two number of Groth16 proofs. See the "Proof scheme changes" section in [FIP-0013](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0013.md#proof-scheme-changes) for a discussion on aggregate proof padding.

The gas cost of aggregate proof verification grows both linearly and step-wise in the number of sector update proofs aggregated (counted prior to proof padding). The linear growth occurs for each sector update proof aggregated, and the step-wise growth occurs when the number of Groth16 proofs aggregated exceeds a power of two.

The gas table below associates ranges of aggregated sector update proofs with their total step-wise gas cost. The total gas cost for verifying an aggregate proof of `n` sector update proofs is the sum of `n`'s linear and step-wise gas costs: `n * gas_per_proof + gas_table(n)`.
```

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+
+##### 32 GiB Gas Cost
+
+Gas cost per Sector Update proof aggregated: 80,000 gas
+
+Total step-wise gas cost for ranges of Sector Update proofs aggregated:
```
jsoares, line 95:
```suggestion
Gas cost per sector update proof aggregated: 80,000 gas

Total step-wise gas cost for ranges of sector update proofs aggregated:
```

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+|      129-256     |   2064-4096    | 92,500,000 |
+|      257-512     |   4112-8192    | 99,000,000 |
+
+![Plot of observed and computed gas costs](../resources/fip-0082/supersnap_gas_32gib.png)
+
+##### 64 GiB Gas Cost
+
+Gas cost per Sector Update proof aggregated: 80,000 gas
+
+Total step-wise gas cost for ranges of Sector Update proofs aggregated:
```
jsoares, line 110:
```suggestion
Gas cost per sector update proof aggregated: 80,000 gas

Total step-wise gas cost for ranges of sector update proofs aggregated:
```

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+number of elements in the `sector_update_inputs` vector will be the
+`partitions per sector update * the number of sector update proofs aggregated`.
+
+**Requirements**: The scheme can only aggregate a power of two number of proofs
+currently. Although there might be some ways to alleviate that requirement, we
+currently pad the number of input proofs to match a power of two. Thanks to the
+logarithmic nature of the scheme, performance is still very much acceptable.
+
+Padding is currently _naive_ in the sense that if the passed in count
+of Sector Update proofs aggregated is not a power of 2 (greater than 1), we pad the Sector Update proofs to the nearest power of two using a duplicate of the last Sector Update proof provided. The minimum number of Groth16 proofs that can be aggregated is 2, thus in the case where a single Groth16 proof is provided for aggregation, the single proof is duplicated in order to pad the number of Groth16 proofs to 2.
```
jsoares, line 209:
```suggestion
Padding is currently _naive_ in the sense that if the passed in count
of sector update proofs aggregated is not a power of 2 (greater than 1), we pad the sector update proofs to the nearest power of two using a duplicate of the last sector update proof provided. The minimum number of Groth16 proofs that can be aggregated is 2, thus in the case where a single Groth16 proof is provided for aggregation, the single proof is duplicated in order to pad the number of Groth16 proofs to 2.
```

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+`sector_update_inputs` for verification as follows (pseudo-code for
+readability):
+
+```rust
+let sector_update_inputs: Vec<Vec<Fr>> = Vec::new();
+sector_update_inputs.extend(get_sector_update_inputs(..., comm_r_old_sector_one, ...));
+sector_update_inputs.extend(get_sector_update_inputs(..., comm_r_old_sector_two, ...));
+```
+
+What this example code does is flattens all of the individual proof
```
jsoares, line 262:
```suggestion
What this example code does is flatten all of the individual proof
```

FIPS/fip-0082.md
```diff
@@ -0,0 +1,357 @@
...
+sector update inputs into a single list, while properly maintaining
+the exact ordering matching the `sector_update_inputs` order going
+into `aggregate_empty_sector_update_proofs`.  When compiled like this,
+the `sector_update_inputs` will be in the exact format required for
+the `verify_aggregate_sector_update_proofs` API call.
+
+Similar to aggregation, padding for verification is currently also
+_naive_. If the passed in count of proof input sets (while noting that
```
jsoares, line 270:
```suggestion
_naive_. If the passed-in count of proof input sets (while noting that
```

jsoares (approved):

cryptonemo:
Thanks for the approvals (all reviewers so far)!  I'm unable to merge this, so I'll have to leave that to someone else if and when it's ready.

PR #852, opened 2023-10-18
FIP-0076 Add Market::GetDealSector method

anorth:
Implementation is https://github.com/filecoin-project/builtin-actors/pull/1443

### DIFF
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -542,6 +542,25 @@ struct OnMinerSectorsTerminateParams {
 The sector is removed from the `ProviderSectors` mapping, if present.
 Any deals mapped to that sector are marked as terminated, and subsequent processing deferred to cron.
 
+#### GetDealSector
+
+A new FRC-0042 exported method "GetDealSector" returns the sector in which a deal's data is stored,
+while a deal is active.
+If a deal is published but not yet activated, aborts with `EX_DEAL_NOT_ACTIVATED = 33`.
+If a deal is not found, aborts with the same exit codes as "GetDealActivation".
+
+```
+// "Transparent" CBOR-encoding, the singleton field is encoeded directly.
+struct GetDealSectorParams {
+    DealID: DealID,
+}
+
+// "Transparent" CBOR-encoding, the singleton field is encoeded directly.
+struct GetDealSectorReturn {
+    SectorNumber: SectorNumber,
+}
+```
+
 ### Migration
 
 The built-in market actor's `ProviderSectors` mapping is initialised from the existing deal state
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -544,8 +544,8 @@ Any deals mapped to that sector are marked as terminated, and subsequent process
 
 #### GetDealSector
 
-A new FRC-0042 exported method "GetDealSector" returns the sector in which a deal's data is stored,
-while a deal is active.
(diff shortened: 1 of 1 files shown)


FIPS/fip-0076.md
```diff
@@ -542,6 +542,25 @@ struct OnMinerSectorsTerminateParams {
 The sector is removed from the `ProviderSectors` mapping, if present.
 Any deals mapped to that sector are marked as terminated, and subsequent processing deferred to cron.
 
+#### GetDealSector
+
+A new FRC-0042 exported method "GetDealSector" returns the sector in which a deal's data is stored,
+while a deal is active.
```
jennijuju, line 548: What if the deal is already expired, the the sector is still live? (Sector lifetime is greater than deal duration and  data is still stored after deals expiration

jennijuju:
How useful is this api without a `GetSectorInfo` like api? One use case I can think of is using this to figure out when the data will actually mo longer be stored in a sector and trigger a renew deal when it’s close to that time, after finding out the sector i will need to further getting the sector start epoch & duration to get the expiration of the sector

FIPS/fip-0076.md
```diff
@@ -542,6 +542,25 @@ struct OnMinerSectorsTerminateParams {
 The sector is removed from the `ProviderSectors` mapping, if present.
 Any deals mapped to that sector are marked as terminated, and subsequent processing deferred to cron.
 
+#### GetDealSector
+
+A new FRC-0042 exported method "GetDealSector" returns the sector in which a deal's data is stored,
```
jennijuju, line 547:
```suggestion
A new FRC-0042 exported method "GetDealSector" (method number 2611213344) returns the sector in which a deal's data is stored,
```

FIPS/fip-0076.md
arajasek: I think the text is clear here -- such a deal is not "active", and so would fall into `If a deal is not found, aborts with the same exit codes as "GetDealActivation".` (which would return `EX_DEAL_EXPIRED`).

FIPS/fip-0076.md
arajasek, line 547:
```suggestion
A new FRC-0042 exported method "GetDealSector" returns the sector number in which a deal's data is stored,
```

arajasek: or SectorID or some such?

arajasek (approved):

FIPS/fip-0076.md
anorth: Yes, sector number, thanks. I'm only intending to return this because the provider ID is available from the deal itself (which the client proposed, or some other party can look up by deal ID).

anorth:
> How useful is this api without a GetSectorInfo like api? One use case I can think of is using this to figure out when the data will actually mo longer be stored in a sector and trigger a renew deal when it’s close to that time

The only thing a client should rely on to understand when data will no longer be stored is the deal's duration. The fact that today one can, externally, tell how much longer the data is committed is an artefact of our very limited system that prevents replacing stale data.

I don't know exactly how useful this API will be. Most likely, today, it could be useful in concert with direct state inspection from off-chain. This is one step towards reducing the need for direct state inspection. I don't really think this step should be blocked on having not taken all the other ones at the same time, but I'll just as happily retract this change.

kaitlin-beegle (approved):

PR #859, opened 2023-11-01
FIP0079: Finalize gas prices

Stebalien:
After re-benchmarking, the pairing price was spot-on but the per-byte cost was an overestimate (likely due to recent optimizations in the libraries we're using, more than anything else).


Stebalien:
cc @DrPeterVanNostrand

anorth (approved):

jennijuju (approved):

PR #862, opened 2023-11-03
On-chain Retrieval Expectations

willscott:
Discussion: [#861](https://github.com/filecoin-project/FIPs/discussions/861)
Rendered version: [here](https://github.com/willscott/FIPs/blob/willscott/retrieval-tiers/FIPS/fip-00xx.md)

This FIP proposes a set of retrieval SLA ‘tiers’. These tiers identify the ‘category’ of data - if it is fully offline, meant for low-volume-retrieval archival usage, or for higher bandwidth activity. The consensus part of this FIP is a proposal to encode the retrieval SLA tier as part of a deal proposal. Standardizing retrieval expectations in this way allows storage providers to apply appropriate policies and pricing to deals.

### DIFF
--- a/FIPS/fip-00xx.md
+++ b/FIPS/fip-00xx.md
@@ -4,7 +4,7 @@ title: On-chain Retrieval Expectations
 author: "Will Scott (@willscott)"
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/
 status: draft
-type: Technical (Core)
+type: Technical
 category: Core
 created: 2023-10-10
 ---
--- a/FIPS/fip-00xx.md
+++ b/FIPS/fip-00xx.md
@@ -1,0 +1,155 @@
+---
+fip: "**TODO**"
+title: On-chain Retrieval Expectations
+author: "Will Scott (@willscott)"
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/
+status: draft
+type: Technical (Core)
+category: Core
+created: 2023-10-10
+---
+
+<!--You can leave these HTML comments in your merged FIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new FIPs. Note that a FIP number will be assigned by an editor. When opening a pull request to submit your FIP, please use an abbreviated title in the filename, `fip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+Today, a mish-mash of heuristics have emerged to attempt to identify client intentions about the retrievability of data stored in filecoin.
+This FIP proposes a simple on-chain mechanism to allow clients to express those expectations at deal time.
+This proposal should be thought of as two distinct sub-components: A concretization of the expectations that clients have about the retrievability of data into a defined set of SLAs, and a mechanism to express those SLAs on-chain.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+
+This FIP defines a set of retrieval SLA ‘tiers’.
+These tiers identify the ‘category’ of data - if it is fully offline, meant for low-volume-retrieval archival usage, or for more regular retrieval.
+The consensus part of this FIP is a proposal to encode the retrieval SLA tier as part of a deal proposal.
+Standardizing retrieval expectations in this way allows storage providers to apply appropriate policies and pricing to deals.
+
+## Change Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+
+There are already several fields which touch on this issue. 
+
+* `FastRetrieval`  - part of `ClientStartDeal` API calls and stored in provider state.
+
+* `Label`  - part of a `DealProposal` on chain.
+
+* `VerifiedDeal` - part of `DealProposal` on chain
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+### Overview
+
+### Tiers
+
+The proposed tiers are:
+
+1. *Offline* - Data that is transferred via physical media. This data is not expected to be retrieved over the network.
+2. *Archival* - Data that is expected to be retrieved infrequently. This data is only expected to need enough network bandwidth to be able to ensure replication. The most common data replication policies on filecoin today at 5x and 10x copy replication, to ensure that data remains even when some copies are unavailable. In order for contracts or users to ‘heal’ data replication policies in the face of failures, they would potentially need (n-1) copies of the data to be re-replicated.
+3. *Online Data* - Data that is expected to have periodic retrieval by an ACLed list of users. This tier allows for negotiation of higher retrieval service levels without committing to make the data available to the network at large.
+4. *Public* - Data that is expected to be openly retrievable by anyone. This tier allows for negotiation of higher bandwidth and low latency provisioning.
+5. *CDN* - The details of this tier are still being worked out, but we reserve this nomenclature for a higher tier of retrieval service that equates to the expectations for bandwidth and latency that would be found at CDN service offerings.
+
+| Tier | Offline | Archival | Online | Public |
+| --- | --- | --- | --- | --- |
+| Example of data | Large datasets onboarded with physical disks | Long term private data storage & backups | Research Datasets | Files, Documents, Images |
+| Amortized anticipated retrieval volume | 0 | 1x / month | 1x / week | 1x / day |
+| Expected resilience to burstiness | n/a | 2x | 3x | 5x |
+| User-expected Latency | n/a | < 6hr | < 1min | < 5sec |
+
+
+
+### Deal flow change (current)
+
+The `DealProposal` [struct](https://github.com/filecoin-project/go-state-types/blob/master/builtin/v12/market/deal.go#L200) is extended to include an additional, optional, field:
+
+```go
+    RetrievalTier abi.RetrievalTier
+```
+
+An undefined or empty value of this field is interpreted using the following logic: 
+
+* If the `VerifiedDeal` flag is set, the tier interpreted as `Archival` 
+
+* Otherwise the tier is interpreted as `Offline`
+
+### Deal flow change (in conjunction with direct data onboarding)
+
+In the network with Direct Data Onboarding (FIP 0076), there will not always be a `DealProposal` for a deal.
+In this case, where no market deal is present, the `RetrievalTier` is encoded as an extension of the `PieceActivationManifest` from the storage provider.
+
+The negotiation of storage deals may contain a client request for which Retrieval Tier to specify, but that process in Direct Data Onboarding occurs off chain.
+What inclusion of this field in the `PieceActivationManifest` allows is that the chain may react and that contracts can be written to monitor and respond to these advertisements.
+
+### FVM extensions
+
+`publishStorageDeals` will retain its existing signature, and will take on the ‘empty’ definition of the `RetrievalTier` field as described in (“Tiers”)[#tiers].
+A parallel method can be implemented allowing for the additional specification of the `RetrievalTier`.
+
+A new method `GetDealRetrievalTier` would be added as with other deal property accessors.
+
+## Design Rationale
+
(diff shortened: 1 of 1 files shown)


FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
+---
+fip: "**TODO**"
+title: On-chain Retrieval Expectations
+author: "Will Scott (@willscott)"
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/
+status: draft
+type: Technical (Core)
```
jennijuju, line 7:
```suggestion
type: Technical
```

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+
+This FIP defines a set of retrieval SLA ‘tiers’.
+These tiers identify the ‘category’ of data - if it is fully offline, meant for low-volume-retrieval archival usage, or for more regular retrieval.
+The consensus part of this FIP is a proposal to encode the retrieval SLA tier as part of a deal proposal.
+Standardizing retrieval expectations in this way allows storage providers to apply appropriate policies and pricing to deals.
+
+## Change Motivation
```
anorth, line 29: This section is lacking describing the motivation for any consensus change here.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+
+There are already several fields which touch on this issue. 
+
+* `FastRetrieval`  - part of `ClientStartDeal` API calls and stored in provider state.
+
+* `Label`  - part of a `DealProposal` on chain.
+
+* `VerifiedDeal` - part of `DealProposal` on chain
```
anorth, line 38: Can you expand on how these touch the issue? `FastRetrieval` isn't part of consensus protocol AFAIK so is a good example of FRC-type data here. `Label` is opaque at the consensus level, but can support conventions being built on it in FRCs. `VerifiedDeal` is an instruction to the built-in market actor to act as an operator for the client's datacap and make a verified allocation concurrent with publishing a deal, but I don't see the relevance to this proposal.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+
+
+
+### Deal flow change (current)
+
+The `DealProposal` [struct](https://github.com/filecoin-project/go-state-types/blob/master/builtin/v12/market/deal.go#L200) is extended to include an additional, optional, field:
+
+```go
+    RetrievalTier abi.RetrievalTier
+```
```
anorth, line 70: This change implies a bunch more changes. The DealProposal is part of built-in market actor API and state. 
- Changing the API (the schema of the struct it will accept) is disruptive. From text below, I think you're suggesting that there be a new DealProposal2 type with a new field that can be used in the API for a new PublishStorageDeals2 method, in order to preserve compatibility
- Changing the state schema requires a migration of the ~50M deals in state (or an alternate design for multiple collections, or something).

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+An undefined or empty value of this field is interpreted using the following logic: 
+
+* If the `VerifiedDeal` flag is set, the tier interpreted as `Archival` 
+
+* Otherwise the tier is interpreted as `Offline`
+
+### Deal flow change (in conjunction with direct data onboarding)
+
+In the network with Direct Data Onboarding (FIP 0076), there will not always be a `DealProposal` for a deal.
+In this case, where no market deal is present, the `RetrievalTier` is encoded as an extension of the `PieceActivationManifest` from the storage provider.
```
anorth, line 81: This is also quite under-specified.

The whole design intent behind a piece activation manifest is that the consensus layer need know nothing about markets and deals etc. There is an opaque payload field in order to pass information through to a market contract but, like deal label, it's completely up to market contracts to interpret. I can't tell if you mean to specify some format of data in that payload, or put a field into the consensus-level schema. If it's the latter, I will object (but can take it to the discussion topic).

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+
+The negotiation of storage deals may contain a client request for which Retrieval Tier to specify, but that process in Direct Data Onboarding occurs off chain.
+What inclusion of this field in the `PieceActivationManifest` allows is that the chain may react and that contracts can be written to monitor and respond to these advertisements.
+
+### FVM extensions
```
anorth, line 86: This section doesn't have much to do with the FVM, and should probably just be folded into the main actor change specification.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+The tier definitions here provide a shorthand encoding of common levels of service that can be efficiently communicated on chain, and which measurement and evaluation services can be designed against.
+Additional agreements can take place in smart contracts on an opt-in basis to provide more complex policy agreements.
+Since we cannot predict the needed common language for these policies in a way for arbitrary other components to enforce against, we defer that to future work.
+
+## Backwards Compatibility
+<!--All FIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The FIP must explain how the author proposes to deal with these incompatibilities. FIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+
+The added field, both in and out of a direct data onboarding network, is optional. As such, any current deal, and any future deal not specifying the added field will take on the specified default behavior.
```
anorth, line 104: Adding an optional field is not a backwards compatible change to actor APIs. As noted above, this probably needs additional methods and a state migration.

anorth (commented):
Overall I think that this topic should be an FRC, but we can take that to the discussion threads.

I've attempted part of an editorial review on the assumption that it really needs a core protocol FIP, but overall this is missing quite a bit of substance in the specification. When that is firmed up, we'll better be able to describe what ought to be covered in the "considerations" sections.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+category: Core
+created: 2023-10-10
+---
+
+<!--You can leave these HTML comments in your merged FIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new FIPs. Note that a FIP number will be assigned by an editor. When opening a pull request to submit your FIP, please use an abbreviated title in the filename, `fip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+Today, a mish-mash of heuristics have emerged to attempt to identify client intentions about the retrievability of data stored in filecoin.
```
kaitlin-beegle, line 17:
```suggestion
Today, a mish-mash of heuristics have emerged to attempt to identify client intentions about the retrievability of data stored on Filecoin. 
```

Nit update to be more aligned with what appears to be convention.

FIPS/fip-00xx.md
kaitlin-beegle, line 29: Please elaborate on how encoding these SLAs at the protocol level will affect Filecoin functionality in the future, and/or address concrete issues in the present. 

Why should retrieval expectations become policy at the protocol level, as opposed to incorporated as more arbitrary (i.e., optional) deal metadata? Why should SLA tiers be a required component of deal-making for all network participants?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+
+This FIP does not itself change incentives. It does provide a hook by which providers can set different pricing and policies for different tiers of service.
+This is a necessary component for enabling a market for different tiers of service. There is sufficient flexibility in other structures built on top of filecoin that we can expect some amount of incentive shift to occur as a result of this FIP.
+The expected direction of this shift is towards an increased incentivization of retrievals.
+
+### Gameability of tiers
+
+Reporting that content is available at a higher tier does not directly help clients or storage providers.
+In fact, it primarily raises costs as the sum of all deal retrievals will lead to a direct calculation of expected network bandwidth provisioning of an SP.
```
kaitlin-beegle, line 126: Question- where does the 'direct calculation of expected network bandwidth provisioning' take place? Is this a business hypothetical, or a concrete part of the deal-making process/fee calculation that I'm unaware of?

More frequently retrieved data of course has higher bandwidth requirements during retrieval, but does this anticipated bandwidth use actually get passed on to the SP or deal client during deal making? Is there a higher gas fee, etc., charged to deals that are likely to require greater network bandwidth, etc.?

kaitlin-beegle:
> Overall I think that this topic should be an FRC, but we can take that to the discussion threads.
> 
> I've attempted part of an editorial review on the assumption that it really needs a core protocol FIP, but overall this is missing quite a bit of substance in the specification. When that is firmed up, we'll better be able to describe what ought to be covered in the "considerations" sections.

FWIW, this was also my comment to @willscott after seeing his team's presentation during LabWeek.  I've asked him to add more information about why these flags should be a network prerogative, rather than a standard.

kaitlin-beegle (approved):
There seems to be interest amongst FIP Editors to be more diligent about merging earlier stage FIP drafts.  For that reason, I'm happy to approve this for now. 

However, there are clearly a lot of sessions that have yet to be fleshed out.  I also think there is a big question still to be answered about why this change ought to be encoded within the protocol, and what future functionality or product changes will be enabled by doing so.

FIPS/fip-00xx.md
jennijuju: +1 to above.

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+Today, a mish-mash of heuristics have emerged to attempt to identify client intentions about the retrievability of data stored in filecoin.
+This FIP proposes a simple on-chain mechanism to allow clients to express those expectations at deal time.
+This proposal should be thought of as two distinct sub-components: A concretization of the expectations that clients have about the retrievability of data into a defined set of SLAs, and a mechanism to express those SLAs on-chain.
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
```
jennijuju, line 22: Is this change specific to f05 market deals, or intended to set a standard for future user programmed storage markets?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+### Overview
+
+### Tiers
+
+The proposed tiers are:
+
+1. *Offline* - Data that is transferred via physical media. This data is not expected to be retrieved over the network.
+2. *Archival* - Data that is expected to be retrieved infrequently. This data is only expected to need enough network bandwidth to be able to ensure replication. The most common data replication policies on filecoin today at 5x and 10x copy replication, to ensure that data remains even when some copies are unavailable. In order for contracts or users to ‘heal’ data replication policies in the face of failures, they would potentially need (n-1) copies of the data to be re-replicated.
```
jennijuju, line 50: For my own understanding- where  does 5x/10x coming from? Fil+?

FIPS/fip-00xx.md
```diff
@@ -0,0 +1,155 @@
...
+What inclusion of this field in the `PieceActivationManifest` allows is that the chain may react and that contracts can be written to monitor and respond to these advertisements.
+
+### FVM extensions
+
+`publishStorageDeals` will retain its existing signature, and will take on the ‘empty’ definition of the `RetrievalTier` field as described in (“Tiers”)[#tiers].
+A parallel method can be implemented allowing for the additional specification of the `RetrievalTier`.
+
+A new method `GetDealRetrievalTier` would be added as with other deal property accessors.
+
+## Design Rationale
```
jennijuju, line 93: First, this retrieval expectation cannot be enforced onchain, just like `fastretrieval`. There have been many discussion on having `fastretrieval ` in the protocol sends wrong expectation and should be removed, therefore I’m not sure adding another explicit field similarly is a good idea.

second, have you considered to propose an FRC for setting such “tag” in the `label` field instead? Im trying to guess the motivation of the fip here as its under specified currently, and it seems like one of the main goal is to allow data client  to signal retrieval expectation? If so, I think adding in the label is worth considering. However, if the goal is also providing retrieval guarantees, then this fip needs more work on that.

jsoares (commented):
I don't have much to add to the comments already provided by other reviewers but I, too, question why this needs to be handled at the consensus level, given the added tiers are neither specific enough nor self-enforceable. As it stands, I don't see solid reasoning for this path articulated in either the current draft or the related discussion, and it seems like the type of thing that would cleanly fit an FRC.

jsoares:
Closing this given no activity in the last year. Can be reopened at any time.

PR #877, opened 2023-12-04
Revert "FIP-0076: Add ListAllocations/ListClaims verified registry methods. (#865)"

anorth:
These methods are insufficiently motivated, and introducing them raises broader design questions which are out of scope for this FIP. These can be re-introduced in a new FIP with more motivation and context.

See [discussion](https://github.com/filecoin-project/FIPs/discussions/730#discussioncomment-7654363).


jsoares (approved):

raulk (approved):

raulk:
(To reiterate, I think these would be fine as private methods if needed to encapsulate observability logic.)

PR #878, opened 2023-12-07
Add missing title to frc-0058.md

alanshaw:


anorth:
A title here is not part of the template. I do see a few documents that do have such a title, but the majority don't.

@filecoin-project/fips-editors let's pick one format or the other and update all FIPs to match (and the template, if we add the header).

jsoares:
I always favour less duplication, so am happy with having the title only in the metadata block. Weakly held opinion, though.

alanshaw:
🤷 This document looks like it is missing it's main title:

Screenshot 2023-12-07 at 17 31 23

My eyes skipped over the metadata section entirely and I did not know what the FRC was about.

This document looks correct:

Screenshot 2023-12-07 at 17 31 55

I would prefer to see a big title than a small one that is hard to pick out and is not as "important" visually as headings for other sections.

Either way I agree that consistency would be good.

jennijuju:
I personally think a title in header is nice as it "speaks out" what's the doc about clearly. Also very weak opinion here tho.

@luckyparadise could you please make a decision here and help updating the existing FIPs & templates for consistency?

raulk (approved):

jsoares:
> It's complementary and omitting it should not cause the loss of important content, like the document's title.

It's a slippery slope. The FIP number is currently only in the metadata; should it also be in the h1?

raulk:
I don't think the FIP number is integral to understanding the content. I wouldn't display the IBAN of a book prominently on the cover, but it'd be a problem if the title was not there, and only appeared in e.g. the copyright page (metadata in that context).

anorth:
I'm fine with either decision so let's add the title. 

I'm not approving this PR, but will approve one that adds it to the template and updates all the other ones too.

PR #883, opened 2023-12-12
Fix FIP number in FIPS 0078/0079/0081

jsoares:
Some recent FIPS have departed from the metadata format and included the string "FIP-" in the FIP number, resulting in inconsistent rendering on fips.filecoin.io. PR removes said strings, but it's something for us to watch out for in the future.

image


luckyparadise:
Thank you for flagging this @jsoares!

anorth (approved):

jennijuju (approved):

PR #884, opened 2023-12-13
Add titles

alanshaw:
Per https://github.com/filecoin-project/FIPs/pull/878#issuecomment-1850882270 this PR updates the template and adds titles to all FIPs that did not already have them.


jsoares (approved):
Still not a huge fan of the principle, but the result looks consistent. Thanks for doing this!

anorth (approved):

PR #896, opened 2023-12-20
FIP-0086: Fast Finality in Filecoin

jsoares:
Discussion: #809 

The current Filecoin consensus mechanism only provides probabilistic finality. To simplify client implementations and provide some form of determinism, the protocol also includes a soft finality threshold, whereby miners at round _N_ reject all blocks that fork off before _N-900_. This finalization delay of 900 epochs (7.5 hours) hinders user experience and limits applications built on Filecoin.

We specify a mechanism for fast finality with the F3 component. F3 is expected to finalize tipsets within tens of seconds during regular network operation, compared to the current 900-epoch finalization delay. It does so by leveraging GossiPBFT, an optimally resilient partially synchronous BFT consensus protocol, which runs in parallel with the current protocol, taking EC tipsets as input and providing finalized prefixes as output. EC’s fork choice rule is modified never to select away from the F3-finalized chain.

See also the summary of the implementers' sync (https://github.com/filecoin-project/core-devs/discussions/148#discussioncomment-7858442), where it was agreed to consider a sooner-than-nv23 F3-only upgrade. We're still working out a concrete timeline for potential deployment if accepted.

There are still open questions listed in the TODO section, none of which we think are blocking at the draft stage.

cc Jie Hou (@mb1896), Alex North (@anorth), Henrique Moniz (@hmoniz), Matej Pavlovic (@matejpavlovic), Aayush Rajasekaran (@arajasek), Alejandro Ranchal-Pedrosa (@ranchalp), Jorge Soares (@jsoares) and Marko Vukolic (@vukolic)


FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+42:       evidence ← m.evidence \* strong PREPARE quorum is inherited evidence
+43:     else \* no participant decided in this round
+44:       evidence ← Aggregate(M) \* strong quorum of COMMITs for 丄 is evidence
+45:     round ← round + 1;
+46:     timeout ← updateTimeout(timeout, round)
+47:   }  \*end while
+
+48:   collect a clean set M of DECIDE messages
+      until (HasStrongQuorumValue(M)) \* Collect a strong quorum of decide outside the round loop
+49:   return (StrongQuorumValue(M))
```
matejpavlovic, line 405:
```suggestion
49:   return (StrongQuorumValue(M), Aggregate())
```

This must have been just a mistake when porting the document between various formats. The `Aggregate(M)` return value is required as a `PoF`. The function signature already says so.

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
+---
+fip: "<to be assigned>"
+title: Fast Finality in Filecoin
```
jennijuju, line 3:
```suggestion
title: Fast Finality in Filecoin (F3)
```

we refer this to F3 quite a lot, adding it to the title so the term `F3` is  more well known

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+author: Jie Hou (@mb1896), Alex North (@anorth), Matej Pavlovic (@matejpavlovic), Aayush Rajasekaran (@arajasek), Alejandro Ranchal-Pedrosa (@ranchalp), Jorge Soares (@jsoares) and Marko Vukolic (@vukolic)
```
jennijuju, line 4: whats about `@hmoniz`?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+
+## Abstract
+
+The current Filecoin consensus mechanism only provides probabilistic finality. To simplify client implementations and provide some form of determinism, the protocol also includes a soft finality threshold, whereby miners at round $N$ reject all blocks that fork off before $N-900$. This finalization delay of 900 epochs (7.5 hours) hinders user experience and limits applications built on Filecoin.
+
+We specify a mechanism for fast finality with the F3 component. F3 is expected to finalize tipsets within tens of seconds during regular network operation, compared to the current 900-epoch finalization delay. It does so by leveraging GossiPBFT, an optimally resilient partially synchronous BFT consensus protocol, which runs in parallel with the current protocol, taking EC tipsets as input and providing finalized prefixes as output. EC's fork choice rule is modified never to select away from the F3-finalized chain.
```
jennijuju, line 23:
```suggestion
We specify a mechanism for fast finality with the F3 component. F3 is expected to finalize tipsets within tens of seconds during regular network operation, compared to the current 900-epoch finalization delay. It does so by leveraging GossiPBFT, an optimally resilient partially synchronous BFT consensus protocol, which runs in parallel with the current protocol, taking EC tipsets as input and providing finalized prefixes as output. The fork choice rule of EC is modified to ensure it never selects a chain other than the F3-finalized chain.
```

the senescence was reading a bit awkwardly for me.. and here's the chatgpt's suggestion 😸

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+
+## Change Motivation
+
+* The long time to finality on Filecoin mainnet restricts or severely affects applications built on Filecoin (e.g., IPC, FVM, Axelar, Wormhole, Glif, …).
```
jennijuju, line 28:
```suggestion
* The long time to finality on Filecoin mainnet restricts or severely affects applications built on Filecoin (e.g., IPC, Axelar, Wormhole, Glif, …).
```

removing given FVM itself is not an application (weird branding side effects..)

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+
+## Specification
+
+
+### Background
+
+[Expected Consensus (EC)](https://spec.filecoin.io/algorithms/expected_consensus/) is the current mechanism by which participants in the Filecoin network reach an agreement on tipsets. A tipset is a set of blocks with the same epoch and the same set of parents. EC is a longest-chain protocol (more accurately, a heaviest-chain protocol) in which each participant independently builds the chain as it receives blocks from the network. Time is divided into slots of 30 seconds, called _epochs_. In each epoch, the protocol elects a set of network participants (i.e., storage providers) to become proposers. Each proposer can construct a new block and broadcast it to the network. On reception, each participant appends the block to its local view of the blockchain. Each tipset has a _weight_ corresponding to the total number of blocks in the path between the genesis and the tipset (the actual weight function is slightly more complex in reality, but this approximation is sufficient for this document). An example blockchain data structure is shown below, indicating the weight of each tipset in parentheses.
```
jennijuju, line 39:
```suggestion
[Expected Consensus (EC)](https://spec.filecoin.io/algorithms/expected_consensus/) is the current mechanism by which participants in the Filecoin network reach an agreement on tipsets. A tipset is a set of blocks with the same epoch and the same set of parents. EC is a longest-chain protocol (more accurately, a heaviest-chain protocol) in which each participant independently builds the chain as it receives blocks from the network. Time is divided into slots of 30 seconds, called _epochs_. In each epoch, the protocol elects a set of network participants (i.e., storage providers) to become block proposers. Each proposer can construct a new block and broadcast it to the network. On reception, each participant appends the block to its local view of the blockchain. Each tipset has a _weight_ corresponding to the total number of blocks in the path between the genesis and the tipset (the actual weight function is slightly more complex in reality, but this approximation is sufficient for this document). An example blockchain data structure is shown below, indicating the weight of each tipset in parentheses.
```

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+
+### F3 Overview and Interaction with EC
+
+We propose implementing fast finality in Filecoin by introducing an F3 component, which works alongside EC in Filecoin client nodes (participants).
+
+The participants in F3 are the storage providers (SPs) of the Filecoin network. The participation of each SP is weighted according to its quality-adjusted power (QAP), which is a function of the storage that the SP has committed to the network. This information is maintained in a system actor called the _power table_.
```
jennijuju, line 52:
```suggestion
The participants in F3 are the storage providers (SPs) of the Filecoin network. The participation of each SP is weighted according to its quality-adjusted power (QAP), which is a function of the storage power that the SP has committed to the network. This information is maintained in a system actor called the _power actor_.
```

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+In short, each participant $p$ in the Filecoin network (i.e., a storage provider with power) runs F3 in a loop and feeds it input from EC. F3 returns a finalized prefix chain to EC, which updates the canonical chain to extend this F3-finalized prefix. More precisely, in each loop iteration $i$ (corresponding to the i-th instance of F3):
```
jennijuju, line 54:
```suggestion
In short, each participant $p$ in the Filecoin network (i.e., a storage provider with power) runs F3 in a loop and feeds its input from EC. F3 returns a finalized prefix chain to EC, which updates the canonical chain to extend this F3-finalized prefix. More precisely, in each loop iteration $i$ (corresponding to the i-th instance of F3):
```

jennijuju, line 54: can we expand on what does _F3 finalized_ mean?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+- **EC/F3 interface:** Participant $p$ feeds its current canonical chain $canonical$ and its previously F3-finalized chain, which we call the $baseChain$, to F3. The $baseChain$ defines the power table and seeds randomness used to configure the F3 instance, while $p$'s current canonical chain is the chain $p$ proposes to be finalized. Periodically, $p$ also feeds to F3 tipset updates from EC that extend $p$'s canonical chain as EC delivers these tipsets.
```
jennijuju, line 56: >Periodically, $p$ also feeds to F3 tipset updates from EC that extend $p$'s canonical chain as EC delivers these tipsets.

for my understanding, does this happen every 30s, in the beginning/end of each epoch?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+- **F3 consensus:** F3 establishes network-wide consensus on finalized tipsets and produces a _Proof of Finality (PoF)_ for a tipset finalized in instance $i$, $t_i$. Every _PoF_ output by F3 is signed by ≥ ⅔ of the total QAP, i.e., a super-majority of the power table vouching that honest participants with more than ⅓ QAP consider tipset $t_i$ final.
```
jennijuju, line 57: where is the total QAP taken from, the base chain?

jennijuju, line 57: > .. that honest participants with more than ⅓ QAP consider tipset $t_i$ final.

should it be ⅔ if I werent mistaken?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+- **EC:** Participant $p$ updates its local EC chain and commits not to reorganize the finalized tipset $t_i$, i.e., the tipset must stay in $p$'s EC canonical chain for good. Apart from this change, EC continues operating as it currently does. In particular, EC still does a 900-epoch lookback for its power table and continues operating “normally” if F3 assumptions are violated and F3 halts, with the combined EC/F3 protocol favoring availability over consistency (in CAP theorem parlance).
+- **F3 synchronization:** Participants disseminate information about finalized tipset $t_i$ and its proof $PoF_i$ to all other participants, light clients, and smart contracts. The main goal of F3 synchronization is to allow an external party (or a lagging F3 participant) to fetch a sequence of messages that prove the finality of some recent final tipset. The sequence demonstrates a chain of eligible participants deciding on a final tipset and, thus, the eligible power table for the next round. Verifying the finality of a tipset from genesis does not require validating the EC chain.
+
```
jennijuju, line 60: are selected block producers for each round are required to participate in BOTH ec & f3 protocol at the same time? or EC & F3 participants at the same epoch two separate set?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+![](../resources/fip-xxxx/flow.png)
+
+### F3 Adversarial Model
+
+Honest participants follow the protocol at all times. Byzantine participants deviate arbitrarily from the protocol. Rational participants deviate from the protocol to pursue a greater (expected) pay-off according to their utility function. We consider a Byzantine adversary able to control and orchestrate up to less than ⅓ QAP.
```
jennijuju, line 65: to confirm, for any participants thats has equal or larger than 1/3 QAP are disqualified from participating in F3?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+F3 uses GossipSub (like Filecoin EC) to disseminate protocol messages, implementing a broadcast channel among participants. We assume and model GossipSub to satisfy the following properties of the _best-effort broadcast primitive_ (denoted _BEBroadcast_ hereafter):
+
+* If $p$ and $p'$ are honest, then every message broadcast by $p$ is eventually delivered by $p'$.
+* No message with sender $p$ is delivered unless it was previously broadcast by $p$.
+
+
+### Partially Synchronous Model and Δ-Synchrony
+
+We say the system is $\Delta$-_synchronous_ if the combined computation and communication delay between any two honest participants is smaller than $\Delta$. We assume that, under $\Delta$-synchrony, any message broadcast by an honest participant is delivered by every honest participant within a time bound of $\Delta$.
```
jennijuju, line 85:
```suggestion
We say the system is $\Delta$-_synchronous_ if the combined computation and communication delay between any two honest participants is smaller than $\Delta$. We assume that, under $\Delta$-synchrony, any message broadcasted by an honest participant is delivered by every honest participant within a time bound of $\Delta$.
```

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+>
+> `F3.invoke (int i, chain canonical, chain baseChain)` **returns** (we say finalizes) `(int i, chain h, proof_of_finality PoF)`
+>
+> **Properties:**
+>
+> **Agreement.** If two honest participants finalize $(i,h,\ast)$ and $(i,h',\ast)$, then $h = h'$.
+>
+> **Validity.** If an honest participant finalizes $(i,h,\ast)$, then $h$ is a prefix of the canonical input chain of some honest participant $p'$ in instance $i$.
+>
+> **Proof of Finality.** If an honest participant finalizes $(i,h,PoF)$, then $PoF$ is signed by ⅔ QAP majority corresponding to the $\texttt{PowerTable}(baseChain)'$ input in instance $i$ of some honest participant $p'$.
```
jennijuju, line 108: (for my understanding: is the honest participant in this step the same in the validity step?

FIPS/fip-xxxx.md
jennijuju, line 52: To clarify, is only SPs thats eligible to mine (>100TiB in power table), eligible to participate in F3 as well?

jennijuju:
Question: Are SPs required to participate in both EC and F3s? currently. If an SP refuses to contribute to EC consensus, they will be punished by not being able to produce blocks and further win block rewards. Is there any punishment for SPs that refuse to participate and contribute to F3?

FIPS/fip-xxxx.md
jennijuju: suggested to change `system actor` to `built-in actor` to avoid confusion with the actual builtin system actor (f00)

FIPS/fip-xxxx.md
jsoares: I'll address the other comments when I'm back (thanks for the thorough review!) but, on this particular matter, it was based on my (mis-)reading of the [authorship rules](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md#transferring-fip-ownership). I'll ask Henrique whether he wants to be listed -- absolutely no objection from my side.

FIPS/fip-xxxx.md
ranchalp: Not necessarily, as it can be that a participant delivers blocks/tipsets from the current epoch later than others (and particularly after having started the execution of the corresponding F3 instance). This relates to lines 54-55 in the pseudocode.

FIPS/fip-xxxx.md
ranchalp: Yes:

> EC/F3 interface: Participant feeds its current canonical chain and its previously F3-finalized chain, which we call the , to F3. The defines the power table and seeds randomness used to configure the F3 instance, while 's current canonical chain is the chain proposes to be finalized

ranchalp: More than 2/3 QAP of which more than 1/3 is QAP from honest participants.

FIPS/fip-xxxx.md
ranchalp: This is rather an assumption that something we enforce (or something we can enforce). Disqualifying a participant that holds 1/3 QAP or more from participating in F3 serves no purpose: the participant can always perform a 'sybil-like' workaround by dividing its QAP into multiple identities each holding less than 1/3. We just make it a requirement for the properties we ensure that Byzantine participants hold less than 1/3 QAP.

FIPS/fip-xxxx.md
ranchalp: These are not steps but properties. Two distinct honest participants, p and p', can finalize the same (i, h, *)  where * is PoF and PoF' respectively, such that the 2/3 QAP majority of PoF and PoF' does not exactly match. But that is ok: both PoF and PoF' are correct and both ensure at least 1/3 QAP from honest participants.

FIPS/fip-xxxx.md
ranchalp: Any participant holding power in the power table of EC is counted for the power table of F3. AFAIK this means any SP with >10TiB in the power table.

ranchalp:
@jennijuju 

> Question: Are SPs required to participate in both EC and F3s? currently. If an SP refuses to contribute to EC consensus, they will be punished by not being able to produce blocks and further win block rewards. Is there any punishment for SPs that refuse to participate and contribute to F3?

Non-participation of SPs in F3 is not directly punished currently. As long as at most <1/3 QAP is not participating, F3 operates as expected. An indirect form of punishment occurs if >=1/3 QAP does not participate: in the sense that no finalization is possible, all participants are punished by no new tipsets being finalized until at most <1/3 QAP is not participating again. 

More complex incentives are the realm of a new FIP, given the complexity of this one already. This can take the form of rewards for participants appearing in PoFs (assuming one PoF becomes part of the state of the blockchain, which is not the case at the moment), whereas punishment can be trickier (as it is hard to distinguish a straggler from a non-participating node when progress is made with only 2/3 QAP). 

Either way, note that full punishment for non-participation (or equivalently, perfectly sound reward for participation) is not desirable or even possible: this would involve all participants agreeing on which participants sent messages, which requires another instance of consensus to agree on the messages sent during the F3-instance of consensus, thus being a circular fallacy.

FIPS/fip-xxxx.md
jsoares: Checked & fixed in 9ad97ce63e4d37b60ed9077ecc0941a08b521e39.

FIPS/fip-xxxx.md
jsoares: I'm going to object here :)

https://www.merriam-webster.com/grammar/is-it-broadcast-or-broadcasted-forecast-or-forecasted-usage

FIPS/fip-xxxx.md
jsoares: Not sure if this requires clarifying in the text, but happy to add to it if so.

FIPS/fip-xxxx.md
jsoares: The current text is accurate and complete. I wouldn't make any changes unless we think it's confusing.

FIPS/fip-xxxx.md
jsoares: Likewise, I don't think we ever suggest they are the same. If you feel like people will read it implicitly, then let's adjust the text to add this note.

FIPS/fip-xxxx.md
jsoares: I think you're talking about the same number. Eligibility is gated [the same as it is for EC](https://github.com/filecoin-project/lotus/blob/886020b7ed1d644c8f7aa0647cb27ae85221b4ca/build/params_mainnet.go#L111), i.e. 10TiB:

```
var ConsensusMinerMinPower = abi.NewStoragePower(10 << 40)
```

I'll add a note to the text.

FIPS/fip-xxxx.md
jsoares: Possibly but no good explanation is coming to mind, other than the one that comes right below. Would you like to take a stab or provide a hint wrt what you think is unclear?

FIPS/fip-xxxx.md
jennijuju: I was taking the "if im an implementer, can I implement this by just reading the spec and sync the chain" approach when reading through this - while I don't think the fip suggests they are the same, but at the same time it doesn't specify what the set should be. maybe its more obvious to people who has more context?

jsoares: One option would be to write

> - **EC/F3 interface:** Participant $p$ feeds its current canonical chain $canonical$ and its previously F3-finalized chain, which we call the $baseChain$, to F3. The $baseChain$ defines the power table **used by F3** and seeds randomness used to configure the F3 instance, while $p$'s current canonical chain is the chain $p$ proposes to be finalized. Periodically, $p$ also feeds to F3 tipset updates from EC that extend $p$'s canonical chain as EC delivers these tipsets.

> In particular, EC still does a 900-epoch lookback for its power table **(which can therefore differ from the one used by F3)** and continues operating “normally” if F3 assumptions are violated and F3 halts, with the combined EC/F3 protocol favoring availability over consistency (in CAP theorem parlance).

Would this be clear enough?

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+33:     else
+34:       value ← 丄 \* vote for not deciding in this round
+35:       evidence ← nil
+
+36:     BEBroadcast <COMMIT, value, evidence, round>; trigger(timeout)
+37:     collect a clean set M of valid COMMIT messages from this round
+        until (HasStrongQuorumValue(M) AND StrongQuorumValue(M) ≠ 丄)
+          OR (timeout expires AND Power(M)>2/3)
+38:     if (HasStrongQuorumValue(M) AND StrongQuorumValue(M) ≠ 丄)    \* decide
+39:       BEBroadcast <DECIDE, StrongQuorumValue(M)
```
matejpavlovic, line 392:
```suggestion
39:       BEBroadcast <DECIDE, StrongQuorumValue(M)>
```

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+This redefinition of the heaviest chain is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of participants in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by the EC proposer.
+
+We illustrate the updated rule in the following figure, where blocks in blue are finalized blocks, and all blocks are assumed to be proposed by a proposer holding only one EC ticket (the weight of a chain is the number of blocks):
+
+![](../resources/fip-xxxx/fork.png)
+
+The current EC fork-choice rule would select the tipset $\lbrace D_0, D_1\rbrace$ as the head of the heaviest chain. However, the heaviest finalized tipset is $\lbrace C_3\rbrace$, which is not an ancestor of $\lbrace D_0, D_1\rbrace$. Therefore, the new fork choice rule selects $\lbrace D_3\rbrace$ as the head of the heaviest chain. The reason why $D_4$ is not selected is that its parent tipset does not exactly match the finalized tipset $\lbrace C_3\rbrace$, but a superset of it, i.e. $\lbrace C_3, C_4\rbrace$.
```
ruseinov, line 199: I remember we discussed that EC should work as it did previously in case f3 fails for some reason. And I guess it will, once F3 stops finalizing - the EC will further do it's choices as it did pre-F3, given that the prefix is the latest `f3-finalized` chain. Correct? Perhaps worth mentioning here.

ranchalp: That's correct

FIPS/fip-xxxx.md
```diff
@@ -0,0 +1,747 @@
...
+
+
+### Bootstrapping
+
+Integrating F3 into Filecoin follows the usual path for Filecoin upgrades. One epoch, $upgradeEpoch$, will be defined as the target epoch upon which participants upgrade Filecoin. Then, every participant starts the first instance with the tipset at the $upgradeEpoch$ minus the 900-epoch lookback as the head tipset of the first $baseChain$, which is assumed by design to be common to all participants.
```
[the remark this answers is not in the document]
ranchalp: What do you mean by implied?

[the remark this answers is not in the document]
ruseinov: I mean when it's said `first instance` what's actually meant is `first instance of F3`.

FIPS/fip-xxxx.md
jsoares:
```suggestion
The current EC fork-choice rule would select the tipset $\lbrace D_0, D_1\rbrace$ as the head of the heaviest chain. However, the heaviest finalized tipset is $\lbrace C_3\rbrace$, which is not an ancestor of $\lbrace D_0, D_1\rbrace$. Therefore, the new fork choice rule selects $\lbrace D_3\rbrace$ as the head of the heaviest chain. The reason why $D_4$ is not selected is that its parent tipset does not exactly match the finalized tipset $\lbrace C_3\rbrace$, but a superset of it, i.e. $\lbrace C_3, C_4\rbrace$.

In the event that F3 halts, the fork choice rule allows EC to continue to progress, building a chain of unbounded length, per the current rules, on the last F3-finalized prefix.
```

@ruseinov @ranchalp would this be clear?

FIPS/fip-xxxx.md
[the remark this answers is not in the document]
jsoares: Fixed in 40fa4f1

FIPS/fip-xxxx.md
ruseinov: Looks good to me.

momack2 (approved):
Very excited for the benefits of F3! This looks to have well thought out security, product, and incentive considerations, and a very thorough protocol description. Think deserves a FIP number and merge

FIPS/fip-0086.md
```diff
@@ -0,0 +1,750 @@
...
+
+
+### Changes to EC: Fork Choice Rule
+
+EC needs to be modified to accommodate the finalization of tipsets by F3. **This is the only change to EC this FIP introduces.**
+
+The current EC fork-choice rule selects, from all the known tipsets, the tipset backed by the most weight. As the F3 component finalizes tipsets, the fork-choice rule must ensure that the heaviest finalized chain is always a prefix of the heaviest chain, preventing reorganizations of the already finalized chain.
+
+We achieve this by adjusting the definition of weight for a finalized prefix: the heaviest finalized chain is the one that **matches exactly the tipsets finalized by F3**, in that a tipset $t'$ that is a superset of finalized tipset $t$ in the same epoch is not heavier than $t$ itself, despite it being backed by more EC power.
```
jennijuju, line 191: oh interesting - what would EC do if F3 is halted and not produce new tipsets?

FIPS/fip-0086.md
```diff
@@ -0,0 +1,750 @@
...
+
+This redefinition of the heaviest chain is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of participants in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by the EC proposer.
```
jennijuju, line 193: To make sure im understanding this correctly - participants in gossipbft will proposes blocks and will be finalized by f3; SPs will propose blocks on EC; only those blocks exists in both f3 and EC would be considered as a final block on filecoin and get block rewards. Does this sound right?

FIPS/fip-0086.md
```diff
@@ -0,0 +1,750 @@
...
+
+Assuming some genesis tipset with a known power table, a _GossiPBFTMessage_ can provide a commitment to (CID of) the resulting power table but doesn't provide the actual entries. A finality certificate includes these as a delta from the previous power table. A verifier can compute the power table for the subsequent certificate by applying these deltas to its base power table.
+
+Note the power table deltas don't need to be signed. All DECIDE messages include a commitment to the CID of the resulting power table. Hence, a verifier only needs to confirm that the power table CID as a result of their computation matches that committed to by the signed messages. A finality certificate with incorrect power table deltas cannot be used as a base to verify subsequent instances.
+
+
+#### Verification by Filecoin node (“fast catch-up”)
+
+A new Filecoin validating node can listen to GossiPBFT gossip to learn about the current instance number. They can then issue exchange requests to other nodes to fetch (in parallel) the complete sequence of finality certificates back to their genesis. **These certificates can be verified without reference to the EC chain.**
```
jennijuju, line 583:
```suggestion
A new Filecoin validator node **
```

jennijuju (approved):

kaitlin-beegle (approved):
Completed full editorial review.  No nits or open asks.  The proposal is complete and well written.  Strong endorsement to proceed.

FIPS/fip-0086.md
rvagg: from below

> In the event that F3 halts, the fork choice rule allows EC to continue to progress, building a chain of unbounded length, per the current rules, on the last F3-finalized prefix.

I assume that just means that there exists no F3-finalised prefix to be a superset of, so it falls back to normal EC rules.

jennijuju: Right - my q is more about whether there is any latency on when the fallback triggers, how does the chain process know whether receiving no tip sets from F3 is cuz it’s halted or cuz it’s a null round on f3, and decided to proceed with EC or not

jsoares: There's no explicit switchover. New tipsets are always appended by EC to the heaviest known chain. By definition, the heaviest known chain _must_ include the _latest_ finalised prefix, but it doesn't "care" if this prefix is 1 or 100 epochs old. If no new prefixes are being finalised, it will still keep building on the heaviest known chain. So there's no decision moment or associated latency.

FIPS/fip-0086.md
jsoares: F3 proposals start from chainHead(), which returns the local EC chain, i.e. tipsets always start their life on the EC side. F3 periodically "freezes" a canonical prefix. Block rewards are disbursed in the context of SBC/EC but, of course, they'll only persist on the canonical chain (i.e. rewards that were earned in a fork that doesn't get selected go away with that fork). So yes, practical block rewards will be those in the canonical chain, which will exist in f3 and EC. Does that make sense?

This is upon convergence, of course. Different blocks may "exist" in different nodes' local views at a specific (yet to be finalised) point in time, as nodes may start with different EC chains and therefore submit different proposals to F3. Deciding for an EC chain that is not locally available is problematic, and this is handled in GossipBFT. But I'll let @ranchalp provide any nuance before I say something very stupid :)

PR #898, opened 2023-12-21
FIP-0076: Add AggregateProofType parameter to ProveCommitSectors3Params.

anorth:
Adds explicit proof type parameter to ProveCommitSectors2Params, rather than hardcoding SnarkPackV2. This follows the pattern already used by ProveReplicaUpdates[2] and supports development of new proofs without necessarily changing method signature.

I don't believe this is a substantive change that should require restarting Last Call.

Suggested by @Kubuxu

### DIFF
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -183,6 +183,8 @@ struct ProveCommitSectors3Params {
     // Aggregate proof for all sectors.
     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
     AggregateProof: []byte,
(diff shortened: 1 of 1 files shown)


jennijuju (approved):

jennijuju:
the change LGTM - out of curiosity, do we need to define a RegisteredAggregateProof enum in builtin actor for this change?

anorth:
> do we need to define a RegisteredAggregateProof enum in builtin actor for this change

I don't think so, these are the same proofs.

kaitlin-beegle (approved):

PR #902, opened 2024-01-03
FIP-0076: update return type name

jennijuju:
https://github.com/filecoin-project/builtin-actors/pull/1485 

just to match the implementation, doesn't propose any substantial changes to the fip itself


anorth (approved):
Thanks

kaitlin-beegle (approved):

PR #904, opened 2024-01-04
Remove unnecessary section

cryptonemo:


kaitlin-beegle:
This isn't an unnecessary section- it's part of the [FIP template](https://github.com/filecoin-project/FIPs/blob/master/templates/template_FTP.md).  Even if it's not fully relevant to your FIP, it should remain.

kaitlin-beegle:
This isn't an unnecessary section- it's part of the [FIP template](https://github.com/filecoin-project/FIPs/blob/master/templates/template_FTP.md). Even if it's not fully relevant to your FIP, it should remain.

PR #909, opened 2024-01-09
FIP-0082: Note FIP-0076 as dependency and propagate upstream changes

anorth:
Note FIP-0076 as dependency, and update to reflect method renaming and renumbering in FIP-0076.

I don't believe this constitutes any substantive change or should interrupt Last Call.

FYI @cryptonemo


cryptonemo (approved):
Thank you!

jsoares (approved):

jennijuju (approved):

PR #914, opened 2024-01-12
FIP-0063: Add upgrade and protocol change details

arajasek:
Two changes of significance:

- Explicitly stating that we no longer include "missed" beacon entries in block headers (strictly one beacon entry per block)
- Tweaking the epoch -> round calculation to include a +1

This is strictly providing a bit more detail to the FIP, and IMO shouldn't materially impact acceptance / Last Call status.

### DIFF
--- a/FIPS/fip-0063.md
+++ b/FIPS/fip-0063.md
@@ -75,12 +75,13 @@ The details of `quicknet` chain info are:
 
 The way beacons are retrieved must be adapted to fetch every 10th beacon instead of every beacon to accommodate for Filecoin epochs. In previous iterations of filecoin, operators could request drand beacons with monotonically increasing round numbers, as the drand and filecoin epochs were aligned. Upon switching to the new scheme, consumers must ensure they request the right drand epoch over either HTTP or gossipsub.
 
-The beacon round corresponding to a filecoin epoch at time `T` is calculated as `(T - 1692803367)/3 + 1`, where 1692803367 is the genesis timestamp of Quicknet, and 3 is the period duration of Quicknet.
+For a Filecoin epoch at time `T` seconds, we source randomness from a round at time `T - 30` seconds. Thus, the beacon round 
+corresponding to a filecoin epoch at time `T` is calculated as `(T - 30 - 1692803367)/3 + 1`, where 1692803367 is the genesis timestamp of Quicknet, and 3 is the period duration of Quicknet.
 
 ### Block Headers
 
 Blocks should only ever include one beacon entry in their header, for the round corresponding to the Filecoin epoch as described above.
-This is a change from existing behaviour, where block headers sometimes include multiple beacon entries (specifically when built on "null" tipsets).
+This is a change from existing behaviour, where block headers sometimes include multiple beacon entries (specifically when built on "null" tipsets). Storing multiple beacons is no longer necessary as the unchained nature of Quicknet does not require us to keep and validate the full history of consecutive rounds.
 
 ### Upgrade
 
@@ -88,9 +89,11 @@ The switch to Quicknet will happen in a network upgrade. All blocks produced in 
 
 ### Catch-up
 
-The new catch-up time of 2s per beacon will result in a longer catch-up period per Filecoin epoch of 20 seconds rather than the current 15 seconds; any monitoring, alerting, reporting or automation relying on the existing catchup time must be updated with the new period of 20s.
-When fetching only every 10th beacon with a catch-up time of 2 seconds per round, the total catch-up time per Filecoin epoch is slightly increased from 15 seconds to 20 seconds. This is only applicable in times of drand outage,  of which there have been none since the launch of Filecoin. 
-If drand goes down, Filecoin stops producing blocks. Once drand comes back online, it produces randomness slightly faster, but Filecoin's recovery is bound by this time-frame.
+If drand halts, Filecoin stops producing blocks. Once drand comes back online, it produces randomness faster until it catches up to the expected present round. 
+
+Filecoin, too, has a similar catch-up mechanism that consumes the expected beacons (per the mapping defined above) as they become available. Quicknet has a catch-up time of 2s per beacon, and so the total catch-up time per Filecoin epoch is increased from 15 seconds to 20 seconds. 
+
+This is only noticeable in case of a drand outage, of which there have been none since the launch of Filecoin. Nevertheless, any monitoring, alerting, reporting, or automation relying on the existing catch-up time should be updated with the new period of 20s. 
 
 ## Design Rationale
 
--- a/FIPS/fip-0063.md
+++ b/FIPS/fip-0063.md
@@ -34,6 +34,8 @@ For further details on these new features and their implementation we refer the 
 
 ## Specification
 
+### Beacon Verification
+
 Currently drand verification is performed in two stages:
 
 1. hash the previous beacon’s signature and the current beacon round number using SHA256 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0063.md
```diff
@@ -69,13 +71,22 @@ The details of `quicknet` chain info are:
 }
 ```
 
-- retrieve every 10th beacon
+### Beacon Sourcing
 
 The way beacons are retrieved must be adapted to fetch every 10th beacon instead of every beacon to accommodate for Filecoin epochs. In previous iterations of filecoin, operators could request drand beacons with monotonically increasing round numbers, as the drand and filecoin epochs were aligned. Upon switching to the new scheme, consumers must ensure they request the right drand epoch over either HTTP or gossipsub.
 
-`quicknet` genesis was on 1692803367, while Filecoin Genesis was on 1598306400, thus Filecoin block 3547000 (2024-01-08 12:20:00 (UTC)) should be mapped to our drand quicknet round 3971011 because `1598306400+3547000*30 == 1692803367+3971011*3`  and all subsequent Filecoin rounds are mapped to multiple of 10 after that: `3971011+x*10`
+The beacon round corresponding to a filecoin epoch at time `T` is calculated as `(T - 1692803367)/3 + 1`, where 1692803367 is the genesis timestamp of Quicknet, and 3 is the period duration of Quicknet.
```
arajasek, line 78: Reviewers please examine this closely, especially confirming the `+1` is necessary.

CluEleSsUK: indeed, the genesis time is round 1

FIPS/fip-0063.md
```diff
@@ -69,13 +71,22 @@ The details of `quicknet` chain info are:
...
+
+### Block Headers
+
+Blocks should only ever include one beacon entry in their header, for the round corresponding to the Filecoin epoch as described above.
+This is a change from existing behaviour, where block headers sometimes include multiple beacon entries (specifically when built on "null" tipsets).
```
CluEleSsUK, line 83: I don't follow why this behaviour would be different, though it's not my area of expertise so I trust your judgement!

CluEleSsUK (approved):

FIPS/fip-0063.md
jsoares:
```suggestion
This is a change from existing behaviour, where block headers sometimes include multiple beacon entries (specifically when built on "null" tipsets). Storing multiple beacons is no longer necessary as the unchained nature of Quicknet does not require us to keep and validate the full history of consecutive rounds.
```

I suppose this is the motivation, which, if not obvious to the authors, probably wouldn't be obvious to the readers either.

jsoares (approved):

FIPS/fip-0063.md
```diff
@@ -69,13 +71,22 @@ The details of `quicknet` chain info are:
...
+
+### Upgrade
+
+The switch to Quicknet will happen in a network upgrade. All blocks produced in the new network version should source their beacons from Quicknet.
 
-- account for new catch-up time
+### Catch-up
 
 The new catch-up time of 2s per beacon will result in a longer catch-up period per Filecoin epoch of 20 seconds rather than the current 15 seconds; any monitoring, alerting, reporting or automation relying on the existing catchup time must be updated with the new period of 20s.
 When fetching only every 10th beacon with a catch-up time of 2 seconds per round, the total catch-up time per Filecoin epoch is slightly increased from 15 seconds to 20 seconds. This is only applicable in times of drand outage,  of which there have been none since the launch of Filecoin.
```
jsoares, line 92:
```suggestion
If drand halts, Filecoin stops producing blocks. Once drand comes back online, it produces randomness faster until it catches up to the expected present round. Filecoin, too, has a similar catch-up mechanism that consumes the expected beacons (per the mapping defined above) as they become available. Quicknet has a catch-up time of 2s per beacon, and so the total catch-up time per Filecoin epoch is increased from 15 seconds to 20 seconds. 

This is only noticeable in case of a drand outage, of which there have been none since the launch of Filecoin. Nevertheless, any monitoring, alerting, reporting, or automation relying on the existing catch-up time should be updated with the new period of 20s. 
```

Apologies for abusing your PR, but the lack of context wrt to beacon inclusion above had me thinking that this would also benefit from rephrasing and an explicit statement that we maintain the drand<>Filecoin mapping in case of outage and catch-up. This is not a substantive change, just editing for clarity and completeness.

The change also require deleting the line below (82/93) but given it's outside the present scope of the PR, GH will not let me do it as suggestion.

FIPS/fip-0063.md
hanabi1224: A quick question, how is epoch time T calculated? In the previous doc it says `1598306400+3547000*30=1704716400` while in the lotus code, it seems to be `1598306400+3547000*30 - 30=1704716370` 

```go
func (db *DrandBeacon) MaxBeaconRoundForEpoch(nv network.Version, filEpoch abi.ChainEpoch) uint64 {
	// TODO: sometimes the genesis time for filecoin is zero and this goes negative
	latestTs := ((uint64(filEpoch) * db.filRoundTime) + db.filGenTime) - db.filRoundTime

	if nv <= network.Version15 {
		return db.maxBeaconRoundV1(latestTs)
	}

	return db.maxBeaconRoundV2(latestTs)
}

func (db *DrandBeacon) maxBeaconRoundV2(latestTs uint64) uint64 {
	if latestTs < db.drandGenTime {
		return 1
	}

	fromGenesis := latestTs - db.drandGenTime
	// we take the time from genesis divided by the periods in seconds, that
	// gives us the number of periods since genesis.  We also add +1 because
	// round 1 starts at genesis time.
	return fromGenesis/uint64(db.interval.Seconds()) + 1
}
```

AnomalRoil: This might have worked previously since the previous drand genesis was on `1595431050` and its interval was 30s, because upon input of `1704716370` we do get the drand round for `1704716400`:
`maxBeaconRoundV2(1704716370) = (1704716370 - 1595431050)/30 + 1 = 3642845`
and we can see that `1595431050 + 3642845*30 == 1704716400` as expected, so the expected `default` drand round for time 1704716400 was the one we would get upon getting as input `1704716370` (it's basically thanks to the `+1` in the code of `maxBeaconRoundV2` and because the period is 30s, just like Filecoin's `filRoundTime`).

However if we get as input `1704716370` with the new `quicknet` network and a period of 3s,  we would get:
`maxBeaconRoundV2(1704716370) = (1704716370 - 1692803367)/3 + 1 = 3971002`
but the drand quicknet round `3971002` actually happened on `1692803367+3*3971002 = 1704716373` which is 27 seconds too early compared to the value we're expecting, which is `1704716400`.

I believe this PR actually requires the introductions of a `maxBeaconRoundV3` function and a new `network.VersionXX` guard, but I'm not familiar with how `network.Version` is actually used or not to verify drand beacons using the right round or not. 

(I have now sadly looked at the `BeaconEntriesForBlock` function in the codebase and my hope that there aren't more bugs lurking is now much lower, but also have given up on the idea of actually understanding how Lotus uses drand rounds.)

hanabi1224: A side note, when I see `dround := (latestTs - drandGenesisTime) / uint64(drandPeriod) + 1` I was wondering if whether meant to be `dround := (latestTs - drandGenesisTime + filEpochDuration) / uint64(drandPeriod)` as they are equivalent on drand mainnet but the later looks more proper to me when it comes to quicknet

arajasek: Thanks for the conversation here! I'm going to clarify this in my suggested changes, but my understanding is:

- A block at time `T` must be _ready to be published_ at time `T`. That means its inputs must be available at some point earlier than `T`, in order to allow for some time to "create" the block. Today (nv21), we fetch drand at time `T - 30` seconds, giving us 30 seconds to mine the block. This is considered enough time to mine the block, and is the smallest possible (non-zero) delay between Beacon availability and block preparation, since the Beacon round time is also 30 seconds.
  -  There is an open question about whether we should be keeping this delay at 30 seconds, or whether we want to reduce it -- since quicknet has a period of 3 seconds, we could hypothetically reduce this delay to any multiple of 3 seconds. I'd like @nicola / @Kubuxu / @irenegia to weigh in here.
- Separately, there's the issue of the `+1`, which simply arises from Drand's first round being 1.

FIPS/fip-0063.md
arajasek: This is the correct motivation, yes. I  have a preference for restricting the `Specification` section of FIPs to the actual changes (not including motivation) so that it can read as a "here's what to implement" section.

I'm happy to include this one sentence though, thanks @jsoares!

FIPS/fip-0063.md
Kubuxu: I think keeping it at 30s is the right choice as of now. We can use it to adjust in the future in separate FIP.

jsoares (approved):

jennijuju (approved):

PR #923, opened 2024-01-22
Update FIP-0076 with the gas costs for emitting built-in Actor events

aarshkshah1992:
Updates FIP-0076 with the gas costs for emitting built-in Actor events for the Miner Actor.


jsoares (approved):
I'm not entirely convinced that FIP-0076 should be listing gas costs in re the "later" FIP-0083 (vs other way around), but the text is clear.

anorth:
I agree this seems like the wrong place. Can you move it to FIP-0083?

aarshkshah1992:
@anorth @jsoares Raised PR to update FIP-0083 with the gas costs. Closing this. Thanks a lot.

PR #924, opened 2024-01-22
Remove TBD from FIP-0082

anorth:


FIPS/fip-0082.md
```diff
@@ -319,9 +319,7 @@ proposal (but _should_ in order to enjoy the reduced gas costs).
 
 ## Test Cases
 
-Test cases will accompany implementation. Suggested scenarios include:
-
-TBD
+Test cases will accompany implementation.
```
jennijuju, line 322: I think either authors should add links to point to existing tests or the TODO remains

anorth: This FIP is already accepted. The author doesn't need to do anything, though we could ask nicely.

If you can arrange for someone to put higher quality information in here, great. If not, leaving the TODO is the worst outcome, so please accept the tidy up that we're not, in fact, going to do it.

jsoares (approved):

jsoares:
> This FIP is already accepted. The author doesn't need to do anything, though we could ask nicely.

I'm fine with the proposed change but don't know that I agree with this general statement. E.g. the implementation section is not in good shape and needs updating with appropriate links before the FIP is finalised. This often happens after acceptance, given implementation timeline, but should still happen. 

Anyhow, can we merge this? It won't be any worse than currently. Pinging @filecoin-project/fips-editors for additional review.

TippyFlitsUK (approved):

PR #928, opened 2024-01-25
FIP-0063: Tweak protocol change details

arajasek:
This makes a minor change to the protocol changes described in #914. This change corrects an oversight in that PR, and continues to guarantee that beacon randomness can be sourced for every epoch, even if the epoch is null.

### DIFF
--- a/FIPS/fip-0063.md
+++ b/FIPS/fip-0063.md
@@ -80,8 +80,10 @@ corresponding to a filecoin epoch at time `T` is calculated as `(T - 30 - 169280
 
 ### Block Headers
 
-Blocks should only ever include one beacon entry in their header, for the round corresponding to the Filecoin epoch as described above.
-This is a change from existing behaviour, where block headers sometimes include multiple beacon entries (specifically when built on "null" tipsets). Storing multiple beacons is no longer necessary as the unchained nature of Quicknet does not require us to keep and validate the full history of consecutive rounds.
+As today, blocks should include beacon entries for the epoch corresponding to their own height, as well as those corresponding to any parent epochs that were null.
+This guarantees that a unique beacon entry can be found for any Filecoin epoch to serve as a source of external randomness to the protocol (currently used for WindowPoSt and ProveCommit verification).
+
+Note that this does NOT mean that every beacon entry is included in the blockchain as it is today. Instead, only every 10th entry will be included.
 
 ### Upgrade
 


jsoares:
@arajasek the amended text (or the reason therefor, I suppose) is not 100% clear to me. We've established that we don't need to store the sequence of rounds for randomness verification purposes. Do we still need to store to be able to verify ostensible tickets for the null round, or is it something else?

FIPS/fip-0063.md
```diff
@@ -80,8 +80,10 @@ corresponding to a filecoin epoch at time `T` is calculated as `(T - 30 - 169280
 
 ### Block Headers
 
-Blocks should only ever include one beacon entry in their header, for the round corresponding to the Filecoin epoch as described above.
-This is a change from existing behaviour, where block headers sometimes include multiple beacon entries (specifically when built on "null" tipsets). Storing multiple beacons is no longer necessary as the unchained nature of Quicknet does not require us to keep and validate the full history of consecutive rounds.
+As today, blocks should include beacon entries for the epoch corresponding to their own height, as well as those corresponding to any epoochs that were null.
+This guarantees that a unique beacon entry can be found within the Filecoin blockchain to serve as a source of external randomness to the protocol.
```
anorth, line 84:
```suggestion
This guarantees that a unique beacon entry can be found for any Filecoin epoch to serve as a source of external randomness to the protocol.
```

arajasek:
> @arajasek the amended text (or the reason therefor, I suppose) is not 100% clear to me. We've established that we don't need to store the sequence of rounds for randomness verification purposes. Do we still need to store to be able to verify ostensible tickets for the null round, or is it something else?

We need to be able to _source_ beacon randomness for the null round (from the blockchain itself), which means we need to include (and verify) these "missed" rounds. This is becaues the protocol needs to be able to supply beacon randomness for any epoch, including null epochs. 

As an example, the challenge epoch for beacon randomness for some WindowPoSt could be epoch 10. If epoch 10 was null, then we expect the block at epoch 11 to include the missed round. If we don't have that, we'd have to use the randomness corresponding to epoch 11 instead, which while prooobably fine opens grinding possibilities.

So for now, just include beacon entries for any immediately preceding null rounds.

I've (hopefully) clarified the text to explain this better, feel free to suggest improvements!

jsoares (approved):

Stebalien (approved):

PR #930, opened 2024-01-29
FIP-0076 fix: s/SectorDeals/SectorDealIDs in market state to disambiguate naming

rvagg:
`SectorDeals` is also used in `BatchActivateDeals` but has a different form there. This change simply renames the struct to disambiguate it. This also reflects how it's implemented in builtin-actors.

---

While I'm in here, can I ask for clarity on why this is a struct in the first place? It's introduced new complexity now that it's adopted "transparent" encoding in builtin-actors which has required new cbor-gen adjustments to make it work with go-state-types too. Why are we containing it in a struct in the first place and not just doing `SectorDealIDs []DealID` or omitting `SectorDealIDs` entirely and saying `HAMT[Address]HAMT[SectorNumber][]DealID`, which would both result in the same forms as they are currently implemented in both of the proposed builtin-actors and go-state-types but may afford a little bit of simplification in the code by removing an intermediary?

_(Edit: here's what such simplification could look like: https://github.com/filecoin-project/builtin-actors/pull/1510)_

### DIFF
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -465,7 +465,7 @@ The unused verified claim ID field is removed from the deal state structure.
 
 ```
 // New structure storing per-sector deal information.
-struct SectorDeals {
+struct SectorDealIDs {
     Deals: []DealID
 }
 
@@ -492,7 +492,7 @@ struct State {
 	States: Cid 
 
     // New mapping of sector IDs to deal IDs, grouped by storage provider.
-    // HAMT[Address]HAMT[SectorNumber]SectorDeals
+    // HAMT[Address]HAMT[SectorNumber]SectorDealIDs
     ProviderSectors: Cid 
 }
 ```


jennijuju (approved):

jennijuju:
The fip change lgtm to reflect the implementation. 

(Cc @Stebalien  @anorth for  the  question asked in PR description

Stebalien:
I agree with the change in https://github.com/filecoin-project/builtin-actors/pull/1510, honestly. Wrapping in a struct makes some sense for method parameters and return values, but doesn't really add anything here (IMO).

anorth:
> Why are we containing it in a struct in the first place and not just doing SectorDealIDs []DealID or omitting SectorDealIDs entirely and saying HAMT[Address]HAMT[SectorNumber][]DealID, which would both result in the same forms as they are currently implemented in both of the proposed builtin-actors and go-state-types but may afford a little bit of simplification in the code by removing an intermediary?

Sure, they're all fine. I don't understand the problem. Pseudocode in the FIP isn't code code, implementations should represent it however makes sense to them. I would happily also switch to an IPLD-CBOR schema language if that was going to be clearly understandable to readers (it might! I'd love to!).

anorth (approved):

PR #931, opened 2024-01-29
Create CODEOWNERS for review requests

jennijuju:
I believe the goal/rule is to have 2 fip editors' review & approval for an FIP update to be merged into the repo. without this configuration, currently any 2 approvals from anyone has write access to this repo can approve and merge PR to the main branch.
image

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,4 +3,4 @@
 
 # Global owners
 # Ensure maintainers team is a requested reviewer for non-draft PRs
-*                       @filecoin-project/fips-editors
+*                       @momack2 @arajasek @jennijuju @kaitlin-beegle @anorth @raulk @jsoares
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-*       @momack2 @arajasek @jennijuju @kaitlin-beegle @anorth @raulk @jsoares


Stebalien (approved):

jsoares (commented):
There is a CODEOWNERS file in the root: https://github.com/filecoin-project/FIPs/blob/master/CODEOWNERS

It doesn't use the team but lists the editors individually. It's a bit more transparent (since the makeup of the team isn't visible to the public), but I think either works in practice. What we shouldn't do is keep both files (though the one in .github would take precedence).

anorth:
What @jsoares said. Please use the existing CODEOWNERS with its transparent list of names. Toggling the box in the settings is 👍 though

jennijuju:
@jsoares @anorth for the `Required review from code owners setting` to work, the code owner must be in ./github https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection 

That is why the even tho we have always had the box toggled as screenshotted, however, still  PR can be merged into protected branch (master) with 2 approval from anyone with write access, instead of 2 approval from code owners (fip editors)

jsoares (approved):
> for the Required review from code owners setting to work, the code owner must be in ./github

I don't think that's accurate or what the page says

> That is why the even tho we have always had the box toggled as screenshotted, however, still PR can be merged into protected branch (master) with 2 approval from anyone with write access, instead of 2 approval from code owners (fip editors)

I believe that the new settings only require review from *1* codeowner (and 2 reviews in total from users with write access). The language of the checkbox seems to agree, i.e. "*an* approved review". I don't think github provides a way to enforce 2 reviews from codeowners, but I might be wrong.

But the change lgtm now that it's not duplicative.

anorth (approved):
I believe @jsoares is correct and there is no way to require approval from two CODEOWNERS. While we can't enforce this easily, I don't think we should try too hard either. We FIP editors know the rules. There's a relatively small set of people with write access outside the editors, who can be educated. If a mistake is made, it's pretty easy to revert (tho in most cases the non-editor-writer who "accidentally" merges is probably doing what we should have done sooner anyway).

Moving the CODEOWNERS file into `.github` is a minor regression in discoverability. I don't object too strongly so will approve, but suggest you close this PR instead if convinced that it will have no practical effect.

PR #934, opened 2024-02-01
FIP-0076: make aggregate proof type parameter an option, absent if not aggregating

anorth:
Implemented in https://github.com/filecoin-project/builtin-actors/pull/1511

### DIFF
--- a/FIPS/fip-0076.md
+++ b/FIPS/fip-0076.md
@@ -186,8 +186,8 @@ struct ProveCommitSectors3Params {
     // Aggregate proof for all sectors.
     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
     AggregateProof: []byte,
-    // The proof type for the aggregate proof (ignored if no aggregate proof).
-    AggregateProofType: RegisteredAggregateProof,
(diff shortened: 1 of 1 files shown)


arajasek (approved):

Stebalien (approved):

FIPS/fip-0076.md
```diff
@@ -186,8 +186,8 @@ struct ProveCommitSectors3Params {
     // Aggregate proof for all sectors.
     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
     AggregateProof: []byte,
-    // The proof type for the aggregate proof (ignored if no aggregate proof).
-    AggregateProofType: RegisteredAggregateProof,
+    // The proof type for the aggregate proof (must be absent if no aggregate proof).
```
Stebalien, line 189: Hm. Actually, this should probably say "must be null".

jennijuju (approved):

FIPS/fip-0076.md
anorth: I've left this as a future-proof comment. It's true that today there must be no aggregate proof, but that's the true cause of the constraint.

PR #936, opened 2024-02-07
FIP-0083: Add title

rvagg:
Following convention that appears to be set for other FIPs

/cc @aarshkshah1992

### DIFF
--- a/FIPS/fip-0083.md
+++ b/FIPS/fip-0083.md
@@ -9,7 +9,9 @@ category: Core
 created: 2023-11-27
 ---
 
-## Summary
+# FIP-0083: Deal-lifecycle Actor Events
+
+## Simple Summary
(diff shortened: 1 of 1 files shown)


jennijuju (approved):

rvagg:
I'm calling it "Deal-lifecycle Actor Events" but maybe "Data-lifecycle Actor Events" would be more appropriate?

jsoares (approved):

FIPS/fip-0083.md
```diff
@@ -9,7 +9,9 @@ category: Core
 created: 2023-11-27
 ---
 
-## Summary
+# FIP-0083: Deal-lifecycle Actor Events
```
anorth, line 12: The title is set in the root README. Please make them consistent. I suggest the easiest way is to match that title here.

> Add built-in Actor events in the Verified Registry, Miner and Market Actors

rvagg: ahh, gotcha, missed that

PR #941, opened 2024-02-11
FRC-0089: A Finality Calculator for Filecoin

jsoares:
Discussion: #919

Finality on Filecoin isn't fast. Expected consensus is a probabilistic finality protocol, and the spec encodes a soft-finality threshold of 900 epochs, which was set in order to ensure a probability of reversal $< 2^{-30}$. Users and applications can use more relaxed confirmation rules, but will generally still wait $>100$ epochs before considering a transaction final. This can lead to hours of wait time, and harms the UX.

In this FRC, we specify a calculator that enables users to compute the reversal probability of any given tipset. Instead of always assuming the worst case, it leverages the local node view and chain history. By doing so, it can provide the same level of assurance in ~30 epochs (30x faster) under normal operating conditions. This does not require any protocol change, and is therefore best interpreted as a technical recommendation to implementers and users.

cc @guy-goren


anorth (approved):
Thanks for submitting a well-formatted and thorough document, and having a lot of discussion about it in public earlier. These both make it much easier to move along.

Please take FRC# 0089. Update the title, filename, and README.

jsoares:
Thank you @anorth! Files updated with FRC number, awaiting further reviews.

TippyFlitsUK (approved):

PR #949, opened 2024-02-22
Update FIPs table to reflect recent statuses

luckyparadise:
FIP0084 has since passed out of Last Call and the new status needs to be updated. FIP0085 has recently entered into Last Call and is no longer in Draft status.


jsoares (approved):

jennijuju (approved):

jaychoucc:
support

PR #961, opened 2024-03-15
FIP-0045: update to match deployed types

rvagg:
**Background**: it was discovered (by @LexLuthr) that `ClaimExtensionRequest` has been implemented wrong in go-state-types so has not been a functional API on the network since nv10, although it was launched (as per this FIP) in nv9 with the types matching. That's being rectified in https://github.com/filecoin-project/go-state-types/pull/254, although it's conforming to the implementation found in builtin-actors rather than the text of the spec in this FIP.

In the process of proposing a retroactive fix to this spec to match the deployed implementation, I've discovered that there are a lot more mismatched types between spec and implementation, so I've adjusted them all here. This isn't a new FIP, but rather an alignment of the spec if we view the spec as a combination of the FIP text and the builtin-actors implementation; and the only thing changing in code as a result of this is a fix to make `ClaimExtensionRequest` work.

---

* Many of the type fields marked as `Address` in here were launched in nv9 as `ActorID`.
* `AllocationRequest` and `ClaimExtensionRequest` went live in nv9 with `provider` as `Address` but in nv10 switched to `ActorID`.
* `ClaimExtensionRequest` was never adjusted in go-state-types to `ActorID` so has been unable to successfully parse or pass on these to builtin-actors. `ExtendClaimTerms` appears to have been the only mechanism used to extend claim terms since this FIP.
* Minor: updated all epoch fields to match within this FIP, I chose to use `int64` which is the base type used in implementations.
* Add missing docs for `AllocationRequest` and `ClaimExtensionRequest` fields.
* Added historical notes section.

### DIFF
--- a/FIPS/fip-0045.md
+++ b/FIPS/fip-0045.md
@@ -84,21 +84,21 @@ built-in storage market (or any future market).
 ```
 struct Allocation {
     // The verified client which allocated the DataCap.
-    Client: Address
+    Client: ActorID
     // The provider (miner actor) which may claim the allocation.
-    Provider: Address
+    Provider: ActorID
     // Identifier of the data to be committed.
     Data: CID
     // The (padded) size of data.
     Size: uint64
     // The minimum duration which the provider must commit to storing the piece to avoid
     // early-termination penalties (epochs).
-    TermMinimum: uint64
+    TermMinimum: int64
     // The maximum period for which a provider can earn quality-adjusted power 
     // for the piece (epochs).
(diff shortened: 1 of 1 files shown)


jsoares (approved):

rvagg:
Thanks to Aayush for pointing out that I was using builtin-actor versions instead of network versions. I've fixed that now, 9->17 and 10->18.

jennijuju (approved):

anorth (approved):

PR #964, opened 2024-03-18
FIP-0083: Add additional event properties to ease DDO transition

rvagg:
This is a revisit of #955, a proposal to add additional information to built-in actor events, but adds more information across more events than #955 did. That PR was discussed further in this thread: https://github.com/filecoin-project/FIPs/discussions/754#discussioncomment-8698293, where it was pointed out regarding the costs of these events:

> On thing that is new since the beginning of this discussion is that we have quantified the gas cost. Built-in actor methods tend to be doing a lot, so the cost of events is very small as a proportion. Perhaps the minimalist approach isn't right here, and we should just include all the near-to-hand information with all events?

After receiving additional feedback on the information currently available in events, including from a raised awareness of the implications of DDO in network version 22 on observability concerns, we're coming back with **this new proposal that we're hoping to get approved and shipped _with_ network version 22**.

Given the tight time constraints and concerns about stability, we've attempted to craft a line between "near-to-hand" information (i.e. low-risk, minimal code changes) and "maximally useful" when considering the requirements we are hearing from parties concerned about a DDO world.

A lot of the feedback we have received is related to the verified registry and the difficulties in tracking claims, pieces, SPs and clients and their relationships. The events, as they are currently structured, require reactive state inspection to provide enough context to make sense of them, such as the example of claims outlined #955.

The diff here also includes some formatting changes, so it's a bit noisier than just the pure event changes, sorry. The changes proposed here are summarised as follows:

**Verified registry events**

* `verifier-balance`
  * Add `"prev-balance"` (zero for `AddVerifier`, whereas `"balance"` is zero for `RemoveVerifier`)
  * Add `"client"` (ID) as optional (only used for `AddVerifiedClient`)
* `allocation`
  * Add `"piece-cid"` and `"piece-size"`
  * Add `"term-max"` and `"term-min"`
  * Add `"expiration"`
* `allocation-removed`
  * Add `"piece-cid"` and `"piece-size"`
  * Add `"term-max"` and `"term-min"`
  * Add `"expiration"`
* `claim`
  * Add `"piece-cid"` and `"piece-size"`
  * Add `"sector"` (ID)
  * Add `"term-max"` and `"term-min"`
  * Add `"expiration"`
* `claim-updated`
  * Add `"piece-cid"` and `"piece-size"`
  * Add `"sector"` (ID)
  * Add `"term-max"` and `"term-min"`
  * Add `"prev-term-max"`
  * Add `"expiration"`
* `claim-removed`
  * Add `"piece-cid"` and `"piece-size"`
  * Add `"sector"` (ID)
  * Add `"term-max"` and `"term-min"`
  * Add `"expiration"`

**Market actor events**

No changes, these are complicated because there's so much information possible to include (e.g. all `DealProposal` could be included on all of these events) and can be revisited in a future FIP according to feedback. Hopefully many existing workflows will be sufficient where the market actor is involved.

**Miner actor events**

* `sector-activated`
  * Add `"expiration-epoch"`
  * Add `"claim"` (ID) to couple with each `"piece-cid"` & `"piece-size"` pair where a claim is made
* `sector-updated`
  * Add `"expiration-epoch"`
  * Add `"claim"` (ID) to couple with each `"piece-cid"` & `"piece-size"` pair where a claim is made

### DIFF
--- a/FIPS/fip-0083.md
+++ b/FIPS/fip-0083.md
@@ -104,6 +104,8 @@ The event payload is defined as:
 | Index Key + Value | “id” | <CLAIM_ID> (int)      |
 | Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int) |
 | Index Key + Value | “provider” | <SP_ACTOR_ID> (int)   |
+| Index Key + Value | "piece-cid" | <PIECE_CID> (cid) |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int) |
 
 #### Claim Updated
 This event is emitted when the term of an existing allocation is extended by the client.
--- a/FIPS/fip-0083.md
+++ b/FIPS/fip-0083.md
@@ -12,7 +12,7 @@ created: 2023-11-27
 # FIP-0083: Add built-in Actor events in the Verified Registry, Miner and Market Actors
 
 ## Simple Summary
-This FIP defines and proposes the implementation of a meaningful subset of fire-and-forget externally observable events that will be emitted by the built-in Verified Registry, Storage Market and Storage Miner Actors. These enable external agents (henceforth referred to as *“clients”*) to observe and track a specific subset of on-chain datacap allocation/claims, deal lifecycle and sector lifecycle state transitions.
+This FIP defines and proposes the implementation of a meaningful subset of fire-and-forget externally observable events that will be emitted by the built-in Verified Registry, Storage Market and Storage Miner Actors. These enable external agents (henceforth referred to as *"clients"*) to observe and track a specific subset of on-chain datacap allocation/claims, deal lifecycle and sector lifecycle state transitions.
 
 While this FIP explicitly delineates the implementation of events tied to specific transitions, it does not encompass the full spectrum of potential built-in actor events. The introduction and integration of additional built-in actor events will be addressed in subsequent FIPs.
 
@@ -38,9 +38,9 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
  _Please see [FIP-0049](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0049.md) for a detailed explanation of the FVM event schema and the possible values each field_ 
  in the event schema can take. The interpretation of the flags field in each event entry is as follows:
 
- 1. {"flags": "0x03”} ⇒ Index both key and value of the event entry
- 2. {"flags": "0x02”} ⇒ Index only the value of the event entry
- 3. {"flags": "0x01”} ⇒ Index only the key of the event entry
+ 1. {"flags": "0x03"} ⇒ Index both key and value of the event entry
+ 2. {"flags": "0x02"} ⇒ Index only the value of the event entry
+ 3. {"flags": "0x01"} ⇒ Index only the key of the event entry
 
 
 
@@ -59,11 +59,11 @@ This event is emitted when the balance of a verifier is updated in the Verified 
 
 The event payload is defined as:
 
-| flags                | key | value                                |
-|----------------------| --- |--------------------------------------|
-| Index Key + Value    | “$type” | “verifier-balance” (string)          |
-| Index Key + Value    | “verifier” | <VERIFIER_ACTOR_ID> (int)            |
-| Index Key            | “balance” | <VERIFIER_DATACAP_BALANCE> (bigint) |
+| flags                | key        | value                               |
+| -------------------- | ---------- | ----------------------------------- |
+| Index Key + Value    | "$type"    | "verifier-balance" (string)         |
+| Index Key + Value    | "verifier" | <VERIFIER_ACTOR_ID> (int)           |
+| Index Key            | "balance"  | <VERIFIER_DATACAP_BALANCE> (bigint) |
 
 _The "bigint" CBOR encoding format used for the "balance" field is the same as is used for encoding bigints on the Filecoin chain: a byte array representing a big-endian unsigned integer, compatible with the Golang `big.Int` byte representation, with a `0x00` (positive) or `0x01` (negative) prefix; with a zero-length array representing a value of `0`._
 
@@ -72,25 +72,24 @@ This event is emitted when a verified client allocates datacap to a specific dat
 
 The event payload is defined as:
 
-| flags | key | value                   |
-| --- | --- |-------------------------|
-| Index Key + Value | “$type” | “allocation” (string)   |
-| Index Key + Value | “id” | <ALLOCATION_ID> (int)   |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int) |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)     |
+| flags             | key        | value                   |
+| ----------------- | ---------- | ----------------------- |
+| Index Key + Value | "$type"    | "allocation" (string)   |
+| Index Key + Value | "id"       | <ALLOCATION_ID> (int)   |
+| Index Key + Value | "client"   | <CLIENT_ACTOR_ID> (int) |
+| Index Key + Value | "provider" | <SP_ACTOR_ID> (int)     |
 
 #### Datacap Allocation Removed
 This event is emitted when a datacap allocation that is past its expiration epoch is removed.
 
 The event payload is defined as:
 
-| flags | key | value                         |
-| --- | --- |-------------------------------|
-| Index Key + Value | “$type” | “allocation-removed” (string) |
-| Index Key + Value | “id” | <ALLOCATION_ID> (int)         |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int)       |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)           |
-
+| flags             | key        | value                         |
+| ----------------- | ---------- | ----------------------------- |
+| Index Key + Value | "$type"    | "allocation-removed" (string) |
+| Index Key + Value | "id"       | <ALLOCATION_ID> (int)         |
+| Index Key + Value | "client"   | <CLIENT_ACTOR_ID> (int)       |
+| Index Key + Value | "provider" | <SP_ACTOR_ID> (int)           |
 
 
 #### Allocation Claimed
@@ -98,38 +97,36 @@ This event is emitted when a client allocation is claimed by a storage provider 
 
 The event payload is defined as:
 
-| flags | key | value                 |
-| --- | --- |-----------------------|
-| Index Key + Value | “$type” | “claim” (string)      |
-| Index Key + Value | “id” | <CLAIM_ID> (int)      |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int) |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)   |
-| Index Key + Value | "piece-cid" | <PIECE_CID> (cid) |
-| Index Key         | "piece-size" | <PIECE_SIZE> (int) |
+| flags             | key        | value                   |
+| ----------------- | ---------- | ----------------------- |
+| Index Key + Value | "$type"    | "claim" (string)        |
+| Index Key + Value | "id"       | <CLAIM_ID> (int)        |
+| Index Key + Value | "client"   | <CLIENT_ACTOR_ID> (int) |
+| Index Key + Value | "provider" | <SP_ACTOR_ID> (int)     |
 
 #### Claim Updated
 This event is emitted when the term of an existing allocation is extended by the client.
 
 The event payload is defined as:
 
-| flags | key | value                    |
-| --- | --- |--------------------------|
-| Index Key + Value | “$type” | “claim-updated” (string) |
-| Index Key + Value | “id” | <CLAIM_ID> (int)         |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int)  |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)      |
+| flags             | key        | value                    |
+| ----------------- | ---------- | ------------------------ |
+| Index Key + Value | "$type"    | "claim-updated" (string) |
+| Index Key + Value | "id"       | <CLAIM_ID> (int)         |
+| Index Key + Value | "client"   | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider" | <SP_ACTOR_ID> (int)      |
 
 ####  Claim Removed
 This event is emitted when an expired claim is removed by the Verified Registry Actor.
 
 The event payload is defined as:
 
-| flags | key | value                    |
-| --- | --- |--------------------------|
-| Index Key + Value | “$type” | “claim-removed” (string) |
-| Index Key + Value | “id” | <CLAIM_ID> (int)         |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int)  |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)      |
+| flags             | key        | value                    |
+| ----------------- | ---------- | ------------------------ |
+| Index Key + Value | "$type"    | "claim-removed" (string) |
+| Index Key + Value | "id"       | <CLAIM_ID> (int)         |
+| Index Key + Value | "client"   | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider" | <SP_ACTOR_ID> (int)      |
 
 ### Market Actor Events
 The Market Actor emits the following deal lifecycle events:
@@ -139,11 +136,11 @@ This event is emitted for each new deal that is successfully published by a stor
 
 The event payload is defined as:
 
-| flags | key | value                             |
-| --- | --- |-----------------------------------|
-| Index Key + Value | “$type” | "deal-published" (string)         |
-| Index Key + Value | "id" | <DEAL_ID> (int)                   |
-| Index Key + Value | "client" | <STORAGE_CLIENT_ACTOR_ID> (int)   |
+| flags             | key        | value                             |
+| ----------------- | ---------- | --------------------------------- |
+| Index Key + Value | "$type"    | "deal-published" (string)         |
+| Index Key + Value | "id"       | <DEAL_ID> (int)                   |
+| Index Key + Value | "client"   | <STORAGE_CLIENT_ACTOR_ID> (int)   |
 | Index Key + Value | "provider" | <STORAGE_PROVIDER_ACTOR_ID> (int) |
 
 #### Deal Activated
@@ -151,11 +148,11 @@ The deal activation event is emitted for each deal that is successfully activate
 
 The event payload is defined as:
 
-| flags | key | value                     |
-| --- | --- |---------------------------|
-| Index Key + Value | “$type” | "deal-activated" (string) |
-| Index Key + Value | “id” | <DEAL_ID> (int)           |
-| Index Key + Value | "client" | <STORAGE_CLIENT_ACTOR_ID> (int)   |
+| flags             | key        | value                             |
+| ----------------- | ---------- | --------------------------------- |
+| Index Key + Value | "$type"    | "deal-activated" (string)         |
+| Index Key + Value | "id"       | <DEAL_ID> (int)                   |
+| Index Key + Value | "client"   | <STORAGE_CLIENT_ACTOR_ID> (int)   |
 | Index Key + Value | "provider" | <STORAGE_PROVIDER_ACTOR_ID> (int) |
 
 #### Deal Terminated
@@ -169,11 +166,11 @@ as terminated by the `OnMinerSectorsTerminate` method.
 
 The event payload is defined as:
 
-| flags | key | value                      |
-| --- | --- |----------------------------|
-| Index Key + Value | “$type” | "deal-terminated" (string) |
-| Index Key + Value | “id” | <DEAL_ID> (int)            |
-| Index Key + Value | "client" | <STORAGE_CLIENT_ACTOR_ID> (int)   |
+| flags             | key        | value                             |
+| ----------------- | ---------- | --------------------------------- |
+| Index Key + Value | "$type"    | "deal-terminated" (string)        |
+| Index Key + Value | "id"       | <DEAL_ID> (int)                   |
+| Index Key + Value | "client"   | <STORAGE_CLIENT_ACTOR_ID> (int)   |
 | Index Key + Value | "provider" | <STORAGE_PROVIDER_ACTOR_ID> (int) |
 
 #### Deal Completed Successfully 
@@ -186,11 +183,11 @@ This applies to new deals made post landing FIP-0074. For deals made before FIP-
 
 The event payload is defined as:
 
-| flags | key | value                     |
-| --- | --- |---------------------------|
-| Index Key + Value | “$type” | "deal-completed" (string) |
-| Index Key + Value | “id” | <DEAL_ID> (int)           |
-| Index Key + Value | "client" | <STORAGE_CLIENT_ACTOR_ID> (int)   |
+| flags             | key        | value                             |
+| ----------------- | ---------- | --------------------------------- |
+| Index Key + Value | "$type"    | "deal-completed" (string)         |
+| Index Key + Value | "id"       | <DEAL_ID> (int)                   |
+| Index Key + Value | "client"   | <STORAGE_CLIENT_ACTOR_ID> (int)   |
 | Index Key + Value | "provider" | <STORAGE_PROVIDER_ACTOR_ID> (int) |
 
 ### Miner Actor Events
@@ -201,23 +198,23 @@ The sector pre-committed event is emitted for each new sector that is successful
 
 The event payload is defined as:
 
-| flags | key | value                          |
-| --- | --- |--------------------------------|
-| Index Key + Value | “$type" | "sector-precommitted" (string) |
-| Index Key + Value | “sector” | <SECTOR_NUMER> (int)           |
+| flags             | key      | value                          |
+| ----------------- | -------- | ------------------------------ |
+| Index Key + Value | "$type"  | "sector-precommitted" (string) |
+| Index Key + Value | "sector" | <SECTOR_NUMER> (int)           |
 
 #### Sector Activated
 This event is emitted for each pre-committed sector that is successfully activated by a storage provider. For now, sector activation corresponds 1:1 with prove-committing a sector but this can change in the future.
 
 The event payload is defined as:
 
-| flags             | key            | value                                                        |
-|-------------------|----------------|--------------------------------------------------------------|
-| Index Key + Value | “$type"        | "sector-activated" (string)                                  |
-| Index Key + Value | “sector”       | <SECTOR_NUMER> (int)                                         |
-| Index Key + Value | “unsealed-cid” | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
-| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                            |
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           |
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-activated" (string)                                   |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
+| Index Key + Value | "unsealed-cid" | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
+| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                             |
+| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                            |
 
 - Note that `piece-cid` and `piece-size` is included for each piece in the sector.
 
@@ -226,13 +223,13 @@ This event is emitted for each CC sector that is updated to contained actual sea
 
 The event payload is defined as:
 
-| flags             | key            | value                                                        |
-|-------------------|----------------|--------------------------------------------------------------|
-| Index Key + Value | “$type"        | "sector-updated" (string)                                    |
-| Index Key + Value | “sector”       | <SECTOR_NUMER> (int)                                         |
-| Index Key + Value | “unsealed-cid” | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
-| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                            |
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           | 
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-updated" (string)                                     |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
+| Index Key + Value | "unsealed-cid" | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
+| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                             |
+| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                            | 
 
 - Note that `piece-cid` and `piece-size` is included for each piece in the updated sector.
 
@@ -241,10 +238,10 @@ The sector termination event is emitted for each sector that is marked as termin
 
 The event payload is defined as:
 
-| flags | key | value                        |
-| --- | --- |------------------------------|
-| Index Key + Value | “$type" | "sector-terminated" (string) |
-| Index Key + Value | “sector” | <SECTOR_NUMER> (int)         |
+| flags             | key      | value                        |
+| ----------------- | -------- | ---------------------------- |
+| Index Key + Value | "$type"  | "sector-terminated" (string) |
+| Index Key + Value | "sector" | <SECTOR_NUMER> (int)         |
 
 ### Changing the Market Actor `BatchActivateDealsResult` type
 The legacy (pre [FIP-0076](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0076.md)) sector activation and 
@@ -298,7 +295,7 @@ in and then query the chain state with the corresponding allocation/claim ID to 
 
 ### Market Actor Events
 Once a deal is published, clients have access to the `dealId` of the deal. All Market Actor events except 
-for the “deal published” event have the `dealId` in their payload, which clients can use to filter for events
+for the "deal published" event have the `dealId` in their payload, which clients can use to filter for events
 they are interested in and then query the chain state for more information.
 
 All Market Actor events have the client and provider Actor IDs in the event payload. This enables users
--- a/FIPS/fip-0083.md
+++ b/FIPS/fip-0083.md
@@ -42,8 +42,6 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
  2. {"flags": "0x02"} ⇒ Index only the value of the event entry
  3. {"flags": "0x01"} ⇒ Index only the key of the event entry
 
-
-
  _All values in the event payloads defined below are encoded with CBOR and so the codec field for each event_ 
  entry will be set to 0x51.  The codec field is left empty in the payloads defined below only for the 
  sake of brevity.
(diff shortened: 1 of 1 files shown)


FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
-| flags                | key | value                                |
-|----------------------| --- |--------------------------------------|
-| Index Key + Value    | “$type” | “verifier-balance” (string)          |
-| Index Key + Value    | “verifier” | <VERIFIER_ACTOR_ID> (int)            |
-| Index Key            | “balance” | <VERIFIER_DATACAP_BALANCE> (bigint) |
+| flags                | key             | value                                        |
+| -------------------- | --------------- | -------------------------------------------- |
+| Index Key + Value    | "$type"         | "verifier-balance" (string)                  |
+| Index Key + Value    | "verifier"      | <VERIFIER_ACTOR_ID> (int)                    |
+| Index Key + Value    | "client"        | <DATACAP_HOLDER_ACTOR_ID> (int)              |
```
aarshkshah1992, line 69: @rvagg Should be Option<Int> (nillable)

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| ----------------- | ------------ | ------------------------ |
+| Index Key + Value | "$type"      | "allocation" (string)    |
+| Index Key + Value | "id"         | <ALLOCATION_ID> (int)    |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)       |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)     |
+| Index Key         | "term-max"   | <TERM_MAXIMUM> (int)     |
+| Index Key + Value | "expiration" | <EXPIRATION_EPOCH> (int) |
```
aarshkshah1992, line 92: We should only be indexing this by key. Indexing epoch values will be wasteful.

jennijuju: Can expiration be easily calculated?

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| ----------------- | ------------ | ----------------------------- |
+| Index Key + Value | "$type"      | "allocation-removed" (string) |
+| Index Key + Value | "id"         | <ALLOCATION_ID> (int)         |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)       |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)           |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)             |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)            |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)          |
+| Index Key         | "term-max"   | <TERM_MAXIMUM> (int)          |
+| Index Key + Value | "expiration" | <EXPIRATION_EPOCH> (int)      |
```
aarshkshah1992, line 109: Same as above.

FIPS/fip-0083.md
rvagg: it's on the Allocation type itself in state, so we just lift it out of there https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md#verified-registry-datacap-allocations

FIPS/fip-0083.md
rvagg: made it nullable, may as well follow pattern

jennijuju:
Peer: other than hearing the feedback from various ecosystem partners, i also felt the pain points myself when working on integration docs, for getting simple data 📊 like “how much data is stored in filecoin”

I believe these additional events fields are necessary for toolings like  starboard dashboards,  http://filplus.d.interplanetary.one/, filecoin.tool, https://filecoin-tracker.com/ and so on to build systems that can continue providing CORRECT and more real-time visibility of the network: how much data are onboarded and currently stored; what data is stored; how long the data is stored. 

To only make the minimal change here is needed to not delay the network upgrade, whether built in actors should include more info or not and a more complete tradeoff analysis via another FIP. 

One more piece I think is missing, is a visible  event that can make f05 deal onboarded via ProveCommitSectors observable. Today the event is triggered in cron which is not observable from the clients atm, (1)  message receipts don't include implicit messages (2) internal traces only has call traces, and dont  Enable events viability in cron is a much bigger change that shouldn’t happen in this late of development phase. Without this information, tooling developers will need to add event indexing while keeping the current heavy state inspection, and dedup entries that are captured twice. I wonder, is there anywhere else, outside of the cron we can safely trigger this event?
￼

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| Index Key + Value | "$type"      | "claim" (string)         |
+| Index Key + Value | "id"         | <CLAIM_ID> (int)         |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)       |
+| Index Key + Value | "sector"     | <SECTOR_ID> (int)        |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)     |
+| Index Key         | "term-max"   | <TERM_MAXIMUM> (int)     |
+| Index Key         | "expiration" | <EXPIRATION_EPOCH> (int) |
```
aarshkshah1992, line 127: @rvagg What does expiration mean here ? Expiration exists for allocations where client must use the allocation by a certain epoch. But once an allocation is claimed by an SP, the allocation expiration no longer applies.

I think we can remove this.

aarshkshah1992:
@rvagg There is some data wrangling we need to do to get the claim info available in all the Sector events. Instead, I propose that we add the Miner Actor ID to the claim event in the verif reg Actor. This allows the client to get the same information without us having to add more code in the Miner Actor.

FIPS/fip-0083.md
```diff
@@ -199,50 +232,58 @@ The sector pre-committed event is emitted for each new sector that is successful
...
-| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                            |
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           |
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-activated" (string)                                   |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
+| Index Key + Value | "unsealed-cid" | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
+| Index Key         | "expiration"   | <EXPIRATION_EPOCH> (int)                                      |
+| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                             |
+| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                            |
```
aarshkshah1992, line 254: Let's not add claim info to Miner Actor events. This needs fair bit of code in the Miner Actor. Instead, let's add the Miner Actor ID to the claim event so users can access the same info.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
-| Index Key + Value | “$type” | “allocation-removed” (string) |
-| Index Key + Value | “id” | <ALLOCATION_ID> (int)         |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int)       |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)           |
-
-
-
-#### Allocation Claimed
+| flags             | key          | value                         |
+| ----------------- | ------------ | ----------------------------- |
```
aarshkshah1992, line 100: Let's add the Miner Actor ID to the Allocation Claimed event.

aarshkshah1992:
> Member

Actually, we already have the provider in the claim event. Given that we now have (provider ID, sector number, piece cid and piece size) in the Claim event, I don't think we need to make any changes to the Miner Actor.

aarshkshah1992:
@jennijuju Unfortunately, that is tricky because the proofs submitted via 'ProveCommitSectors' are validated in a cron job and unless those proofs are validated, we don't know if the deals have been successfully "onboarded"/"activated".  So we can't just emit this info/event when the SP calls 'ProveCommitSectors'. We can only emit this after those proofs have been validated but the proofs are validated in a cron job.

jennijuju:
> Actually, we already have the provider in the claim event. Given that we now have (provider ID, sector number, piece cid and piece size) in the Claim event, I don't think we need to add the claim ID to any of the Miner Actor events.

Hrm you get the claim ID right? I think the expiration here meant sector expiration not claim expiration, and sector expiration is already in the sector info and shouldn’t be hard to get.

jennijuju:
> @jennijuju Unfortunately, that is tricky because the proofs submitted via 'ProveCommitSectors' are validated in a cron job and unless those proofs are validated, we don't know if the deals have been successfully "onboarded"/"activated". So we can't just emit this info/event when the SP calls 'ProveCommitSectors'. We can only emit this after those proofs have been validated but the proofs are validated in a cron job.

If we add notification actor ID to sector activated event that may solve the problem as well.

aarshkshah1992:
@jennijuju @rvagg Just to be clear

1. The allocation events have the expiry of the **_allocation_**
2. The claim events have the expiry of the **_claim_** i..e. 

```
impl Expires for Claim {
    fn expiration(&self) -> ChainEpoch {
        self.term_start + self.term_max
    }
}

```

This is the epoch after which the SP stops getting QA power for the claimed sector

3. The sector  activation and sector update events in the Miner Actor have the expiry of the **_sector_**

Stebalien:
> If we add notification actor ID to sector activated event that may solve the problem as well.

_Summarizing my objections from the call:_

I'm going to object to this pretty strongly. We're asking the user to _infer_ that, because the market actor was notified, the deal necessarily "belongs" to the market actor. The SP can choose to notify any actor and and any number of actors and who they notify won't necessarily be the "storage market" that owns the deal (and I really don't want users to start assuming this). E.g., the SP could choose to notify f05 even if the deal isn't related to the market.

_Proposal from the call:_

1. Add the piece ID, size, and sector ID to the deal activation event.
2. Each time a piece CID + size tuple is mentioned in a deal activation event, ignore one copy of the same piece CID + size from the equivalent sector activation event. These should always happen in the same message, so deduplicating these shouldn't be too difficult.

NOTE: It doesn't matter _which_ piece we ignore from the sector activation event, we don't have IDs anyways.

Stebalien:
_Proposal from a followup conversation with @aarshkshah1992:_

We don't actually need to add anything to the deal activation event. Instead, we can:

1. Record the activated pieces.
2. Record the activated deals (we don't need to associate them with the pieces).
3. Then, when we diff the market state, we can check the database and see that we've already accounted for some of the deals (step 2) and avoid double-counting those pieces.

This will let us get away with adding nothing to our events.

jennijuju:
@Stebalien, like I said, I will only insist on this if there is a solution for the builder that's simple enough to build. BUT last bikeshedding.

The original comment called it the **notification actor ID**, not the storage market ID, for a reason. So, I am not convinced by your "I read builders mind" statement here. A more productive thing would be defining and aligning standards with the builders/ecosystem. 

I am also unconvinced that "the SP could choose to notify f05 even if the deal isn't related to the market." What is the incentive/use case for any client/SP to ever DO that? Who would anyone send any notification if they don't want to have their data deal to participate in some storage market/marketplace or leverage some on-chain payment mechanism? It would be great if you could follow up on #fip-direct-data-onboarding cuz I'm now concerned about why we are not aligned on why we should have this notification design in the protocol and why it is useful.

jennijuju:
> Then, when we diff the market state, we can check the database and see that we've already accounted for some of the deals (step 2) and avoid double-counting those pieces.

@aarshkshah1992 @Stebalien, what's your plan to check this? Where do you get the pieces from and what information other than piece CID will you be using, and the key pairs for checking in the deal DB?

anorth:
A reminder to you all that most of this discussion belongs in the discussion thread https://github.com/filecoin-project/FIPs/discussions/754, where it has the context of prior discussion, can support a level of threading, and will persist and be discoverable after this PR is closed.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)       |
+| Index Key + Value | "sector"     | <SECTOR_ID> (int)        |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)     |
+| Index Key         | "term-max"   | <TERM_MAXIMUM> (int)     |
+| Index Key         | "expiration" | <EXPIRATION_EPOCH> (int) |
+
+#### Datacap Claim Updated
 This event is emitted when the term of an existing allocation is extended by the client.
```
anorth, line 130: Nit: "by the client" is misleading here. It might not be by the original client.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| Index Key + Value | "id"            | <CLAIM_ID> (int)              |
+| Index Key + Value | "client"        | <CLIENT_ACTOR_ID> (int)       |
+| Index Key + Value | "provider"      | <SP_ACTOR_ID> (int)           |
+| Index Key + Value | "piece-cid"     | <PIECE_CID> (cid)             |
+| Index Key         | "piece-size"    | <PIECE_SIZE> (int)            |
+| Index Key + Value | "sector"        | <SECTOR_ID> (int)             |
+| Index Key         | "term-min"      | <TERM_MINIMUM> (int)          |
+| Index Key         | "term-max"      | <TERM_MAXIMUM> (int)          |
+| Index Key         | "prev-term-max" | <PREVIOUS_TERM_MAXIMUM> (int) |
+| Index Key         | "expiration"    | <EXPIRATION_EPOCH> (int)      |
```
anorth, line 146: What's this expiration? Edit: ditto replace with `term_start`.

FIPS/fip-0083.md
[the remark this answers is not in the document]
anorth: Update: Ok I see from the discussion thread that this is representing the computed value `term_start + term_max`. This strikes me as an odd way to do it, and confusing given expiration means something different for allocations. It would be better to include the `term_start`, corresponding to the claim's actual state.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
-####  Claim Removed
+| flags             | key             | value                         |
+| ----------------- | --------------- | ----------------------------- |
+| Index Key + Value | "$type"         | "claim-updated" (string)      |
+| Index Key + Value | "id"            | <CLAIM_ID> (int)              |
+| Index Key + Value | "client"        | <CLIENT_ACTOR_ID> (int)       |
+| Index Key + Value | "provider"      | <SP_ACTOR_ID> (int)           |
+| Index Key + Value | "piece-cid"     | <PIECE_CID> (cid)             |
+| Index Key         | "piece-size"    | <PIECE_SIZE> (int)            |
+| Index Key + Value | "sector"        | <SECTOR_ID> (int)             |
```
anorth, line 142: In the future, `sector` is another field that may be updated for a claim. I guess we could add `prev-sector` at that point without too much trouble. But see discussion https://github.com/filecoin-project/FIPs/discussions/754#discussioncomment-8833445 about whether we should include prev values at all.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| Index Key + Value | "$type"      | "claim-removed" (string) |
+| Index Key + Value | "id"         | <CLAIM_ID> (int)         |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)       |
+| Index Key + Value | "sector"     | <SECTOR_ID> (int)        |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)     |
+| Index Key         | "term-max"   | <TERM_MAXIMUM> (int)     |
+| Index Key         | "expiration" | <EXPIRATION_EPOCH> (int) |
```
anorth, line 164: Ditto `term-start`

FIPS/fip-0083.md
```diff
@@ -199,50 +232,58 @@ The sector pre-committed event is emitted for each new sector that is successful
...
-| Index Key + Value | “sector”       | <SECTOR_NUMER> (int)                                         |
-| Index Key + Value | “unsealed-cid” | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
-| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                            |
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           |
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-activated" (string)                                   |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
+| Index Key + Value | "unsealed-cid" | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
+| Index Key         | "expiration"   | <EXPIRATION_EPOCH> (int)                                      |
```
anorth, line 252: I think we should avoid adding this, it will change. https://github.com/filecoin-project/FIPs/discussions/754#discussioncomment-8833619

FIPS/fip-0083.md
```diff
@@ -199,50 +232,58 @@ The sector pre-committed event is emitted for each new sector that is successful
...
-| flags | key | value                          |
-| --- | --- |--------------------------------|
-| Index Key + Value | “$type" | "sector-precommitted" (string) |
-| Index Key + Value | “sector” | <SECTOR_NUMER> (int)           |
+| flags             | key                | value                                                         |
+| ----------------- | ------------------ | ------------------------------------------------------------- |
+| Index Key + Value | "$type"            | "sector-precommitted" (string)                                |
+| Index Key + Value | "sector"           | <SECTOR_NUMER> (int)                                          |
+
+* `deal` is included for each deal in the sector, i.e. it may appear zero or more times in the event payload.
```
anorth, line 240: Stray line referring to something from a prior draft?

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| Index Key + Value    | "$type"         | "verifier-balance" (string)                  |
+| Index Key + Value    | "verifier"      | <VERIFIER_ACTOR_ID> (int)                    |
+| Index Key + Value    | "client"        | <DATACAP_HOLDER_ACTOR_ID> (nullable int)     |
+| Index Key            | "prev-balance"  | <PREVIOUS_VERIFIER_DATACAP_BALANCE> (bigint) |
+| Index Key            | "balance"       | <VERIFIER_DATACAP_BALANCE> (bigint)          |
 
-_The "bigint" CBOR encoding format used for the "balance" field is the same as is used for encoding bigints on the Filecoin chain: a byte array representing a big-endian unsigned integer, compatible with the Golang `big.Int` byte representation, with a `0x00` (positive) or `0x01` (negative) prefix; with a zero-length array representing a value of `0`._
+* `client` will be `null` for `AddVerifier` and `RemoveVerifier` methods but will be an int for the `AddVerifiedClient` method and indicates the actor ID of the receiver of the datacap.
+* `prev-balance` will be zero for the `AddVerifier` method.
+* `balance` will be zero for the `RemoveVerifier` method.
```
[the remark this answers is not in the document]
rvagg: Removed notes but added a note about non-null case in the table, like the `unsealed-cid` note in `sector-activated` and `sector-updated`. We can continue discussion about the merits of this field in https://github.com/filecoin-project/FIPs/discussions/754#discussioncomment-8833387.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
-| Index Key + Value | “id” | <ALLOCATION_ID> (int)   |
-| Index Key + Value | “client” | <CLIENT_ACTOR_ID> (int) |
-| Index Key + Value | “provider” | <SP_ACTOR_ID> (int)     |
+| flags             | key          | value                    |
+| ----------------- | ------------ | ------------------------ |
+| Index Key + Value | "$type"      | "allocation" (string)    |
+| Index Key + Value | "id"         | <ALLOCATION_ID> (int)    |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
```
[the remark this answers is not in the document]
rvagg: They are already marked as indexable in `sector-activated` and `sector-updated`, I'm not sure there's a good case for having them indexed in one place but not another. There's a solid audience that cares about piece CID even if it ideally should be an internal detail.

TBQH I don't see much merit in the flags field at all and would rather it not be there. Consumers of this information should add their own indexes. There's going to end up being a lot of information coming out of these and I'm not convinced of a case where someone's going to globally obey these hints. Better have standardised names of these things and well-formed schemas of events so that consumers can pick and choose and global providers can index only the well-known things.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,118 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
-#### Claim Updated
+| flags             | key          | value                    |
+| ----------------- | ------------ | ------------------------ |
+| Index Key + Value | "$type"      | "claim" (string)         |
+| Index Key + Value | "id"         | <CLAIM_ID> (int)         |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)       |
+| Index Key + Value | "sector"     | <SECTOR_ID> (int)        |
```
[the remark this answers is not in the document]
rvagg: see comment above, they are already marked as indexable on all of the `sector-*` events, do we not index it here because it's not a primary concern of this event (unlike claim ID?). How do we form a rule about what's indexable and what's not?

FIPS/fip-0083.md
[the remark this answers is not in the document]
rvagg: no, I think you're right and this was a mixup on my part from looking at the code. `term_start` is the epoch of the event which the user should have from other sources so I don't think there's a good reason to include expiry or `term_start`. Removed.

FIPS/fip-0083.md
rvagg: 👍 removed entirely

FIPS/fip-0083.md
rvagg: removed

FIPS/fip-0083.md
rvagg: removed

FIPS/fip-0083.md
rvagg: OK, sad. I thought this was going to be uncomplicated because it can be derived from the piece activation manifest but I suppose the legacy methods make this complicated?

aarshkshah1992:
> > If we add notification actor ID to sector activated event that may solve the problem as well.
> 
> _Summarizing my objections from the call:_
> 
> I'm going to object to this pretty strongly. We're asking the user to _infer_ that, because the market actor was notified, the deal necessarily "belongs" to the market actor. The SP can choose to notify any actor and and any number of actors and who they notify won't necessarily be the "storage market" that owns the deal (and I really don't want users to start assuming this). E.g., the SP could choose to notify f05 even if the deal isn't related to the market.
> 
> _Proposal from the call:_
> 
> 1. Add the piece ID, size, and sector ID to the deal activation event.
> 2. Each time a piece CID + size tuple is mentioned in a deal activation event, ignore one copy of the same piece CID + size from the equivalent sector activation event. These should always happen in the same message, so deduplicating these shouldn't be too difficult.
> 
> NOTE: It doesn't matter _which_ piece we ignore from the sector activation event, we don't have IDs anyways.

@Stebalien @anorth @rvagg This sounds like the most promising solution. The only problem here is that we still miss out on data in sectors onboarded via the legacy `ProveCommitSectors` method as the corresponding deal activation events will be emitted in cron and will not be available to clients. If we're okay with losing just that bit of de-duping, this solution works.

rvagg:
btw I pushed this change:

* Remove notes from verifier-balance
* Remove expiration from claim*
* Remove stray comment

aarshkshah1992:
@rvagg Do we need the verifier balance event anymore ?

FIPS/fip-0083.md
aarshkshah1992: @rvagg Yeah and the users can anyways get all this info from the claim event as the claim event has the (miner ID, piece cid, piece size, sector number) already. So really, adding the claim ID here dosen't get them anything new.

FIPS/fip-0083.md
```diff
@@ -199,50 +225,56 @@ The sector pre-committed event is emitted for each new sector that is successful
...
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           | 
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-updated" (string)                                     |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
+| Index Key + Value | "unsealed-cid" | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
+| Index Key         | "expiration"   | <EXPIRATION_EPOCH> (int)                                     |
+| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                             |
+| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                            |
+| Index Key + Value | "claim"        | <CLAIM_ID> (cid)                                              |
```
aarshkshah1992, line 264: Same as above, we don't need the claim ID here as the claim event will suffice.

rvagg:
* Remove "claim" from sector-* events, non-trivial to gather and is available in reverse from the claim event.
* Remove prev-* fields due to objections

FIPS/fip-0083.md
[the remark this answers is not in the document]
anorth: Ok to leave indexed then

anorth:
This looks good to me at the moment with the exception of the verifier-balance event [discussed here](https://github.com/filecoin-project/FIPs/discussions/754#discussioncomment-8846370).

Note I don't object to adding more simple fields (piece id, size, sector) to market events. I don't think we should attempt to do anything special for the unobservable events from legacy prove commit sector though.

Stebalien:
@jennijuju

> The original comment called it the notification actor ID, not the storage market ID, for a reason. So, I am not convinced by your "I read builders mind" statement here. A more productive thing would be defining and aligning standards with the builders/ecosystem.
> 
> ...
> 
> I am also unconvinced that "the SP could choose to notify f05 even if the deal isn't related to the market." What is the incentive/use case for any client/SP to ever DO that?

Predicting how an API might be misused is an important part of API design (and UX, in general). In this case, my concern is that users may try to use this information the same way you are, but with more severe consequences if something goes wrong (and therefore more incentive for SPs to send bad notifications).

> @aarshkshah1992 @Stebalien, what's your plan to check this? Where do you get the pieces from and what information other than piece CID will you be using, and the key pairs for checking in the deal DB?

We don't need the piece CIDs for deduplication, just the deal IDs from the deal activation events (and later, the deal IDs in the market state).

The key difference between our two proposals is which information we keep:

- Your proposal: Drop information from the events that we expect to be present in the market state diff.
- My proposal: When processing the events, record everything. Then, when we diff the market state, skip the deals we've already processed).

This actually means less work in the future. When we drop support for the old `ProveCommit` method, indexers can simply delete the market diffing code because the event processing code will have all the information.

(and yeah, we should have had this discussion in a threaded forum)

Stebalien:
@aarshkshah1992 

Every solution we've discussed involves diffing the market state to detect deals from ProveCommit. The goal here is to avoid double counting.

I'd actually prefer the last solution I proposed for the reason I stated in my previous comment.

rvagg:
Removed `"client"` from `verifier-balance`, will move this to a separate PR because the discussion is of a different nature and I don't want that to block these less controversial changes.

Updated original post to detail just the changes here now.

FIPS/fip-0083.md
```diff
@@ -199,50 +222,50 @@ The sector pre-committed event is emitted for each new sector that is successful
...
 The event payload is defined as:
 
-| flags | key | value                          |
-| --- | --- |--------------------------------|
-| Index Key + Value | “$type" | "sector-precommitted" (string) |
-| Index Key + Value | “sector” | <SECTOR_NUMER> (int)           |
+| flags             | key                | value                                                         |
+| ----------------- | ------------------ | ------------------------------------------------------------- |
+| Index Key + Value | "$type"            | "sector-precommitted" (string)                                |
+| Index Key + Value | "sector"           | <SECTOR_NUMER> (int)                                          |
```
TippyFlitsUK, line 228: Should this be `SECTOR_NUMBER`?

FIPS/fip-0083.md
```diff
@@ -199,50 +222,50 @@ The sector pre-committed event is emitted for each new sector that is successful
...
-|-------------------|----------------|--------------------------------------------------------------|
-| Index Key + Value | “$type"        | "sector-activated" (string)                                  |
-| Index Key + Value | “sector”       | <SECTOR_NUMER> (int)                                         |
-| Index Key + Value | “unsealed-cid” | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
-| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                            |
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           |
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-activated" (string)                                   |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
```
TippyFlitsUK, line 238: Should this be `SECTOR_NUMBER`?

FIPS/fip-0083.md
```diff
@@ -199,50 +222,50 @@ The sector pre-committed event is emitted for each new sector that is successful
...
-|-------------------|----------------|--------------------------------------------------------------|
-| Index Key + Value | “$type"        | "sector-updated" (string)                                    |
-| Index Key + Value | “sector”       | <SECTOR_NUMER> (int)                                         |
-| Index Key + Value | “unsealed-cid” | <SECTOR_COMMD> (nullable cid) (null means sector has no data) |
-| Index Key + Value | "piece-cid"    | <PIECE_CID> (cid)                                            |
-| Index Key         | "piece-size"   | <PIECE_SIZE> (int)                                           | 
+| flags             | key            | value                                                         |
+| ----------------- | -------------- | ------------------------------------------------------------- |
+| Index Key + Value | "$type"        | "sector-updated" (string)                                     |
+| Index Key + Value | "sector"       | <SECTOR_NUMER> (int)                                          |
```
TippyFlitsUK, line 253: Should this be `SECTOR_NUMBER`?

FIPS/fip-0083.md
```diff
@@ -199,50 +222,50 @@ The sector pre-committed event is emitted for each new sector that is successful
...
 The event payload is defined as:
 
-| flags | key | value                        |
-| --- | --- |------------------------------|
-| Index Key + Value | “$type" | "sector-terminated" (string) |
-| Index Key + Value | “sector” | <SECTOR_NUMER> (int)         |
+| flags             | key       | value                        |
+| ----------------- | --------- | ---------------------------- |
+| Index Key + Value | "$type"   | "sector-terminated" (string) |
+| Index Key + Value | "sector"  | <SECTOR_NUMER> (int)         |
```
TippyFlitsUK, line 268: Should this be `SECTOR_NUMBER`?

TippyFlitsUK (approved):

jennijuju:
>  I don't think we should attempt to do anything special for the unobservable events from legacy prove commit sector though.

@anorth @Stebalien 👍 I hear you and I agree we have other alternatives thats better.

FIPS/fip-0083.md
```diff
@@ -52,82 +50,108 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
-#### Allocation Claimed
+| flags             | key          | value                         |
+| ----------------- | ------------ | ----------------------------- |
+| Index Key + Value | "$type"      | "allocation-removed" (string) |
+| Index Key + Value | "id"         | <ALLOCATION_ID> (int)         |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)       |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)           |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)             |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)            |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)          |
```
jennijuju, line 101: hrm from the discussion - https://github.com/filecoin-project/FIPs/pull/964#discussion_r1529337442 & others, should these be term start && remove expiration?

FIPS/fip-0083.md
```diff
@@ -52,82 +50,108 @@ Filecoin network observability tools, block explorers, SP monitoring dashboards,
...
+| ----------------- | ------------ | ------------------------ |
+| Index Key + Value | "$type"      | "allocation" (string)    |
+| Index Key + Value | "id"         | <ALLOCATION_ID> (int)    |
+| Index Key + Value | "client"     | <CLIENT_ACTOR_ID> (int)  |
+| Index Key + Value | "provider"   | <SP_ACTOR_ID> (int)      |
+| Index Key + Value | "piece-cid"  | <PIECE_CID> (cid)        |
+| Index Key         | "piece-size" | <PIECE_SIZE> (int)       |
+| Index Key         | "term-min"   | <TERM_MINIMUM> (int)     |
+| Index Key         | "term-max"   | <TERM_MAXIMUM> (int)     |
+| Index Key         | "expiration" | <EXPIRATION_EPOCH> (int) |
```
jennijuju, line 86: shouldn't expiration be replaced with term start?

jennijuju:
Editor: thanks all for the user feedback and peer reviews! These changes LGTM, other than the verifreg expiration fields which I believe was suggested to be replaced. If that's changed,  @rvagg @aarshkshah1992 can you link us the latest thread where alignment was made?

FIPS/fip-0083.md
anorth: This is an allocation, not a claim. Allocations haven't started yet, but do expire.

FIPS/fip-0083.md
anorth: That comment is about claims, not allocations.

anorth (approved):
I took the editorial liberty of fixing the typos.

rvagg:
> other than the verifreg expiration fields which I believe was suggested to be replaced 

Discussion inline at: https://github.com/filecoin-project/FIPs/pull/964#discussion_r1529337442

`"expiration"` is left on the `allocation*` events because an allocation has an explicit expiration. The expiration for a claim is `term_start` + `term_max` where `term_start` is the height at which the event is emitted so it seems like a duplicate piece of information. So there is no `"expiration"` on the `claim*` events. Happy to add `term_start` if someone thinks that's a value-add though, it's simple enough but feels duplicative to me.

anorth:
I'm happy to add `term_start`. It's not duplicative for an update event (and would require maintaining history to calculate)

rvagg:
Added `"term-start"` to all `claim*` events

jsoares (approved):

anorth (approved):

PR #967, opened 2024-03-19
Fip0081 gradual activation

kkarrancsu:
This pull request updates FIP0081 to include a ramped activation.  

Pseudocode is provided for how a ramped activation could be realized. The pseudocode is written from the perspective of how to set the Gamma value (`gamma_day`) for each day of the network. This implies that there is a controller that is setting the Gamma value for each day and that controller would be active forever. It is possible to `break` out of the loop after Gamma reaches the target value of 0.7.

This pseudocode omits several realities that need to be taken into account when implementing in the network, including:

- Absolute datetime or epoch numbers, due to the need for this to work in both testnet and mainnet
- What entity/object is computing and setting the Gamma value
- Whether Gamma changes every epoch or every day during the ramping period


FIPS/fip-0081.md
```diff
@@ -79,10 +79,27 @@ a reasonable lower bound on network collateral.
...
 Factor the per-sector initial consensus pledge into two parts: a 70% share which divides by the baseline function like today,
-and a smaller 30% "simple" share which does not divide by the baseline function.
+and a smaller 30% "simple" share which does not divide by the baseline function.  Gradually activate the change from the current
+to the target.
 
 ```
-SectorInitialConsensusPledge = PledgeLockTarget * (0.3 * SectorSimpleShare + 0.7 * SectorBaselineShare) 
+activation_len_days = int(365)
+current_day = 0
+update_day = (datetime(2024,6,1) - datetime.today()).days
```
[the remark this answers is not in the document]
kkarrancsu: Yep - that was what was intended here, let me update.

jsoares (commented):
I might be missing something obvious but introducing a state machine here seems like unnecessary complexity and state, e.g. 
* Do we really need a pre-activation state given it's... erm... pre-activation?
* Why not compute a gamma directly from activation, length, and date, thereby avoiding not just the state but the numerical considerations you mention? Something like $\gamma=0.7+\min(\max(0.3*\frac{(FIP \textunderscore activation \textunderscore date+activation \textunderscore len \textunderscore days) -datettime.today().days}{activation \textunderscore len \textunderscore days},0),0.3)$

Also, the while loop is underspecified as it could run with any frequency, but that's a bit pedantic esp. given your note on needing to choose a period. Regardless of the granularity, this should likely all be defined in epochs (i.e. activation epoch, activation_len_epochs, ...) to not be sensitive to clocks.

kkarrancsu:
> I might be missing something obvious but introducing a state machine here seems like unnecessary complexity and state, e.g.

I don't disagree with anything you've stated, but this is just pseudocode.  It likely has no correlation to how this would actually be implemented in Lotus, since things such as absolute dates are not available. I wrote it as a state machine because I thought it would be easy to understand.

kaitlin-beegle (approved):

FIPS/fip-0081.md
```diff
@@ -79,10 +79,28 @@ a reasonable lower bound on network collateral.
 
 ## Specification
 Factor the per-sector initial consensus pledge into two parts: a 70% share which divides by the baseline function like today,
-and a smaller 30% "simple" share which does not divide by the baseline function.
+and a smaller 30% "simple" share which does not divide by the baseline function.  Gradually activate the change from the current
+to the target.
```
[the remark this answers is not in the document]
kkarrancsu: yep, added, but if unclear or could use further revision lmk and I can update!

anorth (changes requested):
Thanks. This is helpful in communicating the high level intention of a phase-in to humans. However the pseudocode as written can't function as a specification for implementers due to assuming a context for computation that doesn't match the chain's (which the PR description acknowledges). As such, I don't think we should merge this.

If we do want to merge something as a stepping stone to a full specification, I'd suggest avoiding pseudocode and just writing prose, with a TODO to actually specify how to implement that intent.

ipfsforcezuofu:
As suggested by @Lucky N. _FF. The PR is closed for now and will be reopened once the implementation details are finalized.

kkarrancsu:
I've updated the description and removed the pseudocode. Is it clearer now?

FIPS/fip-0081.md
```diff
@@ -236,9 +281,5 @@ so the supply of storage is more than adequate to demand.
 
 To be provided.
 
-## TODO
-
-- Implementation
```
anorth, line 241: This is still TODO. The description of intent is clear enough, but this currently lacks specification of how to achieve it.

anorth (commented):
I tried to view the attached PDF, but could only see 9 blank pages. Can someone else check this? @kkarrancsu do you want to try sending me the file by another channel too?

jsoares:
> > I tried to view the attached PDF, but could only see 9 blank pages. Can someone else check this? @kkarrancsu do you want to try sending me the file by another channel too?
> 
> Hmm, not sure what happened there but I pushed a new pdf that is hopefully uncorrupted. Will also share directly with you.

The new one works for me.

anorth:
New PDF 👍

kkarrancsu:
I've added a link to the draft implementation in the [builtin-actors](https://github.com/filecoin-project/builtin-actors/pull/1557) repository.  Is it OK to move this FIP forward for consideration, while we continue to refine the draft implementation?

kkarrancsu:
> I've added a link to the draft implementation in the [builtin-actors](https://github.com/filecoin-project/builtin-actors/pull/1557) repository. Is it OK to move this FIP forward for consideration, while we continue to refine the draft implementation?

Hi @jennijuju - gentle ping on this post Eth-CC!

anorth:
> I've added a link to the draft implementation in the [builtin-actors](https://github.com/filecoin-project/builtin-actors/pull/1557) repository. Is it OK to move this FIP forward for consideration, while we continue to refine the draft implementation?

What do you mean by move this FIP forward? 

My original objection here still stands. The FIP does not specify how to achieve its outcome. From the draft implementation, I know that it has now been worked out how this can be achieved, but it's not represented in this document. This needs to be specified in the FIP. Linking to the code doesn't suffice for this.

I can lift my objection and allow this to merge as a draft, but the FIP will still not be ready for last call, so I don't want to mislead by doing that.

jennijuju:
@kkarrancsu  - i agree with anorth, lmk if you want to do a pass tgt sometime this week!

kkarrancsu:
> @kkarrancsu - i agree with anorth, lmk if you want to do a pass tgt sometime this week!

Hi @jennijuju - I've updated the Specification section to include a description of the necessary changes.  Can you review and also help me understand how to think about Migration?

FIPS/fip-0081.md
```diff
@@ -14,199 +14,171 @@ spec-sections:
...
+To achieve a ramped activation, vary the gamma parameter linearly from 1.0 at the date of activation to 0.7, on the date when the ramped activation will be completed.  
+
+![ramped_activation_gamma.png](../resources/fip-0081/ramped_activation_gamma.png)
+
+Note that the implemented calculation should be re-arranged to defer division until the final operations, to avoid loss of precision in rounding.
+
+The practical effect of this is that the sector initial pledge calculation will not tend to zero as the network baseline grows,  but toward the `SectorInitialConsensusPledge` (30% of the pre-baseline-crossing total). When/if the baseline function exceeds network QAP, the per-sector initial pledge value will still decrease exponentially, but to this floor rather than toward zero.
+
+### Built-in Actors Change Specification
+In this section, we describe the changes required to the `builtin-actors` necessary to implement this update.  The high level changes that need to occur is that the ramp configuration needs to be passed to the function which computes the consensus pledge.
```
anorth, line 82: Nit: first sentence is redundant.

FIPS/fip-0081.md
```diff
@@ -14,199 +14,171 @@ spec-sections:
...
+
+The ramp configuration consists of:
+1. The epoch when the ramp starts (`ramp_start_epoch: i64`)
+2. The duration of the ramp in epochs (`ramp_duration_epochs: u64`)
+
+To do this, the following changes need to be made:
+
+### State Data Structures
+**Market Actor**
+1. `CurrentTotalPowerReturnParams` - Add ramp configuration parameters defined above to the `CurrentTotalPowerReturnParams` structure.
```
anorth, line 92: This is just a copy of the parameter/return shape defined by the power actor. I think it's sufficient to define the change for the power actor only. Clearly the actors that call into it need to use the right schema. The existence of this copy of the structure is just a code artefact.

FIPS/fip-0081.md
```diff
@@ -14,199 +14,171 @@ spec-sections:
...
+### Built-in Actors Change Specification
+In this section, we describe the changes required to the `builtin-actors` necessary to implement this update.  The high level changes that need to occur is that the ramp configuration needs to be passed to the function which computes the consensus pledge. 
+
+The ramp configuration consists of:
+1. The epoch when the ramp starts (`ramp_start_epoch: i64`)
+2. The duration of the ramp in epochs (`ramp_duration_epochs: u64`)
+
+To do this, the following changes need to be made:
+
+### State Data Structures
```
anorth, line 90: The text below addresses both state and method parameter/return structures.

FIPS/fip-0081.md
```diff
@@ -14,199 +14,171 @@ spec-sections:
...
+1. `CurrentTotalPowerReturnParams` - Add ramp configuration parameters defined above to the `CurrentTotalPowerReturnParams` structure.
 
-The practical effect of this is that the sector initial pledge calculation will not tend to zero
-as the network baseline grows, 
-but toward the SectorSimpleShare (30% of the pre-baseline-crossing total).
-When/if the baseline function exceeds network QAP, the per-sector initial pledge value will still decrease exponentially,
-but to this floor rather than toward zero.
+**Miner Actor**
+1. `NetworkPledgeInputs` - Add ramp configuration parameters defined above to the `NetworkPledgeInputs` structure.
+2. `CurrentTotalPowerReturn` - Add ramp configuration parameters defined above to the `CurrentTotalPowerReturn` structure.
```
anorth, line 96: Ditto

FIPS/fip-0081.md
```diff
@@ -14,199 +14,171 @@ spec-sections:
...
+
+Finally, the changes described in the Miner Actor for the `initial_pledge_for_power` update the actual pledge computation.
+
+### Motivation for Ramped Activation
+The proposed update for `SectorInitialConsensusPledge` requires setting the gamma parameter to 0.7, which results in 70% of the `SectorInitialConsensusPledge` coming from `BaselineConsensusPledge`, and 30% coming from `SimpleConsensusPledge`. 
+
+Under the updated formula for `SectorInitialConsensusPledge`, the network's current setting is `gamma=1.0`, which results in all of the `SectorInitialConsensusPledge` coming from `BaselineConsensusPledge`.
+
+We recommend updating gamma from its current value of 1.0 to the target value of 0.7 over an period of 1Y. The technical reason for this recommendation is that gamma has a direct effect on pledge - changing this gradually will prevent pledge from sharply increasing, which will negatively affect SPs. There are additional economic considerations which support this recommendation and are discussed [here](../resources/fip-0081/gradual_activation_rationale.pdf).
```
anorth, line 123: No "we recommend" in a FIP. Just clearly define what the specification is. The fact that the FIP authors recommend it is implied.

kkarrancsu:
Hi @anorth @jennijuju  - I think that the remaining changes are completed, can you take a look to see if it's ready to move to the next step?

anorth (approved):

jsoares (approved):

FIPS/fip-0081.md
```diff
@@ -14,199 +14,175 @@ spec-sections:
...
-Factor the per-sector initial consensus pledge into two parts: a 70% share which divides by the baseline function like today,
-and a smaller 30% "simple" share which does not divide by the baseline function.
+
+### Pseudocode
+Factor the `SectorInitialConsensusPledge` pledge into two parts, the `SimpleConsensusPledge` and the `BaselineConsensusPledge`, defined as follows:
 
 ```
-SectorInitialConsensusPledge = PledgeLockTarget * (0.3 * SectorSimpleShare + 0.7 * SectorBaselineShare) 
+BaselineConsensusPledge = PledgeLockTarget * SectorQAP / max(NetworkBaseline, NetworkQAP)
+SimpleConsensusPledge = PledgeLockTarget * SectorQAP / max(NetworkBaseline)
```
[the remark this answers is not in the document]
kkarrancsu: This is a pseudocode representation of the SectorInitialConsensusPledge formula given [here](https://spec.filecoin.io/#section-systems.filecoin_mining.miner_collaterals.initial-pledge-collateral).

What was intended to be conveyed here is - we take whatever is larger, either the Baseline or the current QAP to compute the SimpleConsensusPledge.  

They are both scalar values, but what is meant by "scalar and can't be maxed" - I don't think I understand what is mean't to be conveyed here.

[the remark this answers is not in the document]
jsoares: The `max()` function returns the larger of _at least_ two values. Even assuming that it accepts a single value, then `max(a)=a`, so it's just visual pollution.

Unless I'm misunderstanding the type of `NetworkBaseline` but I think you confirmed my understanding.

It makes sense in line 61 (and in the original SectorInitialConsensusPledge) because there are, indeed, two arguments.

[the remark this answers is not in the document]
kkarrancsu: Oh - I see my mistake now, will correct.  Thanks for the flag.

PR #975, opened 2024-03-25
FIP-0086 Propagate some implementation decisions back to the spec.

anorth:
- Update the message schemas to match implementation
- Remove tipset weight
- Removed definition of weak quorum
- Added justification of DECIDE
- Replace short-circuit DECIDE pseudocode with description of optimisation
- Some (partial) fixes to message validation and justification
- Remove drand from timing
- Added some TODOs

@ranchalp please review.

### DIFF
--- a/FIPS/fip-0086.md
+++ b/FIPS/fip-0086.md
@@ -225,16 +225,16 @@ GossiPBFT was designed with the Filecoin network in mind and presents a set of f
 
 #### Message format, signatures, and equivocation
 
-Messages include the following fields: $\langle Sender, Signature, MsgType, Value, Instance, [Round, Justification, Ticket] \rangle$. As $Round$, Justification, and $Ticket$ are fields that not all message types require, when not required by a message type, their default value is used (i.e. $0$, $nil$, and $[] byte \lbrace\rbrace$, respectively). We refer to a _field_ of message $m$, with $m.field$:
+Messages include the following fields: $\langle Sender, Signature, MsgType, Value, Instance, [Round, Evidence, Ticket] \rangle$. As $Round$, $Evidence$, and $Ticket$ are fields that not all message types require, when not required by a message type, their default value is used (i.e. $0$, $nil$, and $[] byte \lbrace\rbrace$, respectively). We refer to a _field_ of message $m$, with $m.field$:
 
 ```
-// A message in the Granite protocol.
+// A message in the GossiPBFT protocol.
 // The same message structure is used for all rounds and phases.
 // Note that the message is self-attesting so no separate envelope or signature is needed.
 // - The signature field fixes the included sender ID via the implied public key;
 // - The signature payload includes all fields a sender can freely choose;
 // - The ticket field is a signature of the same public key, so also self-attesting.
-type GMessage struct {
+type GossiPBFTMessage struct {
   // ID of the sender/signer of this message (a miner actor ID).
   Sender ActorID
   // Vote is the payload that is signed by the signature.
@@ -243,11 +243,11 @@ type GMessage struct {
   Signature []byte
   // VRF ticket for CONVERGE messages (otherwise empty byte array).
   Ticket Ticket
-  // Justification for this message, or nil.
-  Justification *Justification
+  // Evidence that justifies this message, or nil.
+  Evidence *Evidence
 }
 
-type Justification struct {
+type Evidence struct {
   // Vote is the payload that is signed by the signature.
   Vote Payload
   // RLE+ serialization of the indexes in the base power table of the signers.
@@ -340,14 +340,14 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
 \*participants is implicitly used to calculate quorums
 
 1:    round ← 0;
-2:    terminated ← False;
+2:    decideSent ← False;
 3:    proposal ← inputChain;  \* holds what the participant locally believes should be a decision
 4:    timeout ← 2*Δ
 5:    ECCompatibleChains ← all prefixes of proposal, not lighter than baseChain
 6:    value ← proposal \* used to communicate the voted value to others (proposal or 丄)
 7:    evidence ← nil   \* used to communicate optional evidence for the voted value
 
-8:    while (not terminated)  {
+8:    while (not decideSent)  {
 9:      if (round = 0)
 10:       BEBroadcast <QUALITY, value, instance>; trigger (timeout)
 11:       collect a clean set M of valid QUALITY messages
@@ -388,8 +388,7 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
 38:     if (HasStrongQuorumValue(M) AND StrongQuorumValue(M) ≠ 丄)    \* decide
 39:       evidence ← Aggregate(M)
 39:       BEBroadcast <DECIDE, StrongQuorumValue(M), evidence>
-40:       terminated ← True
-          return (StrongQuorumValue(M), Aggregate(M))
+40:       decideSent ← True \* break loop, wait for other DECIDE messages
 41:     if (∃ m ∈ M: m.value ≠ 丄) \* m.value was possibly decided by others
 42:       proposal ← m.value; \* sway local proposal to possibly decided value
 43:       evidence ← m.evidence \* strong PREPARE quorum is inherited evidence
@@ -400,11 +399,12 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
 48:   }  \*end while
 
 49:   collect a clean set M of valid DECIDE messages
-      until (HasStrongQuorumValue(M)) \* Collect a strong quorum of decide outside the round loop
-50:   terminated ← True
-      return (StrongQuorumValue(M), Aggregate(M))
+      until (HasStrongQuorumValue(M)) \* collect a strong quorum of decide outside the round loop
+50:   return (StrongQuorumValue(M), Aggregate(M)) \* terminate the algorithm with a decision
 ```
-Note that the selection of the heaviest prefix in line 16 need not read the tipsets' weights, as only prefixes that extend each other can gather a strong quorum in QUALITY. In other words: if there are two tipsets $t_1, t_2$ that gather a strong quorum of QUALITY, then either the corresponding chain that has $t_1$ as head tipset is a prefix of the analogous chain that has $t_2$ as head, or viceversa (since the adversary controls < ⅓ of the QAP). As a result, selecting the heaviest prefix is as simple as selecting the highest blockheight, while ensuring all proposed prefixes that gather a strong quorum in QUALITY extend each other as a sanity check.
+Note that the selection of the heaviest prefix in line 16 does not need access to the tipsets' EC weights, as only prefixes that extend each other can gather a strong quorum in QUALITY. In other words: if there are two tipsets $t_1, t_2$ that gather a strong quorum of QUALITY, then either the corresponding chain that has $t_1$ as head tipset is a prefix of the analogous chain that has $t_2$ as head, or viceversa (since the adversary controls < ⅓ of the QAP). As a result, selecting the heaviest prefix is as simple as selecting the highest blockheight, while ensuring all proposed prefixes that gather a strong quorum in QUALITY extend each other as a sanity check.
+
+Implementations may optimise this algorithm to treat the reception of an aggregated signature over some (MsgType, Instance, Round, Value) as evidence of a message as if it had received the same tuple of values directly from each participant in the aggregate. This may be used to effectively skip a partially-complete step. In the particular case of a DECIDE message, which carries evidence of a strong quorum of COMMITs for the same round and value, a participant may immediately send its own DECIDE for the same value (copying the evidence) and exit the loop at line 8.
 
 Also, concurrently, we expect that the participant feeds to GossiPBFT chains that are incentive-compatible with EC. To this end, GossiPBFT has a separate invocation called $\texttt{ECUpdate}()$, which is called by an external process at a participant once EC delivers a $chain$ such that $inputChain$ is a prefix of $chain$ (i.e., EC at a participant delivers an extension of $inputChain$). This part is critical to ensuring the progress property in conjunction with lines 24-25.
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0086.md
```diff
@@ -179,7 +179,7 @@ The algorithm starts by initializing $finalizedTipsets[0]$ with the genesis tips
 3. Execute the instance $i$ of GossiPBFT consensus.
 4. Add the returned tipset from the consensus execution to the $finalizedTipsets$ list.
 
-To prevent useless instances of GossiPBFT deciding on the same $baseChain$ because of the lack of a new epoch that provides a new proposal, we make the $\texttt{ChainHead}()$ function blocking in that it does not return a proposal unless (i) the drand epoch value for the epoch immediately following the latest finalized tipset is received and (ii) the current chain head is different from the chain head of the finalized chain.
+To prevent useless instances of GossiPBFT deciding on the same $baseChain$ because of the lack of a new epoch that provides a new proposal, we make the $\texttt{ChainHead}()$ function blocking in that it does not return a proposal unless the current chain head is different from the chain head of the finalized chain.
```
anorth, line 182: This is still not a good description of the proposal, which is instead based on EC tipset timestamps.

FIPS/fip-0086.md
```diff
@@ -225,39 +225,59 @@ GossiPBFT was designed with the Filecoin network in mind and presents a set of f
...
-  // GossiPBFT ticket (only for CONVERGE)
-  Ticket []byte
 }
 
 type ECTipset struct {
   Epoch int
-  Tipset CID // Or TipsetKey (concat of block header CIDs)
-  Weight BigInt
-  PowerTable CID // CID of a PowerTable
+  Tipset []byte // CBOR-serialized tipset key, a canonical array of the tipset's block header CIDs.
```
anorth, line 274: @Stebalien can you confirm this is what we decided?

FIPS/fip-0086.md
```diff
@@ -434,15 +434,21 @@ If m is not properly signed:                | m must be properly signed.
...
-    return False                            | with baseChain.
+If not (IsPrefix(baseChain, m.value)        | A value must be bottom 
+    or m.value = 丄)                        | or start with baseChain.
+  return False                              
+If m.step = QUALITY AND                     | A QUALITY message
+  m.round != 0                              | must be for round 0
+    return false
+If m.step = DECIDE AND                      | A DECIDE message
+  m.round != 0                              | must specify round = 0
+    return false
```
anorth, line 445: This set of validation rules is still incomplete, the code specifies more. But this format for writing them is so inconvenient that I gave up attempting parity. @ranchalp or another author, could you please propose a change (perhaps another PR) that reformats this to be reasonable to edit and align with the code?

FIPS/fip-0086.md
```diff
@@ -469,10 +475,17 @@ OR (m = <COMMIT, value, round, evidence>                | valid COMMIT
         AND ∀ m' ∈ M: m'.value = m.value)               | for the same value, or
     OR (m.value = 丄))                                  | COMMIT is for 丄 with
                                                         | no evidence
+
+OR (m = <DECIDE, value, round, evidence>                | valid DECIDE
+  AND (∃ M: Power(M) > ⅔ AND m.evidence=Aggregate(M)    | evidence is a strong quorum
+    AND ∀ m' ∈ M: m'.step = COMMIT                      | of PREPARE messages
+      AND all the same round (!= m.round)               | from some (all the same) round
```
anorth, line 482: I don't know how to express "they're all the same as each other, but not the same as m" in this pseudo-syntax. @ranchalp halp?

FIPS/fip-0086.md
Stebalien: Yes.

Stebalien (approved):

FIPS/fip-0086.md
```diff
@@ -393,7 +388,8 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
 38:     if (HasStrongQuorumValue(M) AND StrongQuorumValue(M) ≠ 丄)    \* decide
 39:       evidence ← Aggregate(M)
 39:       BEBroadcast <DECIDE, StrongQuorumValue(M), evidence>
-40:       decideSent ← True
+40:       terminated ← True
```
ranchalp, line 391: Calling it `terminated` may make a reader think the entire protocol terminated, when in reality the node still has to gather a strong quorum of DECIDE messages.

FIPS/fip-0086.md
```diff
@@ -403,14 +399,10 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
 
-49:   collect a clean set M of DECIDE messages
+49:   collect a clean set M of valid DECIDE messages
       until (HasStrongQuorumValue(M)) \* Collect a strong quorum of decide outside the round loop
-50:   return (StrongQuorumValue(M), Aggregate(M))
-
-51:   upon reception of a valid <DECIDE, value, instance, evidence> message and not decideSent
-52:     decideSent ← True
-53:     BEBroadcast <DECIDE, value, instance, evidence>
-54:     go to line 49
```
ranchalp, line 413: It is true this is not part of the code currently. I think it is better to change the code to reflect this than it is to remove from here? Since DECIDE messages are justified we can send one as soon as we receive one.

FIPS/fip-0086.md
```diff
@@ -434,20 +426,26 @@ If m is not properly signed:                | m must be properly signed.
...
-If m.step ∈ {CONVERGE, COMMIT} AND          | CONVERGE, COMMIT
-  NOT ValidEvidence(m)                      | must contain valid evidence.
+If m.step ∈ {CONVERGE, COMMIT, DECIDE} AND  | CONVERGE, COMMIT, DECIDE
+  not ValidEvidence(m)                      | must contain valid evidence.
     return False
 return True
 ```
 
-The $\texttt{ValidEvidence}()$ predicate is defined below. Note that QUALITY, PREPARE and DECIDE messages do not need evidence. In fact, DECIDE does not need any protocol-specific validation, since a weak quorum of DECIDE with the same value is required to trigger any  execution of the protocol.
+The $\texttt{ValidEvidence}()$ predicate is defined below. Note that QUALITY and PREPARE messages do not need evidence.
```
ranchalp, line 448: This is going to change to fix the bugs Guy found: PREPARE for rounds>0 will have evidences. But that will come in a new PR.

FIPS/fip-0086.md
ranchalp: A switch statement is asked for at this point like in the code. Will do (in a new PR)

FIPS/fip-0086.md
```diff
@@ -469,10 +467,17 @@ OR (m = <COMMIT, value, round, evidence>                | valid COMMIT
         AND ∀ m' ∈ M: m'.value = m.value)               | for the same value, or
     OR (m.value = 丄))                                  | COMMIT is for 丄 with
                                                         | no evidence
+
+OR (m = <DECIDE, value, round, evidence>                | valid DECIDE
+  AND (∃ M: Power(M) > ⅔ AND m.evidence=Aggregate(M)    | evidence is a strong quorum
+    AND ∀ m' ∈ M: m'.step = COMMIT                      | of PREPARE messages
```
ranchalp, line 473:
```suggestion
    AND ∀ m' ∈ M: m'.step = COMMIT                      | of COMMIT messages
```

FIPS/fip-0086.md
ranchalp: They do need to be the same as m though?

I think what you mean is `AND ∀ m', m'' ∈ M: m'.value = m''.value`. But we do need to leave `AND ∀ m' ∈ M: m'.value = m.value)` I think.

ranchalp (commented):
Some comments

FIPS/fip-0086.md
```diff
@@ -225,39 +225,59 @@ GossiPBFT was designed with the Filecoin network in mind and presents a set of f
...
 ```
-type GossiPBFTMessage struct {
-  // ID of the sender/signer of this message (a miner actor)
+// A message in the Granite protocol.
+// The same message structure is used for all rounds and phases.
+// Note that the message is self-attesting so no separate envelope or signature is needed.
+// - The signature field fixes the included sender ID via the implied public key;
+// - The signature payload includes all fields a sender can freely choose;
+// - The ticket field is a signature of the same public key, so also self-attesting.
+type GMessage struct {
```
jsoares, line 237: `GMessage` isn't the most descriptive name, especially given it's also gossiped...

FIPS/fip-0086.md
jsoares: Yeah, I think it's fine to make the change for parity now and then follow up.

FIPS/fip-0086.md
```diff
@@ -469,10 +467,17 @@ OR (m = <COMMIT, value, round, evidence>                | valid COMMIT
...
+Note that during the validation of each CONVERGE, COMMIT and DECIDE message, two signatures need to be verified: (1) the signature of the message contents, produced by the sender of the message (first check above) and (2) the aggregate signature in $m.evidence$ ($\texttt{ValidEvidence}(m)$ above). The former can be verified using the sender's public key (stored in the power table). The latter, being an aggregated signature, requires aggregating public keys of all the participants whose signatures it combines. Notice that the evidence of each message can be signed by a different set of participants, which requires aggregating a linear number of public BLS keys (in system size) per evidence verification.
```
jsoares, line 480:
```suggestion
Note that during the validation of each CONVERGE, COMMIT, and DECIDE message, two signatures need to be verified: (1) the signature of the message contents, produced by the sender of the message (first check above) and (2) the aggregate signature in $m.evidence$ ($\texttt{ValidEvidence}(m)$ above). The former can be verified using the sender's public key (stored in the power table). The latter, being an aggregated signature, requires aggregating public keys of all the participants whose signatures it combines. Notice that the evidence of each message can be signed by a different set of participants, which requires aggregating a linear number of public BLS keys (in system size) per evidence verification.
```

consistency

FIPS/fip-0086.md
```diff
@@ -485,14 +490,14 @@ An instance of GossiPBFT requires a seed σ that must be common among all partic
...
+[Given prior tests performed on GossipSub](https://research.protocol.ai/publications/gossipsub-v1.1-evaluation-report/vyzovitis2020.pdf) (see also [here](https://gist.github.com/jsoares/9ce4c0ba6ebcfd2afa8f8993890e2d98)), we expect that almost all participants will reach sent messages within $Δ=3s$, with a huge majority receiving them even after $Δ=2s$. However, if several participants start the instance $Δ + ε$ after some other participants, termination is not guaranteed for the selected timeouts of $2*Δ$. Thus, we do not rely on an explicit synchrony bound for correctness. Instead, we increase the estimate of Δ locally within an instance as rounds progress without decision.
```
jsoares, line 493:
```suggestion
[Given prior tests performed on GossipSub](https://research.protocol.ai/publications/gossipsub-v1.1-evaluation-report/vyzovitis2020.pdf) (see also [here](https://gist.github.com/jsoares/9ce4c0ba6ebcfd2afa8f8993890e2d98)), we expect that sent messages will reach almost all participants within $Δ=3s$, with a majority receiving them even after $Δ=2s$. However, if several participants start the instance $Δ + ε$ after some other participants, termination is not guaranteed for the selected timeouts of $2*Δ$. Thus, we do not rely on an explicit synchrony bound for correctness. Instead, we increase the estimate of Δ locally within an instance as rounds progress without a decision.
```

Disjoint from your edits but might as well fix the language.

FIPS/fip-0086.md
```diff
@@ -225,39 +225,59 @@ GossiPBFT was designed with the Filecoin network in mind and presents a set of f
 
 #### Message format, signatures, and equivocation
 
-Messages include the following fields: $\langle Sender, Signature, MsgType, Value, Instance, [Round, Evidence, Ticket] \rangle$. As $Round$, $Evidence$, and $Ticket$ are fields that not all message types require, when not required by a message type, their default value is used (i.e. $0$, $\texttt{AggregatedEvidence}(0, [] ECTipset \lbrace\rbrace, 0, 0, [] byte \lbrace\rbrace, [] byte \lbrace\rbrace)$, and $[] byte \lbrace\rbrace$, respectively). We refer to a _field_ of message $m$, with $m.field$:
+Messages include the following fields: $\langle Sender, Signature, MsgType, Value, Instance, [Round, Justification, Ticket] \rangle$. As $Round$, Justification, and $Ticket$ are fields that not all message types require, when not required by a message type, their default value is used (i.e. $0$, $nil$, and $[] byte \lbrace\rbrace$, respectively). We refer to a _field_ of message $m$, with $m.field$:
```
jsoares, line 228:
```suggestion
Messages include the following fields: $\langle Sender, Signature, MsgType, Value, Instance, [Round, Justification, Ticket] \rangle$. As $Round$, $Justification$, and $Ticket$ are fields that not all message types require, when not required by a message type, their default value is used (i.e. $0$, $nil$, and $[] byte \lbrace\rbrace$, respectively). We refer to a _field_ of message $m$, with $m.field$:
```

FIPS/fip-0086.md
anorth: The `value`s need to be the same as m, but the `round`s (this line) will not be, because `m.round == 0` (from line 436).

anorth:
Thanks @ranchalp @jsoares, please take another look.

FIPS/fip-0086.md
ranchalp: Oh I see now, you meant rounds. 

This then:

```
OR (m = <DECIDE, value, round, evidence>                              | valid DECIDE
  AND (∃ M: Power(M) > ⅔ AND m.evidence=Aggregate(M)    | evidence is a strong quorum
    AND ∀ m' ∈ M: m'.step = COMMIT                                         | of PREPARE messages
      AND  ∀ m', m'' ∈ M: m'.round = m''.round                            | from some (all the same) round
```

ranchalp (approved):
LGTM! But perhaps fix the minor typo I commented on first.

FIPS/fip-0086.md
Stebalien: Actually, I misread. Do we not want this to be `[]Cid`? There really isn't any need to double-encode.

FIPS/fip-0086.md
jsoares: That sounds plausible, yes.

jsoares (approved):
Couple of outstanding issues but lgtm.

FIPS/fip-0086.md
anorth: I've changed to `[]CID` for now.

However, these are completely opaque to GPBFT. There's no good reason for the code to depend on a CID package. Unfortunately, because CID means an actual specific CBOR tag, specifying as `[][]byte` would result in different serialization. I'm unhappy about the broad code impact of these three characters in the spec, so we might come back to it.

PR #990, opened 2024-04-12
 [FIP-XXXX] Freeze stfil user’s transferred asset address

justinjuarez3518:
Discussion: https://github.com/filecoin-project/FIPs/discussions/989

Currently, more than 4.5 million FIL of Stfil user assets have been transferred to “unknown addresses” - https://filfox.info/en/address/f410falck3ysg7e2k4outtq2r24ytd66cuddydnoga6a, which most likely belong to the police.

According to "customary practice" these assets will soon be sold off in the secondary market, and all investors and community members should pay attention and take action.

I propose that FF and PL should make freezing this account and formally consider how to implement it and take action!

### DIFF
--- a/FIPS/fip-0091.md
+++ b/FIPS/fip-0091.md
@@ -1,0 +1,31 @@
+---
+fip: "0091"
+title: Freeze stfil user’s transferred asset address
+author: justin (@justinjuarez3518)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/989
+status: Draft
+type: Technical
+category: Core
+created: 2024-04-12
+---
+# FIP-0091: Freeze stfil user’s transferred asset address
+
+## Simple Summary
+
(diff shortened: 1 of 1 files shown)


jsoares (changes requested):
Early feedback:
- Please do not take a FIP number. The FIP number should remain TBD until an editor assigns one.
- This is arguably a recovery FIP rather than a technical one. Then again, this is new territory for us. 
- The text largely only contains motivation and lacks many of the requires sections. No specifics are provided on this would be operationalised: for instance, freezing an account can be done as a soft fork, by causing a majority of validators to reject txs from that address.
- Even then, I don't think the motivation is objective or accurate, or that it provides sufficient _information_ to the reader.

anorth:
Thank you for opening a discussion #989 but please don't duplicate the discussion premise here as a "draft FIP". I am closing this, but feel free to open a new draft after the discussion has resulted in some clear idea of exactly what protocol change is proposed and how to achieve it.

PR #985, opened 2024-04-08
FIP-0086: Changes from 1st Audit

ranchalp:
Draft for the moment. Track changes [here](https://docs.google.com/document/d/1zhjePPMP0Ycrm8aumHeOzHbHCbRYYXMhNH8W_uq_Og4/edit).

PR #998, opened 2024-04-26
FIP-0086 Include message-rebroadcast and skipping rounds

ranchalp:


FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
 
-Implementations may optimise this algorithm to treat the reception of an aggregated signature over some (MsgType, Instance, Round, Value) as evidence of a message as if it had received the same tuple of values directly from each participant in the aggregate. This may be used to effectively skip a partially-complete step. In the particular case of a DECIDE message, which carries evidence of a strong quorum of COMMITs for the same round and value, a participant immediately sends its own DECIDE for the same value (copying the evidence) and exits the loop at line 8.
+The reBroadcast function rebroadcasts all the messages in the current round if the participant cannot terminate the step that it is in by the time the timeout has expired. 
 
-Also, concurrently, we expect that the participant feeds to GossiPBFT chains that are incentive-compatible with EC. To this end, we restrict the set C of candidate values that the participant contributes to deciding to only values that either (i) pass the QUALITY step or (ii) may have been decided by other participants (hence the functin `mightHaveBeenDecided`).
+```
+71: func reBroadcast(msg):
```
masih, line 444:
```suggestion
71: reBroadcast(msg): 
```

FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+89:         				break;
+90: 			}
+91:			update(timeout_rebroadcast); \* increase and trigger again timeout_rebroadcast 
+92:			trigger(timeout_rebroadcast) 
+```
+
+The updateStep function simply updates the step and delivers locally all buffered messages for the new step and round:
+
+```
+93: func updateStep(round, new_step):
```
masih, line 471:
```suggestion
93: updateStep(round, new_step): 
```

FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+107:			round ← msg.round;
+108:   			timeout ← 2*Δ*2**msg.round
+109:			timeout_rebroadcast ← max(timeout+1, timeout_rebroadcast)
+110:			if msg'.evidence.step = PREPARE: \* either this or strong quorum of COMMITs for 丄
+111:	 			C ← C ∪ {msg'.value} 		\* add to candidate values if not there
+112:				proposal ← msg'.value; 		\* sway local proposal to possibly decided value
+113:			evidence ← msg'.evidence	\* strong PREPARE quorum is inherited evidence for the value (it exists because the value might have been decided)
+114:			go to line 22  \* start new round by jumping forward
+
+115: func shouldJump(round, step, timeout):
```
masih, line 499:
```suggestion
115: shouldJump(round, step, timeout): 
```

masih (commented):
Indentation in pseudocode looks uneven due to intermittent use if tab vs. space. A quick find/replace should get those sorted out. I recommend replacing them all with spaces.

I am worried that I might misinterpret the pseudocode because of the indentation. I'd be grateful if you can get the indentation fixed first before I leave too many silly comments 😊

ranchalp:
@masih the indenting has been fixed. Note indenting is not perfect (at least in edit mode) because the \bottom special character occupies more than a regular character (not fully monospace).

FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+99:   else if msg.step = DECIDE 
+100:    deliver(DECIDE) \*DECIDE immediately delivered within an instance
+101:  else if msg.step = step and msg.round = round
+102:    deliver(msg) \* can be delivered as it is for this round and step
+103:  else if msg.round > round
+104:    receiver_queue.Add(msg) 
+105:    receiver_queue.Trim() \* selectively drop messages if bound met. See text below.
+106:  if  (msg' ← shouldJump(round, step, timeout) s.t. msg' != nil) 	\* one of the rest of conditions to be able to jump
+107:    round ← msg.round;
+108:    timeout ← 2*Δ*2**msg.round
```
anorth, line 491: Should the second `2` here be the same value as the `1.3` in `delta * 1.3 ** r` in "Synchronization of Participants in the Current Instance"?

Please extract it as a parameter and use it in both places.

FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+39:     if (HasStrongQuorumValue AND StrongQuorumValue(M) = proposal) \* strong quorum of PREPAREs for local proposal
+40:       value ← proposal \* vote for deciding proposal (COMMIT)
+41:       evidence ← Aggregate(M) \* strong quorum of PREPAREs is evidence
+42:     else
+43:       value ← 丄 \* vote for not deciding in this round
+44:       evidence ← nil
+
+45:     reBroadcast <COMMIT, value, instance, round, evidence>; trigger(timeout)
+46:     collect a clean set M of valid COMMIT messages from this round and instance
+           OR (NOT mayHaveStrongQuorum(value, round, step, 0) for all value ≠ 丄)
```
anorth, line 403: Can we just say has strong quorum of bottom, instead of does not have for any other value? I believe that's a necessary equivalent condition, and this result is relied on at 55. Looping over all values and checking is inelegant if we have this assumption.

FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+           OR (timeout expires AND Power(M)>2/3)
+47:     if (HasStrongQuorumValue(M) AND StrongQuorumValue(M) ≠ 丄)    \* decide
+48:       evidence ← Aggregate(M)
+49:       reBroadcast <DECIDE, StrongQuorumValue(M), instance, evidence>  \* break loop, wait for other DECIDE messages
+50:     if (∃ m ∈ M: m.value ≠ 丄 s.t. mayHaveStrongQuorum(m.value, r, COMMIT, 1/3)) \* m.value was possibly decided by others
```
anorth, line 408: As I noted while implementing, this is simpler to follow if the branches of this `if` are flipped. Test explicitly for strong quorum of bottom. If that is not found, take the non-buttom value that must exist in some message, with the assertion that it might have been decided by another participant.

FIPS/fip-0086.md
anorth, line 403:
```suggestion
           OR (NOT mayHaveStrongQuorum(value, round, COMMIT, 0) for all value ≠ 丄)
```
?

FIPS/fip-0086.md
```diff
@@ -348,97 +348,193 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+32:              if (value ∈ C)
+33:                      proposal ←value \* we sway proposal if the value is incentive compatible (i.e., in C)
+34:                      prepareReadyToSend ← True    \* Exit loop 
+35:              else
+36:                      M = {m ∈ M | m.value != value AND m.evidence.value != evidence.value} \* Update M for next iteration } 
+
+37:     reBroadcast <PREPARE, value, instance, round, evidence>; trigger(timeout) \* evidence is nil in round=0
+38:     collect a clean set M of valid PREPARE messages from this round and instance
+           until (HasStrongQuorumValue(M) AND StrongQuorumValue(M) = proposal)
+           OR (NOT mayHaveStrongQuorum(proposal, round, step, 0))
```
anorth, line 392:
```suggestion
           OR (NOT mayHaveStrongQuorum(proposal, round, PREPARE, 0))
```
?

anorth (commented):
More substantive technical questions in https://github.com/filecoin-project/FIPs/discussions/809#discussioncomment-9334980 and https://github.com/filecoin-project/FIPs/discussions/809#discussioncomment-9335039

FIPS/fip-0086.md
```diff
@@ -475,37 +571,37 @@ The $\texttt{ValidEvidence}()$ predicate is defined below. Note that QUALITY mes
...
+OR (step=COMMIT                                           | valid COMMIT evidence 
+  AND (∃ M: Power(M)>⅔ AND evidence=Aggregate(M)          | is a strong quorum 
+      AND ∀ m’ ∈ M: m’.step = PREPARE                     | of PREPARE messages
+      AND ∀ m’ ∈ M: m’.round = round                      | from the same round 
+      AND ∀ m’ ∈ M: m’.value = value))                    | for the same value, or
+
+OR (step = DECIDE                                         | valid DECIDE evidence
+      AND (∃ M: Power(M)>⅔ AND evidence=Aggregate(M)      | is a strong quorum 
+        AND ∀ m’ ∈ M: m’.step = COMMIT                    | of COMMIT messages
+        AND ∀ m’ ∈ M: m’.round = round                    | from the same round
```
anorth, line 603: DECIDE round is always zero, so this never passes. This instead wants to say the COMMIT messages all have the same round as each other.

 (I thought we fixed this once already...)

FIPS/fip-0086.md
ranchalp: Yes, thanks, done.

FIPS/fip-0086.md
ranchalp: Yes, thanks, good catch. Done.

ranchalp: Wrote in a commit not to spam with a commit for this. Hence why not committing the suggestion.

FIPS/fip-0086.md
ranchalp: Done, thanks.

FIPS/fip-0086.md
ranchalp: Good catch, thanks. And yes, it sounds just like what we discussed at some point, but it is part of the modifications in this PR 🤔

ranchalp:
Thanks for the review @anorth , the changes have been added.

FIPS/fip-0086.md
ranchalp: Traced back the reintroduction of this typo to #985 . Went through the changes there and could not find any other typo being reintroduced, other than using `Power(M)>2/3` instead of `IsStrongQuorum` (and fixed that in this PR too).

FIPS/fip-0086.md
```diff
@@ -348,97 +348,179 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+7:    evidence ← nil   \* used to communicate optional evidence for the voted value
+8:    receiver_queue ←newQueue(size: 7*n) \* bounded queue, where n is the number of participants.
+9:    C ← {baseChain}
+10:   step ← nil
+11:   broadcast_msgs ← {} \* map with msgs sent by this participant per round and step
+
+
+12:   while (step != DECIDE)  {
+13:      if (round = 0)
+14:       BEBroadcast <QUALITY, value, instance>; trigger (timeout)
```
jsoares, line 362: The indentation step remains inconsistent, also on formatted preview. Can we standardise on 2 spaces per level or something alike?

image

jsoares (approved):
Editorial review. Looks good apart from some formatting issues. Did some checking against the Google doc and they seem consistent -- however, due to some changes in notation and structure, I may have missed something.

anorth:
I made my [comments on the new v3 version in the Google doc](https://docs.google.com/document/d/1RNLyuvyg5v8Va-HonOk-FeluFFjyZEjJitMze7F67-Q/edit) which was published at the same time, for faster/easier initial discussion.

ranchalp:
Went through you comments in the doc and applied most of the changes both here and there. @anorth

FIPS/fip-0086.md
```diff
@@ -348,97 +348,165 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
-54:   return (StrongQuorumValue(M), Aggregate(M)) \* terminate the algorithm with a decision
+16:       updateStep(round, QUALITY)
+17:       C ← C ∪ {prefix : IsPrefix(prefix,proposal) and HasStrongQuorum(prefix,M)}
+18:       proposal ← heaviest prefix ∈ C \* this becomes baseChain or sth heavier
+19:       value ← proposal
+
+20:     if (round > 0)     \* CONVERGE
+21:       ticket ← VRF(Randomness(baseChain) || instance || round)
+22:       value ← proposal \* set local proposal as value in CONVERGE message
+23:       BEBroadcast <CONVERGE, value, instance, round, evidence, ticket>; trigger(timeout)
```
anorth, line 373: Using BEBroadcast instead of ReBroadcast introduces an error. Even though a node can never get stuck in CONVERGE, the CONVERGE message does need to be re-sent if stuck later on. But without calling ReBroadcast, this message never enters the `broadcast_msgs` collection, so cannot be re-sent.

This is probably an artefact of the psuedo-code getting too concrete. Better to keep it abstract and just assume the node has a collection of all messages its sent (in the same way it has a collection of all messages it's received that it can evaluate predicates over). Then specify a predicate selecting message to re-send. Implementation can of course be more efficient.

FIPS/fip-0086.md
```diff
@@ -348,97 +348,165 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+24:       updateStep(round, CONVERGE)
+25:       collect a clean set M of  CONVERGE msgs from this round and instance
+             until timeout expires
+26:       value, justification ← GreatestTicketProposal(M)     \* leader election
+27:       if (evidence is a strong quorum of PREPAREs AND mayHaveStrongQuorum(value, r-1, COMMIT, 1/3))
+28:         C ← C ∪ {value}
+29:       if (value ∈ C)
+30:         proposal ← value \* we sway proposal if the value is incentive compatible (i.e., in C)
+31:         evindence ← justification \* otherwise keep existing proposal and evidence
```
anorth, line 382:
```suggestion
31:         evidence ← justification \* otherwise keep existing proposal and evidence
```

FIPS/fip-0086.md
```diff
@@ -348,97 +348,165 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+```
+91: updateStep(round, new_step): 
+92:   step ← new_step
+```
+
+The conditions for skipping rounds and buffering messages are shown below:
+
+```
+93: upon reception of a valid msg: \* assume validity implies same instance as currently running, different instances treated elsewhere (not scope of this doc)
+94:   if msg.round < round OR (msg.round > round AND msg.step = COMMIT AND msg.value = 丄)  \* drop message
```
anorth, line 477: This is a very concrete specification that is stricter than necessary. In fact, retaining some more messages will still be correct and safe, and also aid convergence in some cases.
- COMMITs for prior rounds should be delivered, since they may immediately cause a decision. Those messages are still live because they can change the outcome.
- COMMITs for bottom for at least some future rounds should be retained, or this will be quite brittle and rely on the rebroadcast mechanism much more than necessary. It can be implementation-defined, but the parameter for "how many rounds of COMMIT bottom can be kept" need not be the "zero" that's implied here.

FIPS/fip-0086.md
```diff
@@ -348,97 +348,165 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+81:           BEBroadcast(broadcast_msgs[msg.round][PREPARE]) \* no break, cascade to broadcast CONVERGE
+82:         default: \* QUALITY or CONVERGE, which will never reach this point, but send CONVERGE here because of cascading effect of switch cases
+83:           if msg.round > 0:
+84:             BEBroadcast(broadcast_msgs[msg.round][CONVERGE])
+85:             BEBroadcast(broadcast_msgs[msg.round-1][COMMIT])
+86:             BEBroadcast(broadcast_msgs[msg.round-1][PREPARE])
+87:             break;
+88:       }
+89:       update(timeout_rebroadcast); \* increase and trigger again timeout_rebroadcast 
+90:       trigger(timeout_rebroadcast)
```
anorth, line 463: This is a very concrete implementation spec that might be better lifted to be more abstract, similar to the predicates over implicit sets of messages in the main protocol.

E.g. "Collect all messages with sender=self and round=msg.round or (round =msg.round-1 and step in COMMIT|PREPARE)". With a special case for DECIDE, etc. Implementations can then implement that in a way that suits their data structures (most likely not with a `broadcast_msgs` structure as given)

FIPS/fip-0086.md
```diff
@@ -348,97 +348,165 @@ GossiPBFT(instance, inputChain, baseChain, participants) → decision, PoF:
...
+89:       update(timeout_rebroadcast); \* increase and trigger again timeout_rebroadcast 
+90:       trigger(timeout_rebroadcast) 
+```
+
+The updateStep function simply updates the step and delivers locally all buffered messages for the new step and round:
+
+```
+91: updateStep(round, new_step): 
+92:   step ← new_step
+```
```
anorth, line 471: The prose line is out of date. Since this function is now one line, I think better just to inline it where it's called.

anorth (commented):
Thanks, the major items I raised in discussion are now addressed. There are parts where this spec is much too concrete and doesn't appear to allow implementers much freedom to do it better. I think these would be better if made more abstract and directly describing the properties required, rather than how to achieve them.

There's lots of inconsistent indentation in the pseudocode. Can we stick to 2 spaces?

vukolic:
@anorth your last batch of comments is hopefully now fixed + I polished the indentation

FIPS/fip-0086.md
```diff
@@ -344,168 +344,236 @@ We illustrate the pseudocode for GossiPBFT below, consisting of 3 steps per roun
...
+83:            if msg.round > 0:
+84:              BEBroadcast(m ∈ own_msgs: m.round=msg.round AND m.type=CONVERGE)
+85:              BEBroadcast(m ∈ own_msgs: m.round=msg.round-1 AND m.type=COMMIT)
+86:              BEBroadcast(m ∈ own_msgs: m.round=msg.round-1 AND m.type=PREPARE)
+87:              break;
+88:        }
+89:        update(timeout_rebroadcast); \* increase and trigger again timeout_rebroadcast 
+90:        trigger(timeout_rebroadcast) 
+```
+The conditions for skipping rounds and buffering messages are shown below (below, retain_future_rounds is an implementation config paameter):
```
masih, line 468:
```suggestion
The conditions for skipping rounds and buffering messages are shown below (below, retain_future_rounds is an implementation config parameter):
```

Subjective suggestion: `max_lookahead_rounds` instead of `retain_future_rounds`, since the latter sounds like a binary flag.

anorth (approved):
Thanks, changes LGTM now

jennijuju:
@masih  / @anorth, can one of you resolve the conflict so we can merge this?

Kubuxu (approved):

PR #993, opened 2024-04-17
FIPXXXX: Introducing sealerID

irenegia:
This proposal introduced the sealerID. This is a new identifier created for parties that participate to the Filecoin network by  creating replicas, which can then transferred to SPs.

This fully enables the Sealing-as-a-Service scenario already introduced in FIP0090 (NI-PoRep).

Discussion: #890

PR #1000, opened 2024-04-29
Add ReplicaInfo to fip-0090

ZenGround0:
Proposed SealerActor FIP #993 is needed to fully unlock SaaS functionality after landing FIP 0090.

To avoid an annoying method deprecation we're predicting the needs of the SealerActor in the NI porep params.  This PR adds fields for specifying sealer actor id and sealer actor specified sector number to the NI porep method.  The protocol should enforce that these values are constrained until the sealer actor fip arrives.

### DIFF
--- a/FIPS/fip-0090.md
+++ b/FIPS/fip-0090.md
@@ -110,7 +110,7 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
   struct ReplicaInfo {
 	  // Sector number used to generate replica id 
 	  // Must be set to SectorID 
-      SectorNumber: SectorNumber,
+	  SectorNumber: SectorNumber,
 	  // Sector number used as unique id in actor state
 	  SectorID: SectorNumber,
 	  // Must be set to ProviderID 
--- a/FIPS/fip-0090.md
+++ b/FIPS/fip-0090.md
@@ -97,12 +97,24 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
 - Introduce a new method `ProveCommitSectorsNI` (method 36), which performs a non-interactive proof for CC sectors, without a preceding PreCommitSector message (ie, this method rejects sectors that were already pre-committed. Such sectors must be activated with one of the existing ProveCommit methods). The method can be used with seal proofs or aggregate proof (but only one option can be chosen per call); the return type is the same  as `ProveCommitAggregate`.
   
   ```go
-  // Note no UnsealedCID because it must be "zero" data.
+    // Note no UnsealedCID because it must be "zero" data.
   struct SectorNIActivationInfo {
-      SectorNumber: SectorNumber,
+	  ReplicaInfo: ReplicaInfo
       SealedCID: Cid, // CommR
       SealRandEpoch: ChainEpoch,
       Expiration: ChainEpoch,
+  }
+  
+  // Organize information used in replica id
+  // Included to avoid breaking change for future SealerActor FIP 
+  struct ReplicaInfo {
+	  // Sector number used to generate replica id 
+	  // Must be set to SectorID 
+      SectorNumber: SectorNumber,
+	  // Sector number used as unique id in actor state
+	  SectorID: SectorNumber,
+	  // Must be set to ProviderID 
+	  SealerID: ActorID
   }
 
   struct ProveCommitSectorsNIParams {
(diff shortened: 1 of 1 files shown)


FIPS/fip-0090.md
```diff
@@ -97,13 +97,25 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
...
+  // Organize information used in replica id
+  // Included to avoid breaking change for future SealerActor FIP 
+  struct ReplicaInfo {
+	  // Sector number used to generate replica id 
+	  // Must be set to SectorID 
+	  SectorNumber: SectorNumber,
+	  // Sector number used as unique id in actor state
+	  SectorID: SectorNumber,
+	  // Must be set to ProviderID 
+	  SealerID: ActorID
```
jennijuju, line 117: Can this be null? https://filecoinproject.slack.com/archives/C04USV0L7PF/p1714484414153329?thread_ts=1714409050.743159&channel=C04USV0L7PF&message_ts=1714484414.153329

FIPS/fip-0090.md
```diff
@@ -97,13 +97,25 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
...
       SealRandEpoch: ChainEpoch,
       Expiration: ChainEpoch,
   }
+  
+  // Organize information used in replica id
+  // Included to avoid breaking change for future SealerActor FIP 
+  struct ReplicaInfo {
+	  // Sector number used to generate replica id 
+	  // Must be set to SectorID 
+	  SectorNumber: SectorNumber,
```
mtrisic, line 113: This is "sector number" that's unique to a sealer, and will change after onboarding a sector?  Can we pick another name to avoid confusion, something like `SectorOrdinal` or  `SealingNumber`?

FIPS/fip-0090.md
ZenGround0: This is an integer so it cannot be set to cbor null.  We could force set it to -1 in case we are using provider ID as sealer ID if we prefer that.  However I think that is less clean than directly using the provider ID which is the SealerID in this case.

FIPS/fip-0090.md
ZenGround0: I like Sealing Number

FIPS/fip-0090.md
jennijuju: Didn’t realize it’s an int - sgtm

jennijuju (approved):

FIPS/fip-0090.md
```diff
@@ -99,11 +99,23 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
...
+  
+  // Organize information used in replica id
+  // Included to avoid breaking change for future SealerActor FIP 
+  struct ReplicaInfo {
+	  // Sector number used to generate replica id 
+	  // Must be set to SectorID 
+	  SealingNumber: SectorNumber,
+	  // Sector number used as unique id in actor state
+	  SectorID: SectorNumber,
+	  // Must be set to ProviderID
```
anorth, line 116: ProviderID isn't defined in this document. Do you mean it must be the ID of the actor receiving this method call (for now)?

FIPS/fip-0090.md
```diff
@@ -99,11 +99,23 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
...
   }
+  
+  // Organize information used in replica id
+  // Included to avoid breaking change for future SealerActor FIP 
+  struct ReplicaInfo {
+	  // Sector number used to generate replica id 
+	  // Must be set to SectorID 
+	  SealingNumber: SectorNumber,
+	  // Sector number used as unique id in actor state
+	  SectorID: SectorNumber,
```
anorth, line 115: This is called `SectorNumber` almost everywhere in existing code, so keep that name here.

FIPS/fip-0090.md
```diff
@@ -99,11 +99,23 @@ Note that, same as with interactive PoRep, each sector has a `SealRandEpoch` tha
...
+	  // Must be set to ProviderID 
+	  SealerID: ActorID
```
anorth, line 117: I suggest ordering these fields as:
- SealerID
- SealingNumber
- SectorNumber

So that the two items defining the replica are adjacent, and the one that's just about the subsequent miner accounting comes last. That latter arguable doesn't belong in this structure at all. And some other fields like CommR arguably do! Those questions could go away if the ReplicaInfo struct was inlined.

anorth (commented):
Conceptual 👍 , but nits

FIPS/fip-0090.md
ZenGround0: I kept everything inline to avoid these problems

anorth (approved):

jsoares (approved):

PR #1008, opened 2024-05-13
FIP-0092: First draft of caller-specified proving deadline for NI-PoRep

anorth:
SPs cannot currently control the allocation of sectors to Window PoST deadlines.
Sectors are automatically spread out over the full 24-hour proving period.

This proposal allows SPs to select a deadline when onboarding sectors with the new NI-PoRep method.

Discussion: https://github.com/filecoin-project/FIPs/discussions/1007

### DIFF
--- a/FIPS/fip-xxxx-proving-deadline.md
+++ b/FIPS/fip-xxxx-proving-deadline.md
@@ -1,0 +1,114 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
+title: Caller-specified proving deadline for NI-PoRep
+author: "Alex North (@anorth)"
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/1007
+status: Draft
+type: Technical (Core)
+created: 2024-05-14
+requires (*optional): FIP-0090
+---
+
+# FIP-XXXX: Caller-specified proving deadline for NI-PoRep
+
+## Simple Summary
+Allows storage providers to specify the Window PoST proving deadline for sectors onboarded with NI-PoRep.
+
+## Abstract
+Storage providers (SPs) cannot currently control the allocation of sectors to Window PoST deadlines.
+Sectors are automatically spread out over the full 24-hour proving period.
+
+This proposal allows SPs to select a deadline when onboarding sectors with the new NI-PoRep method.
+
+## Change Motivation
+SPs would benefit from greater operational flexibility than being required to spread Window PoST workload over 24 hours.
+[FIP-0070](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0070.md)
+attempted to allow moving of sectors between deadlines after onboarding, 
+but the scheduling system has proven too complex to easily accommodate the change.
+
+Allowing SPs to specify the proving deadline when onboarding is not quite as powerful, but much simpler.
+Such a change is most easily made by introducing a new onboarding method.
+Attaching it to the new method introduced by FIP-0090 is an expedient way to get this functionality.
+
+## Specification
+The parameters for the miner actor method `ProveCommitSectorsNI` (number 36) are changed:
+
+```go
+  struct ProveCommitSectorsNIParams {
+     // Information about sealing of each sector.
+     Sectors: []SectorNIActivationInfo
+     // Proof type for each seal (must be an NI-PoRep variant)
+     SealProofType: RegisteredSealProof,
+     // Proofs for each sector, parallel to activation manifests.
+     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+     SectorProofs: [][]byte,
+     // Aggregate proof for all sectors.
+     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+     AggregateProof: []byte,
+     // Proof type for aggregation, if aggregated
+     AggregateProofType: Option<RegisteredAggregateProof>
+     // Whether to abort if any sector activation fails.
+     RequireActivationSuccess: bool,
+     
+     // NEW! The Window PoST deadline index at which to schedule the new sectors.
+     ProvingDeadline uint64 
+   }
+```
+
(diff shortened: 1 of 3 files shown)


FIPS/fip-xxxx-proving-deadline.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Abstract
+SPs cannot currently control the allocation of sectors to Window PoST deadlines.
+Sectors are automatically spread out over the full 24-hour proving period.
+
+This proposal allows SPs to select a deadline when onboarding sectors with the new NI-PoRep method.
+
+## Change Motivation
+SPs would benefit from greater operational flexibility than being required to spread Window PoST workload over 24 hours.
+FIP-0070 attempted to allow moving of sectors between deadlines after onboarding,
```
jennijuju, line 25:
```suggestion
[FIP-0070](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0070.md) attempted to allow moving of sectors between deadlines after onboarding, 
```

FIPS/fip-xxxx-proving-deadline.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Security Considerations
+The existing automatic scheduling was motivated by a caution about overloading the network at particular times of day 
+with too many Window PoSTs due at the same time.
+However, since the introduction of the FVM and a robust gas model,
+the network itself is not at much risk from this anymore.
+We might reasonably expect SPs to make reasonable scheduling decisions to optimize their own cost and risk profile,
+given knowledge and expectations about the scheduling of other SPs.
+
+A Window PoST for every sector must still be submitted every 24 hours.
```
jennijuju, line 89: just want to confirm that the existing dispute windowpost mechanism would still be sufficient with this feature?

jennijuju (approved):

jennijuju:
editor review ✅  - fip draft is simple and clear.

FIPS/fip-xxxx-proving-deadline.md
```diff
@@ -0,0 +1,106 @@
...
+with too many Window PoSTs due at the same time.
+However, since the introduction of the FVM and a robust gas model,
+the network itself is not at much risk from this anymore.
+We might reasonably expect SPs to make reasonable scheduling decisions to optimize their own cost and risk profile,
+given knowledge and expectations about the scheduling of other SPs.
+
+A Window PoST for every sector must still be submitted every 24 hours. 
+
+## Incentive Considerations
+This proposal provides flexibility without any significant changes to incentives.
```
momack2, line 92:
```suggestion
This proposal provides flexibility to Storage Providers, without any significant changes to incentives.
```

momack2 (approved):

FIPS/fip-xxxx-proving-deadline.md
anorth: Good question, I added a paragraph on it. In summary I think it has no impact unless we have so many Window PoST that we're approaching a large fraction of network total gas bandwidth. We're far from that today.

FIPS/fip-xxxx-proving-deadline.md
```diff
@@ -0,0 +1,114 @@
...
+     // Aggregate proof for all sectors.
+     // Exactly one of sector_proofs or aggregate_proof must be non-empty.
+     AggregateProof: []byte,
+     // Proof type for aggregation, if aggregated
+     AggregateProofType: Option<RegisteredAggregateProof>
+     // Whether to abort if any sector activation fails.
+     RequireActivationSuccess: bool,
+     
+     // NEW! The Window PoST deadline index at which to schedule the new sectors.
+     ProvingDeadline uint64
```
NemanjaLu92, line 54: I have two questions about this parameter:
 - Should it perhaps be an optional parameter? Do we still want to support automatic deadline placement in case SP doesn't care?
 - Any particular reason why this parameter is not `u8` as it's range is only [0, 47]?

anorth: > Do we still want to support automatic deadline placement in case SP doesn't care?

No, I don't think we should. I mentioned this briefly in the discussion, but I think we should deprecate the existing behaviour, and eventually be able to remove the code for it. The motivation is to remove complexity from the built-in actors. This complexity can live off-chain instead.

> Any particular reason why this parameter is not u8?

It makes no difference at the API since these are CBOR-encoded. We use Rust or Go syntax here in lieu of a good IPLD-CBOR schema language. So ... consistency with all other representations of a deadline index.

FIPS/fip-xxxx-proving-deadline.md
```diff
@@ -0,0 +1,114 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
```
jennijuju, line 2: 0092

FIPS/fip-xxxx-proving-deadline.md
rvagg: well, we do have an IPLD Schema language that covers all of this, it's just never been used in this repo for whatever reason but it's perfectly good for this kind of thing and even tells you about the tuple encoding.

```
type ProveCommitSectorsNIParams struct {
  # Information about sealing of each sector.
  Sectors [SectorNIActivationInfo]
  # Proof type for each seal (must be an NI-PoRep variant)
  SealProofType RegisteredSealProof
  # Proofs for each sector, parallel to activation manifests.
  # Exactly one of sector_proofs or aggregate_proof must be non-empty.
  SectorProofs [Bytes]
  # Aggregate proof for all sectors.
  # Exactly one of sector_proofs or aggregate_proof must be non-empty.
  AggregateProof Bytes
  # Proof type for aggregation, if aggregated
  AggregateProofType nullable RegisteredAggregateProof
  # Whether to abort if any sector activation fails.
  RequireActivationSuccess Bool

  # NEW! The Window PoST deadline index at which to schedule the new sectors.
  ProvingDeadline Int
} representation tuple
```

PR #1011, opened 2024-05-16
Consistent formatting of FIPs table in README.

anorth:
Consistent use of spaces in the table.

I did not go all the way to fully formatting the table by aligning columns, since the rows would all then be very long, and need a subsequent reformat each time a longer row is added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,94 +38,94 @@ This improvement protocol helps achieve that objective for all members of the Fi
 
 ## FIPs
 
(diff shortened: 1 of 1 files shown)


anorth:
As a follow-up, I would also suggest rationalising the authors column to be only GitHub `@name`s and/or email addresses. The full names can remain in the FIP documents.

jennijuju (approved):

raulk (approved):
This was a pet peeve for me, so thanks for fixing.

PR #1012, opened 2024-05-17
FIP-0090: fix resources dir name, fix links to resources

rvagg:
* Rename `resources/fip-xxx-ni-porep/` -> `resources/fip-0090/`
* Fix links to resources from `FIPS/fip-0090.md`

Note that the links didn't work anyway, they were pointing to `resources/fip-xxx-niporep/` which isn't what it's called now.


jennijuju (approved):
Thank you!

anorth (approved):

PR #1022, opened 2024-06-18
FIP-0086 updates from implementation

anorth:
Two notable changes feeding back from the implementation, plus some smaller improvements:
- There is not (yet) any mechanism to slow F3 if EC is producing only empty tipsets
- The `mayHaveStrongQuorum` function is changed to treat boundary conditions "correctly"
- Described F3 committee lookback
- Specified filters to power table eligibility (TODO: confirm with @Stebalien, especially around ticket math)
- Specified power scaling/truncation
- Specified values for step numbers

Tightened up some language and formatting:
- Removed some unnecessary phrases.
- Removed lingering tab characters.
- Unified commenting style in pseudocode to use `//`.
- Replaced some academic argument-style language (especially use of"we") with declarative statements.

A potentially clarifying change we could make would be a change in terminology from power table & power (EC concepts) to committee and weight (F3 concepts). These terms are currently a bit ambiguous since the concepts are not 1:1. I refrained for now.

### DIFF
--- a/FIPS/fip-0086.md
+++ b/FIPS/fip-0086.md
@@ -54,9 +54,9 @@ The participants in F3 are the storage providers (SPs) of the Filecoin network. 
 In short, each participant $p$ in the Filecoin network (i.e., a storage provider with power) runs F3 in a loop and feeds its input from EC. F3 returns a finalized prefix chain to EC, which updates the canonical chain to extend this F3-finalized prefix. More precisely, in each loop iteration $i$ (corresponding to the i-th instance of F3):
 
 - **EC/F3 interface:** Participant $p$ feeds its current canonical chain $canonical$ and its previously F3-finalized chain, called the $baseChain$, to F3. The $baseChain$ (with some lookback parameter) fixes the power table used by F3 and randomness seed used to configure the F3 instance, while $p$'s current canonical chain is the chain $p$ proposes to be finalized.
-- **F3 consensus:** F3 establishes network-wide consensus on finalized tipsets and produces a _Proof of Finality (PoF)_ for a tipset finalized in instance $i$, $t_i$. Every _PoF_ output by F3 is signed by ≥ ⅔ of the total QAP, i.e., a super-majority of the power table. This vouches that honest participants with more than ⅓ QAP consider tipset $t_i$ final (assuming up to ⅓ QAP may be dishonest).
+- **F3 consensus:** F3 establishes network-wide consensus on finalized tipsets and produces a _Certificate of Finality_ for a tipset finalized in instance $i$, $t_i$. Every certificate output by F3 is signed by ≥ ⅔ of the total QAP, i.e., a super-majority of the power table. This vouches that honest participants with more than ⅓ QAP consider tipset $t_i$ final (assuming up to ⅓ QAP may be dishonest).
 - **EC:** Participant $p$ updates its local EC chain and commits not to reorganize the finalized tipset $t_i$, i.e., the tipset must stay in $p$'s EC canonical chain for good. Apart from this change, EC continues operating as it currently does. In particular, EC still does a 900-epoch lookback for its power table (which can therefore differ from the one used by F3) and continues operating “normally” if F3 assumptions are violated and F3 halts, with the combined EC/F3 protocol favoring availability over consistency (in CAP theorem parlance).
-- **F3 synchronization:** Participants disseminate information about finalized tipset $t_i$ and its proof $PoF_i$ to all other participants, light clients, and smart contracts. The main goal of F3 synchronization is to allow an external party (or a lagging F3 participant) to fetch a sequence of messages that prove the finality of some recent final tipset. The sequence demonstrates a chain of eligible participants deciding on a final tipset and, thus, the eligible power table for the next round. Verifying the finality of a tipset from genesis does not require access to the EC chain.
+- **F3 synchronization:** Participants disseminate information about each finalized tipset and associated certificate to all other participants, light clients, and smart contracts. The main goal of F3 synchronization is to allow an external party (or a lagging F3 participant) to fetch a sequence of messages that prove the finality of some recent final tipset. The sequence demonstrates a chain of eligible participants deciding on a final tipset and, thus, the eligible power table for the next round. Verifying the finality of a tipset from genesis does not require access to the EC chain.
 
 ![](../resources/fip-0086/flow.png)
 
@@ -91,6 +91,25 @@ In practice, if GossipSub is used for _BEBroadcast_, $\Delta$ is an assumed uppe
 
 For termination, a classical partially synchronous model is assumed, with an unknown bound on communication delays among non-Byzantine participants.
 
+### Power Table and Consensus Committee
+
+The EC chain state maintains a table of the quality-adjusted power of each SP, as well as their signing key. Each tipset $t$ in the chain commits to a particular state that results from executing all messages in the chain from genesis to $t$. The committee of participants for an instance of GossiPBFT is determined by the chain state resulting from the tipset finalized by a previous instance: the input parameter $committee$ of instance $i$ is defined by the tipset output by instance $i-PowerTableLookback$. If $i$ is less than $PowerTableLookback$, $committee$ is drawn from the chain state committed by the genesis.
+
+Given an EC chain state at some epoch, the corresponding F3 committee comprises all miner actors with:
+- at least 10TiB of quality-adjusted power;
+- zero fee debt; and
+- no active consensus fault.
+
+The F3 consensus power of each eligible F3 participant is the corresponding miner actor's EC quality-adjusted power, scaled such that the total power of all participants fits in a 16-bit unsigned integer.
+
+```
+ParticipantPower ← truncate(0xffff * MinerQAP / TotalQAP)
+```
+
+The total F3 consensus weight is the sum of all eligible participants' consensus weight (note this differs from EC leader election). An F3 participant's signing key is the associated miner actor's worker key (always a BLS key).
+
+The parameter $PowerTableLookback$ takes the value 10.
+
 ### Properties of F3
 
 F3 is not identical to classical consensus, however similar to it. In a nutshell, the reasons why EC and Filecoin should not rely on classical consensus properties (implemented by existing off-the-shelf BFT consensus protocol) for fast finality are twofold: EC-compatibility and resilience to Filecoin power leakage attacks in case of a strong adversary. 
@@ -99,7 +118,7 @@ With this in mind, the F3 component interface and properties are given below.
 
 > **Interface and properties of F3 at participant p:**
 >
-> `F3.invoke (int i, chain canonical, chain baseChain)` **returns** ("finalizes") `(int i, chain h, proof_of_finality PoF)`
+> `F3.invoke (int i, chain canonical, chain baseChain)` **returns** ("finalizes") `(int i, chain h, proof_of_finality certificate)`
 >
 > **Properties:**
 >
@@ -107,7 +126,7 @@ With this in mind, the F3 component interface and properties are given below.
 >
 > **Validity.** If an honest participant finalizes $(i,h,\ast)$, then $h$ is a prefix of the canonical input chain of some honest participant $p'$ in instance $i$.
 >
-> **Proof of Finality.** If an honest participant finalizes $(i,h,PoF)$, then $PoF$ is signed by ⅔ QAP majority corresponding to the $\texttt{PowerTable}(baseChain)'$ input in instance $i$ of some honest participant $p'$.
+> **Proof of Finality.** If an honest participant finalizes $(i,h,certificate)$, then certificate is signed by ⅔ QAP majority corresponding to the committee input to instance $i$ of some honest participant $p'$.
 >
 > **Progress.** If the system is $\Delta$-synchronous, i.e.,
 > * All honest participants can communicate within a known time bound $\Delta$, and
@@ -129,73 +148,64 @@ The GossiPBFT consensus protocol is detailed [below](#GossiPBFT-Consensus). It i
 Invocation of a consensus instance from a participant is denoted:
 
 ```
-GossiPBFT(i, proposal, baseChain, participants) → decision, PoF
+GossiPBFT(i, proposal, baseChain, committee) → decision, certificate
 ```
 
-A participant that invokes the $\texttt{GossiPBFT()}$ function starts a consensus instance identified by the sequence number $i$ and with an input value $proposal$. The $baseChain$ parameter represents the default value (agreement on not finalizing any new tipset). The $participants$ parameter is composed of the SPs that participate in this instance along with their respective individual weights (i.e., quality-adjusted power). The next section explains how the $participants$ are obtained from the power table. The invocation eventually returns a $decision$ value and a proof of finality $PoF$ that proves $decision$ is indeed the output of this consensus instance. In any invocation from F3, $decision$ is a finalized chain prefix in the form of a tipset. $PoF$ can be verified by using the $\texttt{Verify}()$ function:
+A participant that invokes the $\texttt{GossiPBFT()}$ function starts a consensus instance identified by the sequence number $i$ and with an input value $proposal$. The $baseChain$ parameter represents the default value (agreement on not finalizing any new tipset). The $committee$ parameter is composed of the SPs that participate in this instance along with their respective individual weights (i.e., quality-adjusted power) and signing keys. The invocation eventually returns a $decision$ value and a certificate that proves $decision$ is indeed the output of this consensus instance. In any invocation from F3, $decision$ is a finalized chain prefix in the form of a tipset. Certificates can be verified by using the $\texttt{Verify}()$ function:
 
 ```
-Verify(PoF, decision, participants) → boolean
+Verify(certificate, decision, committee) → boolean
 ```
 
-This function uses the $PoF$ to verify that $decision$ was the output of some instance of consensus executed among the $participants$.
+This function uses the certificate to verify that $decision$ was the output of some instance of consensus executed among the $committee$.
 
-
-### Power Table and Consensus Participants
-
-The _power table_ in the existing power actor maintains, among other things, the quality-adjusted power of each SP. Each tipset $t$ in the chain commits to a particular power table denoted $\texttt{PowerTable}(t)$ that results from executing all messages in the chain from genesis to $t$. The participants of an instance of GossiPBFT are determined by the power table resulting from the tipset finalized by a previous instance. More rigorously, the input parameter $participants$ of an instance $i$ is obtained from $\texttt{PowerTable}(decision_{i-PowerTableLookback})$, where $decision_{i-PowerTableLookback}$ is the tipset output by instance $i-PowerTableLookback$. If $i$ is less than $PowerTableLookback$, $participants$ is the power table committed by genesis tipset.
-
-The parameter $PowerTableLookback$ takes the value 10.
 
 ### F3 Pseudocode
 
 The F3 algorithm is specified by pseudo-code:
 
 ```
-// finalizedTipsets is a list of records with tipset and PoF fields representing
-// all finalized tipsets with their respective proof of finality.
+// A list of all finalized tipsets with their respective certificates.
+// One entry per consensus instance.
+// Initialized with a genesis which needs no proof.
+finalizedTipsets[0] ← {tipset: genesis, certificate: nil}
 
-// It is initialized with the genesis tipset by definition, which needs no proof.
-finalizedTipsets[0].tipset ← genesis
-finalizedTipsets[0].PoF ← nil
-
+// The first instance number after genesis is 1.
 i ← 1
 while True:
-  wait until EC.CurrentEpoch() >= finalizedTipsets[i-1].Head().epoch + 2: // do not start until 2 epochs ahead
-  proposal ← EC.HeaviestUnfinalizedChain() // get heaviest chain from EC
+   // Delay start until EC is 2 epochs ahead of last finalised chain.
+  wait until EC.CurrentEpoch() >= finalizedTipsets[i-1].Head().epoch + 2
+  // Fetch tipsets from locally-observed EC chain from (inclusive) the last finalised tipset.
+  proposal ← EC.HeaviestUnfinalizedChain()
   if proposal.Head().epoch = EC.CurrentEpoch()
-	proposal = proposal[-1] // remove head from current epoch, as pre-agreement is unlikely
-  if i > PowerTableLookback:
-  	participants ← PowerTable(finalizedTipsets[i-PowerTableLookback].tipset)
-  else:
-	participants ← PowerTable(finalizedTipsets[0].tipset) // use genesis table the first 10 instances.
-  finalizedTipset, PoF ← GossiPBFT(i, proposal, finalizedTipsets[i-1], participants)
-  finalizedTipsets[i].tipset ← finalizedTipset
-  finalizedTipsets[i].PoF ← PoF
+    // Remove current epoch tipset, as pre-agreement is unlikely
+    proposal ← proposal[:-1]
+  // Get the power table from previously-finalised tipset from lookback instances ago.
+  // Use genesis power for initial instances.
+  lookback ← min(i, PowerTableLookback)
+  committee ← MakeCommittee(finalizedTipsets[i-lookback].tipset)
+  // Execute GPBFT instance i.
+  finalizedTipset, certificate ← GossiPBFT(i, proposal, finalizedTipsets[i-1], committee)
+  // Record decision and prepare for next instance.
+  finalizedTipsets[i] ← {tipset: finalizedTipset, certificate: certificate}
   i ← i + 1
 ```
 
-Each participant keeps track of the finalized tipsets and respective proofs of finality in a data structure named $finalizedTipsets$. It contains one $finalizedTipsets[i]$ entry per consensus instance where $finalizedTipsets[i].tipset$ and $finalizedTipsets[i].PoF$ denote, respectively, the tipset and PoF output by consensus instance $i$. A participant considers a tipset finalized if it is included in $finalizedTipsets$ or is an ancestor of some tipset included in $finalizedTipsets$.
-
-The algorithm starts by initializing $finalizedTipsets[0]$ with the genesis tipset. This tipset is pre-defined as finalized and does not require a PoF. Then, in each iteration $i$ of the loop, it takes the following steps:
-
-1. Wait until the current epoch is at least two epochs from that of the latest finalized head.
-2. Obtain the set of participants of instance $i$ from the power table determined by the finalized tipset $PowerTableLookback$ instances ago.
-3. Call the $\texttt{HeaviestUnfinalizedChain}()$ function that returns the tipsets of the locally observed chain constructed by EC from finalizedTipset[i-1].Head (including the head of the finalized chain) and sets the $proposal$ variable to the returned value. Note that the proposal is pruned so that its head is at most a tipset from the previous epoch, not the current epoch.
-4. Execute the instance $i$ of GossiPBFT consensus.
-5. Add the returned tipset from the consensus execution to the $finalizedTipsets$ list.
+A participant considers a tipset _final_ if it is included in, or is an ancestor of, any finalized tipset.
 
 Note that if EC produces empty tipsets, EC epochs progress regardless. F3 will then repeatedly propose and finalize the same base chain until a non-empty tipset is produced. 
 
 ### Changes to EC: Fork Choice Rule
 
-EC is modified to accommodate the finalization of tipsets by F3. **This is the only change to EC this FIP introduces.**
+The current EC fork-choice rule selects, from all the known tipsets, the tipset backed by the most weight.
+This fork choice rule is modified to accommodate the finalization of tipsets by F3. **This is the only change to EC this FIP introduces.**
 
-The current EC fork-choice rule selects, from all the known tipsets, the tipset backed by the most weight. As the F3 component finalizes tipsets, the fork-choice rule must ensure that the heaviest finalized chain is always a prefix of the heaviest chain, preventing reorganizations of the already finalized chain.
+The fork choice rule is modified in the presence of a finalized prefix:
+* the fork choice rule selects the heaviest chain out of all chains with a prefix that **includes all tipsets finalized by F3**.
 
-The fork choice rule is adjusted in the presence of a finalized prefix: the fork choice rule selects the heaviest chain out of all chians with a prefix that **matches exactly all tipsets finalized by F3**, in that a tipset $t'$ that is a superset of finalized tipset $t$ in the same epoch is not heavier than $t$ itself, despite it being backed by more EC power.
+A competing tipset $t'$ that is a superset of finalized tipset $t$ at the same epoch is not heavier than $t$ itself, despite it being backed by more EC power.
 
-This redefinition of the fork choice rule is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of participants in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by the EC proposer.
+This redefinition of the fork choice rule is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of the committee in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by it's proposer.
 
 The updated rule is illustrated in the following figure, where blocks in blue are finalized blocks, and all blocks are assumed to be proposed by a proposer holding only one EC ticket (the weight of a chain is the number of blocks):
 
@@ -213,9 +223,9 @@ Integrating F3 into Filecoin follows the usual path for Filecoin upgrades. One e
 
 ### GossiPBFT Consensus
 
-This section provides the specification of GossiPBFT, the consensus protocol that is iteratively instantiated by the F3 loop and returns a decision (in the form of an F3-finalized chain) and a PoF per instance.
+This section provides the specification of GossiPBFT, the consensus protocol that is iteratively instantiated by the F3 loop and returns a decision (in the form of an F3-finalized chain) and a certificate per instance.
 
-GossiPBFT is a Byzantine fault-tolerant consensus protocol that is resilient optimal, i.e., it tolerates up to less than ⅓ QAP being controlled by a Byzantine adversary. Each instance of the protocol has a known set of participants. Each participant inputs a proposal value, and the protocol outputs one of the input values as the final decision value. Unlike a longest-chain protocol, the output of GossiPBFT is permanent.
+GossiPBFT is a Byzantine fault-tolerant consensus protocol that is resilient optimal, i.e., it tolerates up to less than ⅓ QAP being controlled by a Byzantine adversary. The committee for each instance of the protocol is known. Each participant inputs a proposal value, and the protocol outputs one of the input values as the final decision value. Unlike a longest-chain protocol, the output of GossiPBFT is permanent.
 
 GossiPBFT was designed with the Filecoin network in mind and presents a set of features that make it desirable in that context:
 
@@ -230,24 +240,24 @@ GossiPBFT was designed with the Filecoin network in mind and presents a set of f
 
 #### Message format, signatures, and equivocation
 
-Messages include the following fields: $\langle Sender, Signature, MsgType, Value, Instance, [Round, Evidence, Ticket] \rangle$. As $Round$, $Evidence$, and $Ticket$ are fields that not all message types require, when not required by a message type, their default value is used (i.e. $0$, $nil$, and $[] byte \lbrace\rbrace$, respectively). We refer to a _field_ of message $m$, with $m.field$:
+The schema for messages exchanged in the GPBFT protocol is defined below.
+The same message structure is used for all rounds and phases.
+
+Note that the message is self-attesting so no separate envelope or signature is needed:
+- The signature field fixes the included sender ID via the implied public key;
+- The signature payload includes all fields a sender can freely choose;
+- The ticket field is a signature of the same public key, so also self-attesting.
 
 ```go
-// A message in the GossiPBFT protocol.
-// The same message structure is used for all rounds and phases.
-// Note that the message is self-attesting so no separate envelope or signature is needed.
-// - The signature field fixes the included sender ID via the implied public key;
-// - The signature payload includes all fields a sender can freely choose;
-// - The ticket field is a signature of the same public key, so also self-attesting.
 type GossiPBFTMessage struct {
   // ID of the sender/signer of this message (a miner actor ID).
   Sender ActorID
-  // Vote is the payload that is signed by the signature.
+  // The payload that is signed by the signature.
   Vote Payload
   // Signature by the sender's public key over serialized vote payload.
   Signature []byte
   // VRF ticket for CONVERGE messages (otherwise empty byte array).
-  Ticket Ticket
+  Ticket []byte
   // Evidence that justifies this message, or nil.
   Evidence *Evidence
 }
@@ -267,7 +277,7 @@ type SupplementalData struct {
   // Merkle-tree of instance-specific commitments. Currently empty (0) but this will
   // eventually include things like snark-friendly power-table commitments.
   Commitments [32]byte
-  // The DagCBOR-blake2b256 CID of the power table used to validate the next
+  // The DagCBOR-blake2b256 CID of the power table / committee to be used to validate the *next*
   // instance, taking lookback into account.
   PowerTable CID
 }
@@ -288,13 +298,13 @@ type Payload struct {
 }
 
 type ECTipset struct {
-  Epoch       uint64   // The Filecoin epoch
-  TipsetKey   []byte   // A canonical concatination of the block-header CIDs in a tipset.
-  PowerTable  CID      // A blake2b-256 CID of the cbor-encoded power-table.
+  Epoch uint64         // The Filecoin epoch
+  TipsetKey []byte     // A canonical concatination of the block-header CIDs in a tipset.
+  PowerTable CID       // A blake2b-256 CID of the cbor-encoded power-table.
   Commitments [32]byte // Root of a Merkle tree containing additional commitments.
 }
 
-// Table of nodes with power > 0 and their public keys.
+// Table of nodes eligible for the committee, with their weights public keys.
 // This is expected to be calculated from the EC chain state and provided to GossiPBFT.
 // Ordered by (power descending, participant ascending).
 type PowerTable = []PowerTableEntry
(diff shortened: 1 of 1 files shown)


FIPS/fip-0086.md
```diff
@@ -178,25 +180,24 @@ Each participant keeps track of the finalized tipsets and respective proofs of f
...
+2. Obtain the set of participants of instance $i$ from the power table determined by the finalized tipset $PowerTableLookback$ instances ago.
+3. Call the $\texttt{HeaviestUnfinalizedChain}()$ function that returns the tipsets of the locally observed chain constructed by EC from finalizedTipset[i-1].Head (including the head of the finalized chain) and sets the $proposal$ variable to the returned value. Note that the proposal is pruned so that its head is at most a tipset from the previous epoch, not the current epoch.
 4. Execute the instance $i$ of GossiPBFT consensus.
 5. Add the returned tipset from the consensus execution to the $finalizedTipsets$ list.
 
-To prevent useless instances of GossiPBFT deciding on the same $baseChain$ because of the lack of a new epoch that provides a new proposal, we make the $\texttt{ChainHead}()$ function blocking in that it does not return a proposal unless the current chain head is different from the chain head of the finalized chain.
-
+Note that if EC produces empty tipsets, EC epochs progress regardless. F3 will then repeatedly propose and finalize the same base chain until a non-empty tipset is produced.
```
anorth, line 188: This is a semantic change, there is no such blocking implemented. It seems difficult to add because participants would observe the first interesting chain at different times and so become unsynchronised. We need them all to start at the ~same time. 

There could be a more complicated function of how many times the same tipset has been finalised and its epoch to compute an increasing delay.

Stebalien: cc @Kubuxu you had a proposal, IIRC?

FIPS/fip-0086.md
```diff
@@ -354,170 +377,181 @@ Then, we build a [merkle tree](#merkle-tree-format) from these encoded `ECTipset
...
 
 ```
 61:  mayHaveStrongQuorum(value, round, step, advPower): 
-62:    M  ← { m | m.step = step AND m.round = round AND m is valid} /* clean set of messages
+62:    M  ← { m | m.step = step AND m.round = round AND m is valid} // Clean set of messages
 63:    M' ← { m | m ∈ M AND m.value = value } 
-64:    if Power(M') + (1-Power(M)) < ⅔ - advPower  : \* value cannot have been decided 
-65:      return False
-66:    return True
+64:    return Power(M') + (1-Power(M)) + advPower > ⅔ :
```
anorth, line 480: @ranchalp I believe this expression is more clear. However there's an edge condition change here in the change of operator from `<` to `>`. I think the code I've proposed here does what it says: if all the power adds up only to _equal to 2/3_ then there is no strong quorum possible. But I recall this edge case came up in earlier discussions and I was unable to recall if it was critical.

I can change the operator to `>=`, but then I think we'd want a clear description of why the `=` case was considered a possible strong quorum.

FIPS/fip-0086.md
Kubuxu: The proposal I had was something akin to back-off when we finalize base. The time gets larger (up to a cap) each time we observe the base is finalized. It would depend on the previous instances.

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
+
+Given an EC chain state at some epoch, the corresponding F3 committee comprises all miner actors with:
+- at least 10TiB of quality-adjusted power;
+- zero fee debt; and
+- no active consensus fault.
+
+The F3 consensus power of each eligible F3 participant is the corresponding miner actor's EC quality-adjusted power, scaled such that the total power of all participants fits in a 16-bit unsigned integer.
+
+```
+ParticipantPower ← truncate(0xffff * MinerQAP / TotalQAP)
```
anorth, line 106: @Stebalien I inferred this from code. But currently this is an over-simplification because the code also preserves unscaled values, and uses them in ticket computations. 

Can we make it an accurate simplification by removing unscaled power from the code?

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
-In practice, if GossipSub is used for _BEBroadcast_, $\Delta$ is effectively the assumed upper bound on GossipSub latency (e.g., a few seconds).
+### Power Table and Consensus Committee
 
-For termination, we assume a classical partially synchronous model with an unknown bound on communication delays among non-Byzantine participants.
+The EC chain state maintains a table of the quality-adjusted power of each SP, as well as their signing key. Each tipset $t$ in the chain commits to a particular state that results from executing all messages in the chain from genesis to $t$. The committee of participants for an instance of GossiPBFT is determined by the chain state resulting from the tipset finalized by a previous instance: the input parameter $committee$ of instance $i$ is defined by the tipset output by instance $i-PowerTableLookback$. If $i$ is less than $PowerTableLookback$, $committee$ is drawn from the chain state committed by the genesis.
+
+Given an EC chain state at some epoch, the corresponding F3 committee comprises all miner actors with:
+- at least 10TiB of quality-adjusted power;
+- zero fee debt; and
+- no active consensus fault.
```
anorth, line 101: @Kubuxu could you please check the EC leader election and confirm there are no other relevant checks? I'm aware that in EC these are applied against both lookback state and current state: we can only do one.

FIPS/fip-0086.md
Stebalien: Probably? Although I'm not sure what the impact of a ticket clash is (I assume rather small).

FIPS/fip-0086.md
Kubuxu: In the code, I'm using the same rules that apply to the EC election but without the Finality lookback.

FIPS/fip-0086.md
```diff
@@ -116,87 +138,76 @@ With this in mind, the F3 component interface and properties are given below.
...
 i ← 1
 while True:
-  wait until EC.CurrentEpoch() >= finalizedTipsets[i-1].Head().epoch + 2: // do not start until 2 epochs ahead
-  if i > 10:
-  	participants ← PowerTable(finalizedTipsets[i-10].tipset)
-  else:
-	participants ← PowerTable(finalizedTipsets[0].tipset) // use finalizedTipsets[0]'s table the first 10 instances.
-  proposal ← EC.HeaviestUnfinalizedChain() // get heaviest chain from EC
+   // Delay start until EC is 2 epochs ahead of last finalised chain.
+  wait until EC.CurrentEpoch() >= finalizedTipsets[i-1].Head().epoch + 2
```
Kubuxu, line 177: We can't do that exactly. We want strict timing on instance start, so it should depend only on the previously finalized chain. This depends on each participant's local view of EC chain.

Now I noticed that it is already part of the spec.

FIPS/fip-0086.md
anorth: That sounds great. Can you please clarify that I have captured those rules here? Are there any more?

FIPS/fip-0086.md
anorth: `EC.CurrentEpoch()` is a bit ambiguous here, but I interpreted it to mean "when the clock says the epoch has started", with no reference to the chain. Does that match your understanding?

FIPS/fip-0086.md
masih: This proposal is now implemented/merged.

FIPS/fip-0086.md
ranchalp: You are right, it should be `>2/3`and not `>=2/3`. However I noticed the FIP defines a strong quorum as `Power(p) >= \lceil Power(P)*2/3 \rceil` so that should be addressed and changed to `>2/3`. This is only ever a problem if the total power is divisible by 3 and there is a partition in 3 equal parts, one of them Byzantine.

ranchalp: Actually, since the adversary always hold strictly less than a third of the total power, the definition of strong quorum is correct. And therefore that case I suggest you maintain your modification but replacing `>` by `>=`.

FIPS/fip-0086.md
anorth: Please either push a commit to or make a PR on this PR

FIPS/fip-0086.md
Stebalien: Fixed.

FIPS/fip-0086.md
anorth: Wow 🤯 . The code has defined strong quorum as `> 2/3` for ever. How did we miss this?

Looking back, I see that strong quorum was defined using `>` in the specification docs at least up until [v2](https://docs.google.com/document/d/1Orm58L63SEESM6tDXK9QmWLnnMfF5i-Drj8MB8D3kXw/edit) (and v3 omitted the definition). So this appears to be a concrete change between those earlier documents and the FIP.

If we take a definition of the adversary as strictly <1/3, then it makes sense. But just to confirm: you are proposing that the FIP as it stands is correct with strong quorum as `>= 2/3` everywhere and we should change the code to match.

anorth: Actually, @Stebalien changed the code definition from `>` to `>=` in https://github.com/filecoin-project/go-f3/pull/118 when implementing scaled power, but I didn't review that change, or hence notice.

ranchalp: Yes, what I meant is that you wrote `return Power(M') + (1-Power(M)) + advPower > ⅔` and I suggest you write `return Power(M') + (1-Power(M)) + advPower >= ⅔` instead. And yes, if the adversarial power is advPower we always can wait for P - advPower units of weight (where P is the total power). Since the adversarial power advPower<1/3 always (because of the LSP82 and DSN88 impossibility results), then P-advPower >= 2P/3 in all cases, so we can wait for >= 2P/3 and this is safe.

In practice for Filecoin this will not matter as the difference is an edge case that will never happen anyways.  (If we wait for >2P/3 then if the adversary is exactly advPower = \lceil P/3 \rceil - 1 then it is possible that we may be waiting for an adversarial node that will never reply, i.e. suppose P = 7 so advPower=2 and we are waiting for >2P/3 or equivalently > 5, then we are waiting for at least 6 votes that we would never get. But this does not matter when P is large enough and even more so when dealing with units of weights and not just number of participants.

anorth:
Thanks for the contributions @masih @Stebalien.

This PR is ready for editorial and detailed technical review, and then merging.

FIPS/fip-0086.md
```diff
@@ -116,87 +138,96 @@ With this in mind, the F3 component interface and properties are given below.
...
+// A list of all finalized tipsets with their respective certificates.
+// One entry per consensus instance.
+// Initialized with a genesis which needs no proof.
+finalizedTipsets[0] ← {tipset: genesis, certificate: nil}
 
-// It is initialized with the genesis tipset by definition, which needs no proof.
-finalizedTipsets[0].tipset ← genesis
-finalizedTipsets[0].PoF ← nil
+// The number of epochs by which instance start is delayed relative to 
+// epoch of last finalised chain head.
```
masih, line 174: The FIP has a mix of British-Engish and American-english wording. We probably want to do a scan of all such cases and pick one (probably American-english)? Example: the text contains both "finalise" and "finalize" in number of places.

masih (approved):

FIPS/fip-0086.md
Kubuxu: Yeah, that makes sense, I read the spec as `loop on this check until ChainHead.Epoch() >= ...` but I'm fine moving on from this.

FIPS/fip-0086.md
Kubuxu: There is a caveat about not having enough miners (if there are fewer miners than X, the minimum power no longer applies). Other than that, it is the 3 checks you mentioned.

Kubuxu (approved):
LGTM

FIPS/fip-0086.md
momack2: Should we add that line here that the number of miners must be X to reach consensus?

FIPS/fip-0086.md
```diff
@@ -780,26 +842,29 @@ And probably more. Importantly, the network can easily commit to additional valu
 
 #### Wire Format v. Signing Format
 
-For both the `Payload` and the `ECTIpset`s themselves, we encode them one way for signing and another when sending them over the network.
+Both the `Payload` and the `ECTIpset`s themselves are encoded them one way for signing and another when sending them over the network.
```
momack2, line 845:
```suggestion
Both the `Payload` and the `ECTIpset`s themselves are encoded one way for signing and another when sending them over the network.
```

FIPS/fip-0086.md
```diff
@@ -780,26 +842,29 @@ And probably more. Importantly, the network can easily commit to additional valu
...
+| **32-63**   | x | x | x | x | x | x | x | x | x | x | x  | x  | x  | x  | x  | x   | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  | x  |
+| **64-95**   | y | y | y | y | y | y | y | y | y | y | y  | y  | y  | y  | y  | y   | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  | y  |
+| **96-127**  | z | z | z | z | z | z | z | z | z | z | z  | z  | z  | z  | z  | z   | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  | z  |
+| **126-133** | z | z | z | z | z | z |   |   |   |   |    |    |    |    |    |     |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |
 
-Where `.` is a zero byte, `xxx` is the instance (64bit BigEndian integer) and `yyy` is the root hash of the tipset merkle tree.
+Where `.` is a zero byte, `iii` is the instance (64bit BigEndian integer), `xxx` is the root hash of the per-instance commitmenets tree (from the supplemental data), `yyy` is the root hash of the tipset merkle tree, and `zzz` is the power table CID used to validate the next instance.
```
momack2, line 865: what is "commitmenets" supposed to be? commitments?

momack2 (approved):

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
-- **F3 consensus:** F3 establishes network-wide consensus on finalized tipsets and produces a _Proof of Finality (PoF)_ for a tipset finalized in instance $i$, $t_i$. Every _PoF_ output by F3 is signed by ≥ ⅔ of the total QAP, i.e., a super-majority of the power table vouching that honest participants with more than ⅓ QAP consider tipset $t_i$ final.
+- **EC/F3 interface:** Participant $p$ feeds its current canonical chain $canonical$ and its previously F3-finalized chain, called the $baseChain$, to F3. The $baseChain$ (with some lookback parameter) fixes the power table used by F3 and randomness seed used to configure the F3 instance, while $p$'s current canonical chain is the chain $p$ proposes to be finalized.
```
jennijuju, line 56: maybe worth specifying what is the loockback parameter?

jennijuju: (later on line 111 it specifies the value to be 10 it seems like

FIPS/fip-0086.md
```diff
@@ -932,22 +991,19 @@ Because of changes to the EC fork choice rule, this FIP requires a network upgra
...
+* **Committees.** This FIP proposes to have all eligible EC participants run all instances of GossiPBFT. While this ensures optimal resilience against a Byzantine adversary, it can render the system unusable if the number of participants grows too large. This introduces an attack vector: if the scalability limit is 100,000 participants, a participant holding as little as 3% of the current QAP can perform a Sybil attack to render the system unusable, with the minimum QAP required per identity. As a result, the implementation should favor the messages of the more powerful participants if the number of participants grows too large. Given that favoring more powerful participants discriminates against the rest, affecting decentralization, amending F3 to use committees in the event of the number of participants exceeding the practical limit might be the topic of a future FIP, as well as the analysis of optimized message aggregation in the presence of self-selected committees.
```
jennijuju, line 1000: what's the practical scalability limit today, i.e., if participant # is approaching X, a FIP is necessary to harden the protocol further?

jennijuju:
editorial ✅

jennijuju (approved):

jsoares:
I'm hoping to still read through today, assuming it's not merged before.

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
 
-F3 uses GossipSub (like Filecoin EC) to disseminate protocol messages, implementing a broadcast channel among participants. We assume and model GossipSub to satisfy the following properties of the _best-effort broadcast primitive_ (denoted _BEBroadcast_ hereafter):
+F3 uses GossipSub (like Filecoin EC) to disseminate protocol messages, implementing a broadcast channel among participants. GossipSub is assumed to satisfy the following properties of the _best-effort broadcast primitive_ (denoted _BEBroadcast_ hereafter):
 
-* If $p$ and $p'$ are honest, then every message broadcast by $p$ is eventually delivered by $p'$.
-* No message with sender $p$ is delivered unless it was previously broadcast by $p$.
+* _No creation_: No message with sender $p$ is delivered unless it was previously broadcast by $p$.
+* _Validity_: If $p$ and $p'$ are honest, then every message broadcast by $p$ is eventually delivered by $p'$.
 
+Note that academic nomenclature may be misleading, and "every message broadcast is eventually delivered" does not mean the broadcast is reliable. Messages may be lost if the sender, receiver, or link fail, but will be delivered if they don't.
```
jsoares, line 82: I don't think this is a matter of misleading language. This is merely a stated assumption, which (as is often the case) doesn't 100% match reality (e.g. if honest $p'$ dies and doesn't recover, it obviously won't receive it).

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
+
+The _no duplication_ property of a traditional best-effort broadcast primitive is not needed; indeed F3 explicitly rebroadcasts messages to overcome loss. 
 
 ### Partially Synchronous Model and Δ-Synchrony
 
-We say the system is $\Delta$-_synchronous_ if the combined computation and communication delay between any two honest participants is smaller than $\Delta$. We assume that, under $\Delta$-synchrony, any message broadcast by an honest participant is delivered by every honest participant within a time bound of $\Delta$.
+A system is $\Delta$-_synchronous_ if the combined computation and communication delay between any two honest participants is smaller than $\Delta$. Under $\Delta$-synchrony, any message broadcast by an honest participant is delivered by every honest participant within a time bound of $\Delta$.
+
+In practice, if GossipSub is used for _BEBroadcast_, $\Delta$ is an assumed upper bound on GossipSub latency (e.g., a few seconds).
```
jsoares, line 90:
```suggestion
In practice, if GossipSub is used for _BEBroadcast_, $\Delta$ is an assumed upper bound on GossipSub latency.
```

Doesn't add much information.

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
+```
+
+The total F3 consensus weight is the sum of all eligible participants' consensus weight (note this differs from EC leader election). An F3 participant's signing key is the associated miner actor's worker key (always a BLS key).
+
+The parameter $PowerTableLookback$ takes the value 10.
 
 ### Properties of F3
 
-F3 is not identical to classical consensus, however similar to it. In a nutshell, the reasons why EC and Filecoin should not rely on classical consensus properties (implemented by existing off-the-shelf BFT consensus protocol) for fast finality are twofold: EC compatibility and resilience to Filecoin power leakage attacks in case of a strong adversary. 
+F3 is not identical to classical consensus, however similar to it. In a nutshell, the reasons why EC and Filecoin should not rely on classical consensus properties (implemented by existing off-the-shelf BFT consensus protocol) for fast finality are twofold: EC-compatibility and resilience to Filecoin power leakage attacks in case of a strong adversary.
```
jsoares, line 115:
```suggestion
F3 is not identical to classical consensus, however similar to it. In a nutshell, the reasons why EC and Filecoin should not rely on classical consensus properties (implemented by existing off-the-shelf BFT consensus protocol) for fast finality are twofold: EC compatibility and resilience to Filecoin power leakage attacks in case of a strong adversary. 
```

It's actually defined as EC compatibility (no hyphen) above.

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
 
-F3 is not identical to classical consensus, however similar to it. In a nutshell, the reasons why EC and Filecoin should not rely on classical consensus properties (implemented by existing off-the-shelf BFT consensus protocol) for fast finality are twofold: EC compatibility and resilience to Filecoin power leakage attacks in case of a strong adversary. 
+F3 is not identical to classical consensus, however similar to it. In a nutshell, the reasons why EC and Filecoin should not rely on classical consensus properties (implemented by existing off-the-shelf BFT consensus protocol) for fast finality are twofold: EC-compatibility and resilience to Filecoin power leakage attacks in case of a strong adversary. 
 
 With this in mind, the F3 component interface and properties are given below.
 
 > **Interface and properties of F3 at participant p:**
 >
-> `F3.invoke (int i, chain canonical, chain baseChain)` **returns** (we say finalizes) `(int i, chain h, proof_of_finality PoF)`
+> `F3.invoke (int i, chain canonical, chain baseChain)` **returns** ("finalizes") `(int i, chain h, proof_of_finality certificate)`
```
jsoares, line 121: Following the changes above, a "proof of finality" is never introduced or defined until line 129 below, and even then it's a bit shaky. Certificate is also not defined as a value of type proof_of_finality. Could it be more intuitive to just use `finality_certificate fc` or something alike?

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
 >
 > **Properties:**
 >
 > **Agreement.** If two honest participants finalize $(i,h,\ast)$ and $(i,h',\ast)$, then $h = h'$.
 >
 > **Validity.** If an honest participant finalizes $(i,h,\ast)$, then $h$ is a prefix of the canonical input chain of some honest participant $p'$ in instance $i$.
 >
-> **Proof of Finality.** If an honest participant finalizes $(i,h,PoF)$, then $PoF$ is signed by ⅔ QAP majority corresponding to the $\texttt{PowerTable}(baseChain)'$ input in instance $i$ of some honest participant $p'$.
+> **Proof of Finality.** If an honest participant finalizes $(i,h,certificate)$, then certificate is signed by ⅔ QAP majority corresponding to the committee input to instance $i$ of some honest participant $p'$.
```
jsoares, line 129:
```suggestion
> **Proof of Finality.** If an honest participant finalizes $(i,h,certificate)$, then $certificate$ is signed by ⅔ QAP majority corresponding to the committee input to instance $i$ of some honest participant $p'$.
```

But see previous comment.

FIPS/fip-0086.md
```diff
@@ -116,87 +138,96 @@ With this in mind, the F3 component interface and properties are given below.
 >
 > **EC compatibility.** An honest participant never contributes to gathering ⅔ QAP (super-majority) of votes for a chain $c$ which its local EC instance did not already deliver unless it already observes evidence of such a super-majority for chain $c$.
 
-We later show the specific estimate chosen for $\Delta$ to optimize termination (See [here](#Synchronization-of-Participants-in-the-Current-Instance)).
+The $\Delta$ parameter is specified (See [below](#Synchronization-of-Participants-in-the-Current-Instance)).
```
jsoares, line 141:
```suggestion
The $\Delta$ parameter is specified [below](#Synchronization-of-Participants-in-the-Current-Instance).
```

This was extremely dangling. It's still a bit dangling. Do we need this line?

FIPS/fip-0086.md
jsoares: I'd rather go the other way, but that's probably the sensible choice. But let's definitely make it consistent.

FIPS/fip-0086.md
```diff
@@ -116,87 +138,96 @@ With this in mind, the F3 component interface and properties are given below.
...
+    startBackoffEpochs ← startBackoffEpochs + nextStartBackoffIncrement(startRetryAttempts)
+    startRetryAttempts ← startRetryAttempts + 1
+    continue // Skip to the next iteration.
+  else
+    // Reset backoff parameters. 
+    startBackoffEpochs ← 0
+    startRetryAttempts ← 0
+  // Get the power table from previously-finalised tipset from lookback instances ago.
+  // Use genesis power for initial instances.
+  lookback ← min(i, PowerTableLookback)
```
jsoares, line 203: Is `PowerTableLookback` supposed to be EC's 900? It's undefined here.

FIPS/fip-0086.md
```diff
@@ -116,87 +138,96 @@ With this in mind, the F3 component interface and properties are given below.
...
+A competing tipset $t'$ that is a superset of finalized tipset $t$ at the same epoch is not heavier than $t$ itself, despite it being backed by more EC power.
 
-This redefinition of the fork choice rule is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of participants in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by the EC proposer.
+This redefinition of the fork choice rule is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of the committee in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by it's proposer.
```
jsoares, line 228:
```suggestion
This redefinition of the fork choice rule is consistent with the abstract notion of the heaviest chain being backed by the most power because a finalized tipset has been backed by a super-majority of the committee in GossiPBFT. In contrast, any non-finalized block in the same epoch is only backed in that epoch by its proposer.
```

FIPS/fip-0086.md
```diff
@@ -212,41 +243,41 @@ Integrating F3 into Filecoin follows the usual path for Filecoin upgrades. One e
...
+This section provides the specification of GossiPBFT, the consensus protocol that is iteratively instantiated by the F3 loop and returns a decision (in the form of an F3-finalized chain) and a certificate per instance.
 
-GossiPBFT is a Byzantine fault-tolerant consensus protocol that is resilient optimal, i.e., it tolerates up to less than ⅓ QAP being controlled by a Byzantine adversary. Each instance of the protocol has a known set of participants. Each participant inputs a proposal value, and the protocol outputs one of the input values as the final decision value. We emphasize _final_ because, unlike a longest-chain protocol, the output of GossiPBFT is permanent.
+GossiPBFT is a Byzantine fault-tolerant consensus protocol that is resilient optimal, i.e., it tolerates up to less than ⅓ QAP being controlled by a Byzantine adversary. The committee for each instance of the protocol is known. Each participant inputs a proposal value, and the protocol outputs one of the input values as the final decision value. Unlike a longest-chain protocol, the output of GossiPBFT is permanent.
```
jsoares, line 248:
```suggestion
GossiPBFT is a Byzantine fault-tolerant consensus protocol that is resilient-optimal, i.e., it tolerates up to less than ⅓ QAP being controlled by a Byzantine adversary. The committee for each instance of the protocol is known. Each participant inputs a proposal value, and the protocol outputs one of the input values as the final decision value. Unlike a longest-chain protocol, the output of GossiPBFT is permanent.
```

This should likely be _resilient-optimal_ (i.e. with hyphen) or _optimally resilient_. It doesn't make much sense as is, but perhaps it's accepted in the field.

jsoares (commented):
part 1/3

FIPS/fip-0086.md
```diff
@@ -354,170 +408,181 @@ Then, we build a [merkle tree](#merkle-tree-format) from these encoded `ECTipset
...
+62:    M  ← { m | m.step = step AND m.round = round AND m is valid} // Clean set of messages
 63:    M' ← { m | m ∈ M AND m.value = value } 
-64:    if Power(M') + (1-Power(M)) < ⅔ - advPower  : \* value cannot have been decided 
-65:      return False
-66:    return True
+64:    return Power(M') + (1-Power(M)) + advPower >= ⅔ 
 ```
 
-The reBroadcast function rebroadcasts all the messages in the current round if the participant cannot terminate the step that it is in by the time the timeout has expired. 
+The reBroadcast function broadcasts and remembers a message. It then sets a timeout to rebroadcast a subset of the messages in the current and previous round if the participant cannot terminate the step that it is in.
```
jsoares, line 514:
```suggestion
The reBroadcast function broadcasts and remembers a message. It then sets a timeout to rebroadcast a subset of the messages in the current and previous round if the participant cannot terminate the step in which it is.
```

FIPS/fip-0086.md
```diff
@@ -354,170 +408,181 @@ Then, we build a [merkle tree](#merkle-tree-format) from these encoded `ECTipset
...
-51:      evidence ← m.evidence \* strong PREPARE quorum is inherited evidence
+45:        reBroadcast <DECIDE, StrongQuorumValue(M), evidence>  
+46:      else  
+           // No participant decided in this round, carry forward existing proposal
+47:        evidence ← Aggregate(M)  // Strong quorum of COMMITs for 丄 is evidence
+48:    if (∃ m ∈ M: m.value ≠ 丄)  // Value was possibly decided by others
+49:      C ← C ∪ {m.value}  // Add to candidate values if not there
+50:      proposal ← m.value  // Sway local proposal to possibly decided value
+51:      evidence ← m.evidence  // Strong PREPARE quorum is inherited evidence
 52:    round ← round + 1;
```
jsoares, line 486:
```suggestion
52:    round ← round + 1
```
There are random terminal semicolons sprinkled throughout the pseudocode. Should all be removed for consistency.

FIPS/fip-0086.md
```diff
@@ -354,170 +408,181 @@ Then, we build a [merkle tree](#merkle-tree-format) from these encoded `ECTipset
...
-            AND msg'.round > round          \* round must be greater
+116:    if (∃ msg' st. msg'.step = CONVERGE // There must be a CONVERGE for the new round
+            AND msg'.round > round          // Round must be greater
             AND ∃ M s.t. M is clean and valid and contains a weak quorum of PREPAREs for msg'.round)
 117:      return msg'
 118:    return nil
 ```
 
-Note that the selection of the heaviest prefix in line 18 does not need access to the tipsets' EC weights, as only prefixes that extend each other can gather a strong quorum in QUALITY. In other words: if there are two tipsets $t_1, t_2$ that gather a strong quorum of QUALITY, then either the corresponding chain that has $t_1$ as head tipset is a prefix of the analogous chain that has $t_2$ as head, or viceversa (since the adversary controls < ⅓ of the QAP). As a result, selecting the heaviest prefix is as simple as selecting the highest blockheight (greatest number of blocks), while ensuring all proposed prefixes that gather a strong quorum in QUALITY extend each other as a sanity check.
+The parameter max_lookahead_rounds may be set at the discretion of an implementation.
```
jsoares, line 579:
```suggestion
The parameter `max_lookahead_rounds` may be set at the discretion of an implementation. 
```

FIPS/fip-0086.md
```diff
@@ -591,32 +656,27 @@ OR (step=COMMIT                                            | valid COMMIT eviden
 OR (step = DECIDE                                          | valid DECIDE evidence
       AND (∃ M: IsStrongQuorum(M) AND evidence=Aggregate(M)| is a strong quorum 
         AND ∀ m’ ∈ M: m’.step = COMMIT                     | of COMMIT messages
-        AND ∀ m’,m'' ∈ M: m’.round = m''.round             | from the same round 
+        AND ∀ m’,m’’ ∈ M: m’.round = m’’.round             | from the same (any) round as each other
```
jsoares, line 659: Not sure the added "any" contributes to clarity.

FIPS/fip-0086.md
```diff
@@ -591,32 +656,27 @@ OR (step=COMMIT                                            | valid COMMIT eviden
...
+Note that during the validation of each CONVERGE, COMMIT, and DECIDE message, two signatures need to be verified: (1) the signature of the message contents, produced by the sender of the message (first check above) and (2) the aggregate signature in $m.evidence$ ($\texttt{ValidEvidence}(m)$ above). The former can be verified using the sender's public key (stored in the power table). The latter, being an aggregated signature, requires aggregating public keys of all the participants whose signatures it combines. Notice that the evidence of each message can be signed by a different set of participants, which requires aggregating a linear number of public BLS keys (in system size) per evidence verification. However, a participant only needs to verify the evidence once for each *unique* message (in terms of (step, round, value)) received.
```
jsoares, line 664:
```suggestion
Note that during the validation of each CONVERGE, COMMIT, and DECIDE message, two signatures need to be verified: (1) the signature of the message contents, produced by the sender of the message (first check above) and (2) the aggregate signature in $m.evidence$ ($\texttt{ValidEvidence}(m)$ above). The former can be verified using the sender's public key (stored in the power table). The latter, being an aggregated signature, requires aggregating public keys of all the participants whose signatures it combines. Notice that the evidence of each message can be signed by a different set of participants, which requires aggregating a linear number of BLS public keys (in system size) per evidence verification. However, a participant only needs to verify the evidence once for each *unique* message (in terms of (step, round, value)) received.
```

jsoares (commented):
part 2/3

FIPS/fip-0086.md
```diff
@@ -654,12 +718,12 @@ type PowerTableDelta struct {
...
 
-A finality certificate is analogous to a full block in the EC block exchange protocol.
-
+Certificates for the same instance and decision may vary across participants in which signatures are aggregated to provide the strong quorum.
 
 #### Exchange protocol
 
-Like with EC block exchange, participants follow **a simple request/response protocol to provide a contiguous sequence of finality certificates**. Note that the sequence of certificates is traversed backward, matching block exchange. A finality certificate contains the tipset and the PoF of the tipset ($Signature$). PoFs may vary across participants, but all PoFs must be for the same tipset at the same instance number $i$.
+A finality certificate is analogous to a full block in the EC block exchange protocol.
+Like with EC block exchange, participants follow **a simple request/response protocol to provide a contiguous sequence of finality certificates**. Note that the sequence of certificates is traversed backward, matching block exchange.
```
jsoares, line 726:
```suggestion
Like with EC block exchange, participants follow a simple request/response protocol to provide a contiguous sequence of finality certificates. Note that the sequence of certificates is traversed backward, matching block exchange. 
```

No reason to bold.

FIPS/fip-0086.md
```diff
@@ -765,13 +827,13 @@ The `ECTipset` contains 4 values:
...
+The tipset key is required as this is the main value F3 is finalizing. While off-chain light-clients will use the tipset key, cross-chain bridges likely won't. Thus, the tipset key is _hashed_ before signing (so the full tipset key need not be sent to bridge contracts, saving gas).
 
-We include a blake2b hash of the power table to aid in syncing the F3 chain. In theory we could have added it to the commitment merkle tree and left it out of the root `ECTipset` object, but being able to validate a chain of finality certificates without any additional merkle-proofs is convenient and the an extra 32 bytes is not an unreasonable cost (512 gas on Ethereum) .
+A blake2b hash of the power table is included to aid in syncing the F3 chain. In theory it could be added to the commitment merkle tree and left it out of the root `ECTipset` object, but being able to validate a chain of finality certificates without any additional merkle-proofs is convenient and the an extra 32 bytes is not an unreasonable cost (512 gas on Ethereum) .
```
jsoares, line 834:
```suggestion
A blake2b hash of the power table is included to aid in syncing the F3 chain. In theory, it could be added to the commitment merkle tree and left out of the root `ECTipset` object, but being able to validate a chain of finality certificates without any additional merkle-proofs is convenient and the extra 32 bytes is not an unreasonable cost (512 gas on Ethereum) .
```

FIPS/fip-0086.md
```diff
@@ -817,27 +882,21 @@ Additionally, the epoch comes _before_ the commitments as it lets the bridge arr
 
 #### Merkle Trees
 
-We put both the tipsets and the commitments inside tipsets into merkle trees to keep proof-sizes (for bridges) small, even if we have a large instance with many tipsets.
+Both the tipsets and the commitments inside tipsets are nested into merkle trees to keep proof-sizes (for bridges) small, even if a large instance has many tipsets.
```
jsoares, line 885:
```suggestion
Both the tipsets and the commitments inside tipsets are nested into merkle trees to keep proof sizes small, even if a large instance has many tipsets.
```

FIPS/fip-0086.md
```diff
@@ -817,27 +882,21 @@ Additionally, the epoch comes _before_ the commitments as it lets the bridge arr
...
 
-Our merkle tree is heavily based on the [merkle tree from CometBFT][comet-merkle], but that one:
+The merkle tree is heavily based on the [merkle tree from CometBFT][comet-merkle], but that one:
 
 1. Uses sha256 (not EVM friendly).
 2. Requires more complex math on validation. This lets it avoid a few hash operations in some cases, but that likely isn't worth the complexity.
 3. Needs additional out-of-band information to be a _true_ vector commitment. Optimization 2 in the CometBFT merkle tree means that a proof for item 7/7 can be treated as a proof of an item 6/6.
 
-Most other well-known merkle tree formats (verkle trees, etc.) optimize for sparse trees and where we have _full_ trees, so the complexity isn't worth it.
+Most other well-known merkle tree formats (verkle trees, etc.) optimize for sparse trees and where this protocol will have _full_ trees, so the complexity isn't worth it.
```
jsoares, line 893:
```suggestion
Most other well-known merkle tree formats (verkle trees, etc.) optimize for sparse trees whereas this protocol will have _full_ trees.
```

FIPS/fip-0086.md
```diff
@@ -40,71 +40,93 @@ We specify a mechanism for fast finality with the F3 component. F3 is expected t
...
+The actions of honest and rational participants are aligned by ensuring that rational participants never contribute to finalizing a chain they do not locally possess. Without this, If a participant $p$ were to finalize a chain it didn't have, then $p$ would not be able to mine in EC until it retrieved the required blocks and thus, in the meantime, could not obtain the cryptoeconomic rewards that EC offers. To this end, F3 provides _EC compatibility_.
 
 F3 and its implementation GossiPBFT provide additional robustness for strong adversaries that control more than ⅓ QAP, such as:
 
 * Censorship resistance against a coalition of rational participants trying to leak power, without providing firm guarantees.
 * Resilience to long-range attacks by an adversary controlling up to less than ⅔ QAP of any old power table.
 
 
-### Best-effort Broadcast and Gossipsub
+### Best-effort Broadcast and GossipSub
```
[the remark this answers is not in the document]
anorth: I'll change all to Gossipsub to match that project's naming.

FIPS/fip-0086.md
anorth: I've reworded this to make explicit the assumptions behind "honest".

FIPS/fip-0086.md
anorth: > Should we add that line here that the number of miners must be X to reach consensus?

There's no minimum on the number of participants, except implicitly 1. The protocol would be unable to recover from dropping below any such number.

FIPS/fip-0086.md
anorth: Changed to "finalize", and there were no more "ise" that aren't consistent between dialects.

FIPS/fip-0086.md
anorth: It was defined at 111 (= `10`).

FIPS/fip-0086.md
```diff
@@ -292,43 +336,53 @@ type PowerTableEntry struct {
...
 
-A set of messages $M$ that does not contain equivocating messages is called _clean_. Participants discard all equivocating messages when forming clean sets. In fact, in the proposed version of GossiPBFT in this document, only one of the equivocating messages needs to be stored (thanks to evidences explicitly validating a message).
+The receiver of a message only considers messages with matching supplemental data and valid signatures, and discards all other messages as invalid. The sender ID and signatures are omitted in further descriptions for better readability.
+
+Two (or more) messages $m1$ and $m2$ are called _equivocating messages_ if $m1.Sender=m2.Sender \land m1.Instance=m2.Instance \land m1.Value ≠ m2.Value \land m1.MsgType=m2.MsgType \land \texttt(if applicable)\ m1.Round=m2.Round$. $m1.Sender$ is called an _equivocating sender_.
```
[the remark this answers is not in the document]
anorth: Replaced with prose.

FIPS/fip-0086.md
```diff
@@ -354,170 +408,181 @@ Then, we build a [merkle tree](#merkle-tree-format) from these encoded `ECTipset
 
 #### GossiPBFT pseudocode (main algorithm)
 
-We illustrate the pseudocode for GossiPBFT below, consisting of 3 steps per round (QUALITY/CONVERGE, PREPARE, COMMIT) and an additional step outside the round loop (DECIDE). The $Sender$, $Signature$, and $Instance$ fields are omitted from messages for better readability. See also the simplified [PlusCal/TLA+ specification](https://github.com/filecoin-project/tla-f3).
+The GossiPBFT algorithm is specified by pseudo-code below. It consists of 3 steps per round (QUALITY/CONVERGE, PREPARE, COMMIT) and an additional step outside the round loop (DECIDE). Message fields that are constant for the instance are omitted for readability. 
+
+Messages are delivered to the participant in some order, whereupon a predicate over the set of received messages is evaluated, depending on the algorithm's current round and step.
+
+See also the simplified [PlusCal/TLA+ specification](https://github.com/filecoin-project/tla-f3).
```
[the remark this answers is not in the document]
anorth: Removed

FIPS/fip-0086.md
```diff
@@ -354,170 +408,181 @@ Then, we build a [merkle tree](#merkle-tree-format) from these encoded `ECTipset
...
+       // QUALTIY
 12:    if (round = 0)
-13:      BEBroadcast <QUALITY, value, instance>; trigger (timeout)
+13:      BEBroadcast <QUALITY, value>; trigger (timeout)
 14:      step ← QUALITY
 15:      collect a clean set M of valid QUALITY messages from this instance 
            until HasStrongQuorum(proposal, M) OR timeout expires
 17:      C ← C ∪ {prefix : IsPrefix(prefix,proposal) and HasStrongQuorum(prefix,M)}
-18:      proposal ← heaviest prefix ∈ C \* this becomes baseChain or sth heavier
+18:      proposal ← longest prefix ∈ C  // At least baseChain, or something longer
```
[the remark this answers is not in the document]
anorth: Changed the other uses of "heaviest" in F3 to "longest". F3 doesn't know or define chain weight. ("Heaviest" is still the right word when describing EC behaviour, e.g. the heaviest candidate chain)

FIPS/fip-0086.md
anorth: Removed the semicolons that are redundantly terminating statements.

PR #1027, opened 2024-06-27
Common Node API

ansermino:
A work-in-progress to establish a common API for all node implementations.

Many todos remain at this point, most notably:
- Further discussion on which methods are included and the associated rationale
- Sufficient conformance testing to validate implementations are adherent to the specification


jennijuju:
@filecoin-project/lotus-maintainers 👀 plz

Stebalien:
IMO, this should go under the FRC section instead of the FIP section as it's not a consensus critical change (and therefore doesn't have to go through the entire FIP process).

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,194 @@
...
+
+This interface is contrived from the existing [Lotus RPC V1 interface](https://github.com/filecoin-project/lotus/blob/48287574d82c81ffc50f259134c6596add32a2f5/documentation/en/api-v1-unstable-methods.md). Due to the large number of methods presently supported, we have limited the scope of this specification to include methods with a well established use case. We have observed the requirements of Lotus-miner, Curio, Web3Mine and Boost to inform which methods are actively used. We have also considered insights from Glif based on the usage of their RPC service.
+
+We target two groups of users: *end users* and *node operators*. These each have their own requirements for particular data and access control. While no attempt is made in this FIP to specify considerations  for these groups (eg. access control), these groups are used to justify the need for particular methods.
+
+
+Methods included in the API fall into one of these categories:
+
+- State Queries
+  - These primarily concern the chain state, actor state or network (p2p) state. They are read-only methods that are necessary to expose all elements of global state.
```
Stebalien, line 101: Right now we have a pretty _wide_ API here. Maybe a bit of a bikeshed, but I've always wanted to move some of the state-query logic into read-only methods on the actors themselves so state query operations simply become method implementations (simplifying client implementations).

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,194 @@
...
+Using the OpenRPC specification format we define several groups of methods. These methods are defined in the OpenRPC specification found here: https://github.com/ChainSafe/filecoin-openrpc-spec. A list of the methods is included below for posterity.
+
+All methods are reproduced identically to their present implementation in Lotus 1.26.1.
+
+Categories have been assigned to indicate the identified use case (see Design Rationale).
+
+Node implementers must include all specified methods in their API without any modifications. Implementers may choose to provide additional methods beyond what is included in this specification. The endpoint used to serve RPC requests must be versioned acccording the majo
+
+## Subscription methods
+The `ChainNotify` method is not a simple request/response, but rather a bidirectional exchange of JSON-RPC-like messages as follows:
```
Stebalien, line 56: Note: F3 will also introduce some form of `FinalizedChainNotify()` method which should be significantly simpler...

We should also consider:

1. Events (which currently use a different API).
2. Messages (there's actually no great "native" API for watching for messages).

jennijuju:
I am supportive of the motivation and direction of this proposal. Will give a more thorough review once the methods list are proposed

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,194 @@
+---
+fip: "<to be assigned>"
+title: Filecoin Common Node Interface V1
+author: "David Ansermino (@ansermino), Aatif Syed (@aatifsyed)"
+discussions-to: <URL>
+status: Draft
+type: FTP
```
jennijuju, line 7:
```suggestion
type: FIP - Interface
```

Stebalien:
FWIW, I think this is a great idea.

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,194 @@
...
+    ```json
+    { "jsonrpc": "2.0", "method": "xrpc.ch.val", params: [2, {}] }
+    ```
+4. Caller sends a JSON-RPC request to cancel the subscription, with their Channel ID
+    ```json
+    { "jsonrpc": "2.0", "id": 1, "method": "xrpc.cancel", params: [2] }
+    ```
+5. Callee responds with a JSON-RPC notification of cancellation
+    ```json
+    { "jsonrpc": "2.0", "id": 1, "method": "xrpc.ch.close", params: [2] }
```
rvagg, line 76:
```suggestion
    { "jsonrpc": "2.0", "method": "xrpc.ch.val", "params": [2, {}] }
    ```
4. Caller sends a JSON-RPC request to cancel the subscription, with their Channel ID
    ```json
    { "jsonrpc": "2.0", "id": 1, "method": "xrpc.cancel", "params": [2] }
    ```
5. Callee responds with a JSON-RPC notification of cancellation
    ```json
    { "jsonrpc": "2.0", "id": 1, "method": "xrpc.ch.close", "params": [2] }
```

aatifsyed: Thanks! We're actually reewriting this section - you can peep https://github.com/ChainSafe/FIPs/pull/2 to see the current progress :)

aatifsyed:
> IMO, this should go under the FRC section instead of the FIP section as it's not a consensus critical change (and therefore doesn't have to go through the entire FIP process).

@Stebalien We chose to be a technical FIP based on this guidance:
https://github.com/filecoin-project/FIPs/blob/020bcb412ee20a2879b4a710337959c51b938d3b/FIPS/fip-0001.md?plain=1#L30

Happy to move to FRC if that fits better - maybe the above could be amended?

ansermino:
> I am supportive of the motivation and direction of this proposal. Will give a more thorough review once the methods list are proposed

🙏 

We have a working doc with all the methods here: 
https://docs.google.com/spreadsheets/d/1fFkQuEjvFAd2s1dGX5zGmhxsEMLMUZ4uQFnIXgSZA5w/edit?usp=sharing

A few methods we're still unsure of, but we've taken a stance of most of them. Rationale is included for those that are excluded, and those that are included should fit into one of the categories we describe

FIPS/fip-common_node_api.md
ansermino: Great suggestion! 

Another thought we had is whether a generic state query method could be sufficient. Having a more direct relationship between what the actors return and the RPC responses would make this more feasible.

FIPS/fip-common_node_api.md
aatifsyed: Thanks for your comment Steven.

We're currently rewriting this section if you'd like to sneak a preview: https://github.com/ChainSafe/FIPs/pull/2

Could you clarify what the `Events` and `Messages` APIs you're talking about are, or maybe point me to some docs? I've not heard of them before.

jennijuju: https://github.com/filecoin-project/lotus/blob/f8c5b737efa10b84c649ae2471030d691902e688/api/api_full.go#L197 This is basically Filecoin’s version of eth_getlog

jennijuju: Messages are the ones started with `Mpool`

BigLep:
Since GitHub didn't backlink it, the accompanying discussion for this PR is https://github.com/filecoin-project/FIPs/discussions/1032

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,341 @@
...
+| StateMinerInfo                      | State query      | returns info about the specified miner                                                                                                                                                                                                                                                                                                                                                                                                         |
+| StateMinerInitialPledgeCollateral   | State query      | returns the initial pledge collateral for the specified miner's sector                                                                                                                                                                                                                                                                                                                                                                         |
```
rvagg, line 220: FYI we're going to have to deprecate `StateMinerInitialPledgeCollateral` and introduce a replacement because we can no longer use precommit to perform this calculation, so this is mostly useless now. Currently the new one is called `StateMinerInitialPledgeForSector` in https://github.com/filecoin-project/lotus/pull/12384

ansermino: Thanks for the heads up! WIll look into this and update the included methods :+1:

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,341 @@
...
+| ChainGetTipSetByHeight              | State query      | Returns the tipset at the specified height                                                                                                                                                                                                                                                                                                                                                                                                     |
+| ChainHasObj                         | State query      | checks if a given CID exists in the chain blockstore                                                                                                                                                                                                                                                                                                                                                                                           |
```
rvagg, line 156:
```suggestion
| ChainHasObj                         | Node operation   | Checks if a given CID exists in the chain blockstore                                                                                                                                                                                                                                                                                                                                                                                           |
```

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,341 @@
...
+| ChainNotify                         | State query      | Subscribe to changes to the chain head                                                                                                                                                                                                                                                                                                                                                                                                         |
+| ChainPutObj                         |                  | Puts a given object into the block store                                                                                                                                                                                                                                                                                                                                                                                                       |
```
rvagg, line 159:
```suggestion
| ChainPutObj                         | Node operation   | Puts a given object into the block store                                                                                                                                                                                                                                                                                                                                                                                                       |
```

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,341 @@
...
+| ChainReadObj                        | State query      | Reads IPLD nodes referenced by the specified CID from chain blockstore and returns raw bytes.                                                                                                                                                                                                                                                                                                                                                  |
```
rvagg, line 160:
```suggestion
| ChainReadObj                        | Node operation   | Reads IPLD nodes referenced by the specified CID from chain blockstore and returns raw bytes.                                                                                                                                                                                                                                                                                                                                                  |
```

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,341 @@
...
+We target two groups of users: *end users* and *node operators*. These each have their own requirements for particular data and access control. While no attempt is made in this FIP to specify considerations  for these groups (eg. access control), these groups are used to justify the need for particular methods.
+
+
+Methods included in the API fall into one of these categories:
+
+- State Queries
+  - These primarily concern the chain state, actor state or network (p2p) state. They are read-only methods that are necessary to expose all elements of global state.
+
+- Miner Operations
+  - A small set of specialized methods miner applications require to participate in consensus.
```
rvagg, line 263: yes, but arguably mpoolpush and wallet* operations are useful for "client" operations too; maybe a broader category name would be good here, or split it into two?

rvagg: although, MpoolPushMessage is "miner" and the other mpool* are "transactions"

maybe all mpool and all wallet should be "transactions" or make a new "client" category to replace "transactions"?

FIPS/fip-common_node_api.md
rvagg: Well, currently these operations read from _the_ blockstore, not some special segment of it that's for state, it'll read anything with a CID, messages, events, whatever. I'm mainly taking issue with it being "state" query. Not a big deal, just seems to be a slight misscategorisation. ("Chain" in the name is also slightly misleading, but less so than "state")

FIPS/fip-common_node_api.md
```diff
@@ -0,0 +1,341 @@
...
+Note that Lotus hosts two different API's at HTTP paths [`/rpc/v0`](https://github.com/filecoin-project/lotus/blob/v1.28.0-rc5/documentation/en/api-v0-methods.md) and [`/rpc/v1`](https://github.com/filecoin-project/lotus/blob/v1.28.0-rc5/documentation/en/api-v1-unstable-methods.md),
+with the latter being marked as _unstable_.
+This document aims to specify the `v1` API.
+
+Categories have been assigned to indicate the identified use case (see Design Rationale).
+
+Node implementers must include all specified methods in their API without any modifications. Implementers may choose to provide additional methods beyond what is included in this specification. The endpoint used to serve RPC requests must be versioned acccording the majo
+
+## Subscription methods
+Certain methods like `ChainNotify` are not oneshot request/response exchanges,
```
masih, line 61: I have brought up my concern regarding such push-based APIs [here](https://github.com/filecoin-project/lotus/pull/12719#issuecomment-2513994235). 

Without concrete usecases, clear delivery semantics (both in terms of guarantees, and latency) I advise against introducing such push-based APIs in favour of a simpler more maintainable pull-based API.

My nudge is to research the usecases carefully before introducing this style of API, and do it now because v2 APIs are a good place to de-clutter the APIs exposed for a set that is more focused on correctness+simplicity vs. featurefulness.

I believe that goes a long way as far as builders are concerned. 

I am keen to hear from the users of existing push-based APIs.

ansermino: Thanks for raising this! 

I believe it was the SP software that was highly dependent on `ChainNotify` in particular. This perhaps also suggests that an API for SPs should be defined independently of the "end user" API, they should maybe even use different protocols 🤷 

I will move to remove these from the proposed spec. As implementations already have support for this it should have no immediate impact on SPs.

Stebalien: `ChainNotify` has reasonably well-defined semantics (although they're not perfect):

- When you subscribe, you get a single event telling you the current "head", followed by all changes from that head.
- If you read slowely, you'll be unsubscribed.
- If you need the "full sequence", you can use `ChainGetPath` from your last tipset to the current head to fill in any missing pieces.

I'd prefer it if `ChainNotify(ctx)` were changed to `ChainNotify(ctx, from /* optional */)`, handling catch-up internally. But having a way to react to events immediately without having to poll is really nice.

NOTE: polling v. subscribing was a huge issue for F3 but, IMO, that was because subscribing had side effects (where `ChainNotify` does not).

masih: > where `ChainNotify` does not

Indeed. FWIW, even for F3 I don't think push would have made things any easier to get correct.

> having a way to react to events immediately without having to poll is really nice.

I have blindspots here; apologies. What is the concrete usecase that is common enough to grant having this push API in Common API specification?

BigLep:
2025-05-12 conversation with FilOz and Chainsafe:
- Need to add OpenAPI json spec
- Chainsafe will open a new PR

rvagg:
The Forest team is going to open this PR fresh since the original authors are no longer around. We're 95% of the way with this PR and just want to get something merged at this stage.

I'm assigning number **0104** to this FRC. @AlexeyKrasnoperov @LesnyRumcajs please use that number in your new PR.

BigLep:
Replacement PR: https://github.com/filecoin-project/FIPs/pull/1158

PR #1046, opened 2024-08-21
add lemmih as author

jennijuju:
hi @kkarrancsu @anorth I would like to propose adding @lemmih as a co-author of the fip, who supported fip spec session.


TippyFlitsUK (approved):

kkarrancsu:
Yes good idea - thanks for this catch @jennijuju

jsoares (approved):
Given 👍 from @kkarrancsu (existing author), I think we can merge.

PR #1052, opened 2024-08-27
Adjust ticket function in F3 to provide fairness

Kubuxu:
Adjust ticket function in F3 to resolve fairness issue in ticket selection.
Adjusting for power with uniformly distributed tickets leads to bias towards larger power holders which can lead to very long instance times. By transforming tickets into exponentially distributed random variables and adjusting them in that space we avoid that issue.

### DIFF
--- a/FIPS/fip-0086.md
+++ b/FIPS/fip-0086.md
@@ -390,10 +390,11 @@ A [merkle tree](#merkle-tree-format) is built from these encoded `ECTipset` valu
       ```
 * `StrongQuorumValue(M)`
     * Returns $v$ if $\texttt{HasStrongQuorumValue}(M)$ holds, $nil$ otherwise.
-* `GreatestTicketProposal(M)`
+* `BestTicketProposal(M)`
     * Let $M$ be a clean set of CONVERGE messages for the same round.
     * If $M = ∅$, the predicate returns $baseChain$.
-    * If $M \ne ∅$, the predicate returns $m.value$ and $m.evidence$, such that $m$ is the message in $M$ with the greatest ticket after adjusting for the power of the sender ($m.ticket*m.sender.power$). M will never be empty as the participant must at least deliver its own CONVERGE message.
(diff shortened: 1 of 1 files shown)


FIPS/fip-0086.md
```diff
@@ -390,10 +390,11 @@ A [merkle tree](#merkle-tree-format) is built from these encoded `ECTipset` valu
       ```
 * `StrongQuorumValue(M)`
     * Returns $v$ if $\texttt{HasStrongQuorumValue}(M)$ holds, $nil$ otherwise.
-* `GreatestTicketProposal(M)`
+* `BestTicketProposal(M)`
     * Let $M$ be a clean set of CONVERGE messages for the same round.
     * If $M = ∅$, the predicate returns $baseChain$.
-    * If $M \ne ∅$, the predicate returns $m.value$ and $m.evidence$, such that $m$ is the message in $M$ with the greatest ticket after adjusting for the power of the sender ($m.ticket*m.sender.power$). M will never be empty as the participant must at least deliver its own CONVERGE message.
+    * If $M \ne ∅$, the predicate returns $m.value$ and $m.evidence$, such that $m$ is the message in $M$ with the smallest ticket after adjusting for the power of the sender ($\frac{-log(t)}{m.sender.power}$). M will never be empty as the participant must at least deliver its own CONVERGE message.
```
Stebalien, line 396: So... about that second part. What if the local node isn't participating?

Stebalien (approved):

Stebalien:
Explanation for other reviewers: this aligns the FIP with the original intent by ensuring that the probability of a participant winning a ticket is proportional to their power. I.e., a participant with 2/3rds of the power now draw a winning ticket 2/3rds of the time.

FIPS/fip-0086.md
Kubuxu: I don't think that is in the scope of the FIP, but in go-f3, we have the self-delivery mechanism.

Stebalien: > we have the self-delivery mechanism

Yeah, but what if we're not even participating? This is definitely an implementation question, I'm just making sure we're not making any assumptions.

PR #1056, opened 2024-09-06
correct Title FIP number

lanzafame:
FIP-0009 was title with FIP-0008


rvagg (approved):
there's also a `replaces:` that's been used in 0015

and we should probably use `superseded-by` in the ones that are pointed to with a `replaces:` like 0003 and 0090

lanzafame:
The title of 0015 makes it fairly obvious but happy to add that to this PR to make the changes clear.

Happy to open a PR for 0003 and 0090 later, unless you want it @rvagg?

jennijuju (approved):

PR #1058, opened 2024-09-06
Move to Final status

lanzafame:
I believe all three implementations have implemented this FIP.


jennijuju:
This code is in actor, which we only have one implementation of today and it is finalized there! Can you update the readme too?

jennijuju (approved):

lanzafame:
The FIP table in the readme already has FIP-0029 listed as Final. @jennijuju That is the place you were referring to correct?

jsoares (approved):

PR #1079, opened 2024-10-19
FIPTBD: Simplify termination fee calculation to a fixed percentage of initial pledge

Schwartz10:
Computing the collateral value for a miner actor is critical for making economically secure defi protocols on filecoin; however, computing the collateral value correctly is hard on its own, and impossible in a small time frame and/or from inside the FEVM. The variance in termination fees relative to initial pledge for different miners on the network vary so significantly that DeFi protocols cannot safely use heuristics to estimate miner collateral values.

This FIP intends to significantly simplify the termination fee calculation by using a fixed "percentage-of-pledge" method. The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector.

[Discussion](https://github.com/filecoin-project/FIPs/discussions/1036)

### DIFF
--- a/FIPS/fip-draft_term_fee.md
+++ b/FIPS/fip-draft_term_fee.md
@@ -50,7 +50,14 @@ This FIP intends to keep those original motivations in tact for the Filecoin net
 
 <!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
 
-The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations.
+A new constant variable called `TERM_PENALTY_PLEDGE_PERCENTAGE` should be created, which is fixed for the whole network and configurable by future governance.
+
+Refactor the `pledge_penalty_for_termination` method to:
+
+1. Take the sector's `initial_pledge` `TokenAmount` as an argument
+2. Return `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+
+We may also be able to remove `expected_storage_pledge` from the `SectorOnChainInfo` struct type as I believe this is only used in computing termination penalties.
 
 ## Design Rationale
 
--- a/FIPS/fip-draft_term_fee.md
+++ b/FIPS/fip-draft_term_fee.md
@@ -1,0 +1,114 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
+title: Simplify termination fee calculation to a fixed percentage of initial pledge
+author: @anorth, @Schwartz10
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/1036
+status: Draft
+type: Technical
+category (*only required for Standard Track): Core
+created: 2024-09-26
+---
+
+<!--You can leave these HTML comments in your merged FIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new FIPs. Note that a FIP number will be assigned by an editor. When opening a pull request to submit your FIP, please use an abbreviated title in the filename, `fip-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+
+# FIP-Number: Simplify termination fee calculation to a fixed percentage of initial pledge
+
+## Simple Summary
+
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+Computing the collateral value for a miner actor is critical for making economically secure defi protocols on filecoin; however, computing the collateral value correctly is hard on its own, and impossible in a small time frame and/or from inside the FEVM. The variance in termination fees relative to initial pledge for different miners on the network vary so significantly that DeFi protocols cannot safely use heuristics to estimate miner collateral values.
+
+This FIP intends to significantly simplify the termination fee calculation by using a fixed "percentage-of-pledge" method. The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector.
+
+## Abstract
+
+<!--A short (~200 word) description of the technical issue being addressed.-->
+
+Today, it is overly sophisticated and computationally expensive to compute termination fees for miner actors. As a result, DeFi applications that leverage miner actors must make major security and/or performance sacrifices in order to operate.
+
+This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 6%.
+
+As a result, DeFi applications on Filecoin can operate with significantly better performance, UX, and economic security. Additionally, Filecoin economics and code implementations will be significantly simplified, as well as state bloat removed.
+
+## Change Motivation
+
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+
+In the year+ since FEVM’s launch, we’ve seen a number of protocols use Miner Actors as collateral for lending or leasing applications. Economic security of applications being built on the FEVM is important for developing trust, and understanding precise collateral values for any given miner actor is critical for a DeFi protocol’s economic security.
+
+Today, it is: (1) computationally expensive, (2) economically sophisticated, and (3) impossible in a FEVM runtime to compute the maximum termination fee for a miner actor’s sectors. As a result, FEVM applications need to either (a) use heuristics to determine a collateral value for a miner (economically insecure), or (b) use off-chain solutions to compute collateral values (significantly bloating the surface area of attack vectors for any lending / leasing protocol). Both of these approaches make major sacrifices towards the economic security of FEVM applications that use miner actors as collateral.
+
+Two of the original motivations for termination fees are:
+
+1. Create a more resilient storage network for storage clients, such that if an SP stores a client's data, the client isn't just suddenly left out to dry and their data lost
+2. Network stability
+
+This FIP intends to keep those original motivations in tact for the Filecoin network.
+
+## Specification
+
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations.
+
+## Design Rationale
+
+<!--The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale may also provide evidence of consensus within the community, and should discuss important objections or concerns raised during discussion.-->
+
+There are a few different approaches one could take to accomplish the same end result:
+
+1. Optimize data structures based on the current methodology to make an aggregate termination sector method feasible in constant time lookup
+   - Problems:
+     - Unsure if this is technically feasible, and if so, most likely a lot of sophistication and ugly accounting
+     - Would most likely add state to the Filecoin blockchain
+2. Use a "multiple of daily rewards" approach, which computes a termination fee based on a % or multiple of expected daily rewards for a period of time based on the current block rewards
+   - Problems:
+     - Unpredictable - DeFi apps now have to be extremely knowledgeable on the future of Filecoin block rewards in order to safely underwrite DeFi loans
+     - Computing block rewards is still a non-trivial task, and most likely requires more/new solidity FEVM precompiles
+
+The current design is the simplest and most predictable approach to accomplishing the end goal.
+
+## Backwards Compatibility
+
+<!--All FIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The FIP must explain how the author proposes to deal with these incompatibilities. FIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+
+This FIP would introduce backwards incompatibility with how previous tipsets calculate their termination fees. As a result, previous statistical data collected about Filecoin termination fees (which I don't think there's much of), would become incorrect. It's possible to maintain the historical state, but it seems like a lot of work for a little gain.
+
+## Test Cases
+
+<!--Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
+
+- When terminating a single sector, the realized termination fee should equal the expected % of pledge termination fee
+- When terminating all sectors on a miner, the realized termination fee should equal the expected % of pledge termination fee
+
+## Security Considerations
+
+<!--All FIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. FIP submissions missing the "Security Considerations" section will be rejected. A FIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
+
+The security concern with introducing this FIP is that Storage Providers will not face a severe enough penalty to leave the network very quickly, which could cause major power volatility on the network. We don't believe this is a major concern for the following reasons:
+
+1. As Filecoin matures, the average Storage Provider will likely be accepting paid deals in some form according to an SLA in addition to the SLA guaranteed by PoRep. These additional SLAs can/will enforce their own termination clauses, which should provide adequate motivation and verification to achieve a good experience for the average storage client.
+2. It seems appropriate to charge an exit fee for breaking a commitment to the network, but we also do not want to charge a capture fee. Storage Providers who do not wish to use Filecoin anymore shouldn't have prohibitively high expenses for leaving the network early - ultimately for the network, burning some of this SPs tokens through sector terminations is better than 18 months of selling.
+3. The 6% proposed termination penalty percentage is what the network is currently charging (on average) for terminations.
+4. Governance can always choose to increase the termination penalty percentage.
+
+## Incentive Considerations
+
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
+
+It is important that Storage Providers are incentivized to honor their commitments to the network - this prevents major network volatility. It makes sense to maintain a disincentive for breaking commitments to the network, but this FIP doesn't change the incentive to provide useful storage to the network, it makes it somewhat easier.
+
+## Product Considerations
+
+<!--All FIPs must contain a section that discusses the product implications/considerations relative to the proposed change. Include information that might be important for product discussion. A discussion on how the proposed change will enable better storage-related goods and services to be developed on Filecoin. FIP submissions missing the "Product Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Product Considerations discussion deemed sufficient by the reviewers.-->
+
+From a product perspective, the network will massively simplify its UX. Potential Storage Providers can calculate their risk of mining on the back of a napkin, whereas today it takes a team of highly trained Filecoin experts. Developers can integrate safer DeFi practices.
+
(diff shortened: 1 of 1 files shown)


FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
+
+A new constant variable called `TERM_PENALTY_PLEDGE_PERCENTAGE` should be created, which is fixed for the whole network and configurable by future governance.
+
+Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
+
+1. Take the sector's `initial_pledge` `TokenAmount` as an argument
+2. Return `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+
+We may also be able to remove `expected_storage_pledge` from the `SectorOnChainInfo` struct type as I believe this is only used in computing termination penalties.
```
anorth, line 60: The following sector info fields are used exclusively for the termination fee calculation:
- expected day reward
- expected storage pledge
- replaced day reward

This spec should specify that:
- these three values are unused by the built-in actors code
- for all new sectors or snapped sectors, the values are set to zero

You can note that these redundant fields can be removed from state in a future migration.

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+
+There are a few different approaches one could take to accomplish the same end result:
+
+1. Optimize data structures based on the current methodology to make an aggregate termination sector method feasible in constant time lookup
+   - Problems:
+     - Unsure if this is technically feasible, and if so, most likely a lot of sophistication and ugly accounting
+     - Would most likely add state to the Filecoin blockchain
+2. Use a "multiple of daily rewards" approach, which computes a termination fee based on a % or multiple of expected daily rewards for a period of time based on the current block rewards
+   - Problems:
+     - Unpredictable - DeFi apps now have to be extremely knowledgeable on the future of Filecoin block rewards in order to safely underwrite DeFi loans
```
anorth, line 74: Less predictable, not unpredictable.

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+     - Computing block rewards is still a non-trivial task, and most likely requires more/new solidity FEVM precompiles
```
anorth, line 75: It's not that hard. It would not be a big ask to export a block reward projection method from, e.g., the power or reward actor. Given the proposed change, this is hardly a problem.

I think the multiple of rewards option should be laid out as a serious alternative, but then explain that the %-of-pledge method is (1) even simpler to calculate, (2) fixed while the sector's pledge is constant, (3) will not decay towards zero as block rewards decline, thus more effectively protecting against churn.

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+
+Today, it is: (1) computationally expensive, (2) economically sophisticated, and (3) impossible in a FEVM runtime to compute the maximum termination fee for a miner actor’s sectors. As a result, FEVM applications need to either (a) use heuristics to determine a collateral value for a miner (economically insecure), or (b) use off-chain solutions to compute collateral values (significantly bloating the surface area of attack vectors for any lending / leasing protocol). Both of these approaches make major sacrifices towards the economic security of FEVM applications that use miner actors as collateral.
+
+Two of the original motivations for termination fees are:
+
+1. Create a more resilient storage network for storage clients, such that if an SP stores a client's data, the client isn't just suddenly left out to dry and their data lost
+2. Network stability
```
anorth, line 45: It might be worth expanding on what network stability means. IMO one part of it is that there is some friction to rapid on-then-offboarding of power and pledge.

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+   - Problems:
+     - Unpredictable - DeFi apps now have to be extremely knowledgeable on the future of Filecoin block rewards in order to safely underwrite DeFi loans
+     - Computing block rewards is still a non-trivial task, and most likely requires more/new solidity FEVM precompiles
+
+The current design is the simplest and most predictable approach to accomplishing the end goal.
+
+## Backwards Compatibility
+
+<!--All FIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The FIP must explain how the author proposes to deal with these incompatibilities. FIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+
```
anorth, line 82: Note that this changes built-in actor code and so requires a network upgrade.

Note that it makes those three sector state fields redundant and set them to zero, so off-chain tooling that uses them must change how it works. (You're kinda hinting at that below)

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+   - Problems:
+     - Unsure if this is technically feasible, and if so, most likely a lot of sophistication and ugly accounting
+     - Would most likely add state to the Filecoin blockchain
+2. Use a "multiple of daily rewards" approach, which computes a termination fee based on a % or multiple of expected daily rewards for a period of time based on the current block rewards
+   - Problems:
+     - Unpredictable - DeFi apps now have to be extremely knowledgeable on the future of Filecoin block rewards in order to safely underwrite DeFi loans
+     - Computing block rewards is still a non-trivial task, and most likely requires more/new solidity FEVM precompiles
+
+The current design is the simplest and most predictable approach to accomplishing the end goal.
+
```
anorth, line 78: This design rational section needs a good rationale for the 6% proposal. Can you demonstrate what the current effective % is by showing the calculation for a new sector using recent network state?

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+3. The 6% proposed termination penalty percentage is what the network is currently charging (on average) for terminations.
+4. Governance can always choose to increase the termination penalty percentage.
+
+## Incentive Considerations
+
+<!--All FIPs must contain a section that discusses the incentive implications/considerations relative to the proposed change. Include information that might be important for incentive discussion. A discussion on how the proposed change will incentivize reliable and useful storage is required. FIP submissions missing the "Incentive Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Incentive Considerations discussion deemed sufficient by the reviewers.-->
+
+It is important that Storage Providers are incentivized to honor their commitments to the network - this prevents major network volatility. It makes sense to maintain a disincentive for breaking commitments to the network, but this FIP doesn't change the incentive to provide useful storage to the network, it makes it somewhat easier.
```
anorth, line 107: This section needs to clearly call out that the termination fee for sectors committed a long time ago will immediately decrease, significantly (but only to match the fee:pledge ratio of new sectors). This will increase the incentive for some providers with high-pledge sectors to terminate and re-onboard to gain more power for the same amount of pledge (but more hardware).

anorth, line 107: Perhaps also worth noting that, in contrast to the current system, the TF for a sector will be constant regardless of age. Currently the TF starts low and increases with age up to a max. This is a change in the incentive structure for short-term-view SPs.

anorth (commented):
This first review contains a mix of editorial and technical/peer review.

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
+---
+fip: "<to be assigned>" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
+title: Simplify termination fee calculation to a fixed percentage of initial pledge
+author: @anorth, @Schwartz10
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/1036
+status: Draft
+type: Technical
+category (*only required for Standard Track): Core
```
rvagg, line 8:
```suggestion
category: Core
```

FIPS/fip-draft_term_fee.md
rvagg: Is there a straightforward way to grab some basic data about this so we can attach a sense of scale to the incentive problem you're outlining @anorth ?

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,121 @@
...
+
+The security concern with introducing this FIP is that Storage Providers will not face a severe enough penalty to leave the network very quickly, which could cause major power volatility on the network. We don't believe this is a major concern for the following reasons:
+
+1. As Filecoin matures, the average Storage Provider will likely be accepting paid deals in some form according to an SLA in addition to the SLA guaranteed by PoRep. These additional SLAs can/will enforce their own termination clauses, which should provide adequate motivation and verification to achieve a good experience for the average storage client.
+2. It seems appropriate to charge an exit fee for breaking a commitment to the network, but we also do not want to charge a capture fee. Storage Providers who do not wish to use Filecoin anymore shouldn't have prohibitively high expenses for leaving the network early - ultimately for the network, burning some of this SPs tokens through sector terminations is better than 18 months of selling.
+3. The 6% proposed termination penalty percentage is what the network is currently charging (on average) for terminations.
```
rvagg, line 100: Echoing Alex above, this is the only part that gives a rationale for choosing 6% but it doesn't provide a means to verify. Is there a source for this we can either point to or call out? Can I, as a reader, verify this claim somehow?

FIPS/fip-draft_term_fee.md
Schwartz10: Semantics but I'd argue its unpredictable - you have to know if Filecoin is going to grow or contract

FIPS/fip-draft_term_fee.md
Schwartz10: Here's some data anorth collected in the discussion https://github.com/filecoin-project/FIPs/discussions/1036#discussioncomment-10215706

Schwartz10:
@anorth @rvagg agree with both of you that we need more data / verification around the `TERM_PENALTY_PLEDGE_PERCENTAGE` variable. 

GLIF had built a tool in this PR https://github.com/glifio/go-pools/pull/84/files that estimates the termination fee for an amount of FIL pledged in sectors. We can update this tool and collect some data. Anyone is welcome to use it too, but its currently hardcoded to a specific timestamp when we were testing.

Give us a little bit of time and we'll collect the necessary data to properly set `TERM_PENALTY_PLEDGE_PERCENTAGE`. The rest of the review comments have been resolved in https://github.com/filecoin-project/FIPs/pull/1079/commits/40cf2c2102cdbd21111f50291b443ca5638059d3

FIPS/fip-draft_term_fee.md
Schwartz10: Yeah thanks for calling this out - see [here](https://github.com/filecoin-project/FIPs/pull/1079#issuecomment-2430847623)

FIPS/fip-draft_term_fee.md
Schwartz10: See comment here https://github.com/filecoin-project/FIPs/pull/1079#issuecomment-2430847623

README.md
```diff
@@ -131,3 +131,4 @@ This improvement protocol helps achieve that objective for all members of the Fi
 | [0092](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0092.md) | Non-Interactive PoRep | FIP | luca (@lucaniz), kuba (@Kubuxu), nicola (@nicola), nemo (@cryptonemo), volker (@vmx), irene (@irenegia), Alex North (@anorth), orjan (@Phi-rjan) | Final |
 | [0094](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0094.md) | Add Support for EIP-5656 (MCOPY Opcode) in the FEVM | FIP | Michael Seiler (@snissn), Raúl Kripalani (@raulk), Steven Allen (@stebalien) | Accepted |
 | [0095](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0095.md) | Add FEVM precompile to fetch beacon digest from chain history | FIP | @ZenGround0, Alex North (@anorth) | Accepted |
+| [XXXX](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-XXXX.md) | Add FEVM precompile to fetch beacon digest from chain history | FIP | Jonathan Schwartz (@Schwartz10), Alex North (@anorth) | Draft |
```
jsoares, line 134:
```suggestion
| [XXXX](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-XXXX.md) | Simplify termination fee calculation to a fixed percentage of initial pledge | FIP | Jonathan Schwartz (@Schwartz10), Alex North (@anorth) | Draft |
```

Schwartz10:
Thanks to the help of @rvagg and @jimpick for collecting data for this FIP.

A few different data points have been collected:

1. @rvagg exported the full, raw SectorOnChainInfo. Copying and pasting from [this slack thread](https://filecoinproject.slack.com/archives/C01EU76LPCJ/p1731960003828789) - 

ok, here’s the full SectorOnChainInfo, without the CIDs in it: https://filvagg.s3.amazonaws.com/soci.csv.zst
7.1Gb compressed, 63G uncompressed
you could pull it and only store a subset using something like:
```
curl -s https://filvagg.s3.amazonaws.com/soci.csv.zst | zstdcat | awk -F, 'NR==1; NR>1 && $6 > 4000000 && rand() <= 0.01' > soci-part.csv
```
^ that would keep the header line and only entries whose expiration epoch is > 4000000 and a random sample of 1/100 of those (as an example).

no termination fee calculations or anything in here, just raw data

2. @jimpick used GLIF's sector terminator code to create this tool - https://github.com/glifio/simplify-terminations-fip-data, which produced this csv https://gist.github.com/jimpick/ee8d35fc40841fb9c611b6a3743816aa. I've copied this CSV into google sheets and added another column with the term penalty % of initial pledge [here](https://docs.google.com/spreadsheets/d/1P5ioPq7z50oNt45Tu1ezsNM7YwDjzNYhObY-h-_AgOE/edit?gid=306466513#gid=306466513).

Some stats that I calculated about the termination penalty % of pledge based on the CSV:
```
Average: 18.9%
Median: 11.2%
Min: 1.9%
Max: 184%
```

It's not obvious to me that we should use aggregate statistics based on the current state of the network to arrive at the correct % for this fip, but it's quite helpful information to have to inform the decision. 

3. @jimpick wrapped the `StateMinerInitialPledgeForSector` (method [here](https://github.com/filecoin-project/lotus/commit/06bec3a72fa1c3a60e94fcb7194633ad51213487)) to create another cli tool for computing the termination penalty % of initial pledge for new sectors (or sectors at a specific height). That tool can be found here https://github.com/glifio/simplify-terminations-fip-data/blob/main/cmd/penalty-on-pledge.go. 

According to the data Jim last computed, it looks like the new sector onboarding has a termination penalty % of pledge around 1.8%, which seems far too low. This will change presumably with the introduction of FIP81 economics.

anorth:
I agree we shouldn't base the constant fee on average/mean/min/max of existing miners nor sectors. We should use approximately the current value, so that the introduction of the change doesn't affect the economics of new sectors much.

I too am skeptical of 1.8% being the current value, based on my own sector observations from some months ago (~8%). Perhaps Jim could check the logic there? I also have a calculator (approximate) for it in https://docs.google.com/spreadsheets/d/1Z9Tjf_sMWU9nsnEpghHWCazSicSPREkWOENxi3yzZEM/edit?gid=0#gid=0, but need up to date network numbers to compute the present value

anorth:
Updated analysis on the raw sector data from @rvagg : https://github.com/filecoin-project/FIPs/discussions/1036#discussioncomment-11341308

jimpick:
I converted the termination fee calculation from the rust code to a Google Sheets spreadsheet...

(except for smooth::extrapolated_cam_sum_of_ratio() ... I just copied the values)

https://docs.google.com/spreadsheets/d/1-kT9IWEtFfNGkr8SxqqIWgYkumNCRDRakORDLx33J5s/edit?usp=sharing

I selected a recent sector and used the sector info + the power and reward actor values.

jimpick:
> I agree we shouldn't base the constant fee on average/mean/min/max of existing miners nor sectors. We should use approximately the current value, so that the introduction of the change doesn't affect the economics of new sectors much.
> 
> I too am skeptical of 1.8% being the current value, based on my own sector observations from some months ago (~8%). Perhaps Jim could check the logic there? I also have a calculator (approximate) for it in https://docs.google.com/spreadsheets/d/1Z9Tjf_sMWU9nsnEpghHWCazSicSPREkWOENxi3yzZEM/edit?gid=0#gid=0, but need up to date network numbers to compute the present value
> 
> Update: My calculator gives 8.8%. It's using a slightly simplified calculation, but let's compare with the value calculated from Go code (which might have an error in inputs) and the value observed in sectors recently committed.

Looking at the values in the spreadsheet and comparing to what I see with a real live sector (see my modified spreadsheet I posted), the first thing I notice is that the "Network Raw Power" for 2024-11-20 is 10.7EiB on the spreadsheet, but it should be ~4.4EiB. The other numbers look off too. I'm wondering if those values are from a different date?

jimpick:
I notice in the spreadsheet that the formula for "Termination Fee Simplified (FIL+ Sector)" in cell C28 is:

`=(70+20)*2880 * $B$10*C$15/$B$7`

That seems to imply that it's not for terminating a sector immediately after it was added to the network, since it's using 90 days, not 20 days. So that's probably a case for the discrepancy. If I change it to just 20 days, and I correct the Network QAP, Circulating Supply and Epoch reward, the numbers are pretty close.

anorth:
> That seems to imply that it's not for terminating a sector immediately after it was added to the network, since it's using 90 days, not 20 days. So that's the main source of the discrepancy. If I change it to just 20 days, and I correct the Network QAP, Circulating Supply and Epoch reward, the numbers are pretty close.

Right, that makes sense. And indeed I had intended to model the steady state termination fee (the ramp is another undesirable feature IMO).

Since the new formulation will apply to all sectors, most of which will be "old", I think it should aim for parity with the steady state (90 days) value, not the just-onboarded value. This will be disadvantageous to SPs intending to onboard and then shortly remove sectors, but I don't think pandering to them is a goal.

Schwartz10:
Hey @anorth i have updated the FIP to reflect the 8.5% termination penalty %. I also added @jimpick as an author as he's helped with the analysis

anorth (commented):
Great, thanks @Schwartz10 , I think this is about ready to merge. 

Please take # FIP-0098 and update the headers, title, and index page. This will then be good to merge in my view.

FIPS/fip-draft_term_fee.md
```diff
@@ -0,0 +1,143 @@
...
+
+1. As Filecoin matures, the average Storage Provider will likely be accepting paid deals in some form according to an SLA in addition to the SLA guaranteed by PoRep. These additional SLAs can/will enforce their own termination clauses, which should provide adequate motivation and verification to achieve a good experience for the average storage client.
+2. It seems appropriate to charge an exit fee for breaking a commitment to the network, but we also do not want to charge a capture fee. Storage Providers who do not wish to use Filecoin anymore shouldn't have prohibitively high expenses for leaving the network early - ultimately for the network, burning some of this SPs tokens through sector terminations is better than 18 months of selling.
+3. The 8.5% proposed termination penalty percentage is what the network is currently charging (on average) for steady state terminations.
+4. Governance can always choose to increase the termination penalty percentage.
```
lanzafame, line 119:
```suggestion
1. The 8.5% proposed termination penalty percentage is what the network is currently charging (on average) for steady state terminations.
2. Governance can always choose to increase the termination penalty percentage.
3. It seems appropriate to charge an exit fee for breaking a commitment to the network, but we also do not want to charge a capture fee. Storage Providers who do not wish to use Filecoin anymore shouldn't have prohibitively high expenses for leaving the network early - ultimately for the network, burning some of this SPs tokens through sector terminations is better than 18 months of selling.
4. As Filecoin matures, the average Storage Provider will likely be accepting paid deals in some form according to an SLA in addition to the SLA guaranteed by PoRep. These additional SLAs can/will enforce their own termination clauses, which should provide adequate motivation and verification to achieve a good experience for the average storage client.
```
I suggest reordering based on current concrete reasons security is maintained, i.e. fee the same, governance, followed by incentives and lastly by future constraints. Personally, the SLA point shouldn't be there as it is not a currently reality and therefore doesn't offset any security concerns with this FIP. Ideally, it is a thing in the future though.

lanzafame:
Just adding here the directly impacted spec.filecoin.io sections:
`section-systems.filecoin_mining.sector.sector-faults`
`section-algorithms.cryptoecon.initial-parameter-recommendation`

FIPS/fip-draft_term_fee.md
Schwartz10: Yeah i don't mind reordering. As for the SLA point - this is happening today, its not a majority of the network, but its also not 0.

Schwartz10:
All set @anorth ! Let me know if any other changes are required

FIPS/fip-0098.md
```diff
@@ -0,0 +1,143 @@
+---
+fip: "0098" <!--keep the qoutes around the fip number, i.e: `fip: "0001"`-->
```
rvagg, line 2:
```suggestion
fip: "0098"
```
(optional nit)

rvagg (approved):

TippyFlitsUK (approved):

PR #1084, opened 2024-11-23
Add Support for EIP-1153 (Transient Storage) in the FEVM (FIP-0097)

snissn:
## Abstract
EIP-1153 defines transient storage as temporary data accessible only during the originating transaction, cleared automatically at the end of the transaction. This FIP adapts transient storage to the FEVM using `TLOAD` and `TSTORE` opcodes, ensuring compatibility with Ethereum contracts and libraries that rely on this feature. Lifecycle validation mechanisms enforce the transaction-scoped behavior of transient storage, achieving functional equivalence with Ethereum’s implementation and seamless integration with Filecoin’s architecture.

## Discussion:
- https://github.com/filecoin-project/FIPs/discussions/855


FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+created: 2024-11-19  
+---
+
+# FIP-XXXX: Add Support for EIP-1153 (Transient Storage) in the FEVM
+
+## Simple Summary
+This proposal introduces support for **[EIP-1153: Transient Storage](https://eips.ethereum.org/EIPS/eip-1153)** in the Filecoin Ethereum Virtual Machine (FEVM). Transient storage provides ephemeral data storage during transaction execution. To maintain compatibility with Ethereum contracts utilizing this feature, the FEVM will implement transient storage with lifecycle validation.
+
+## Abstract
+EIP-1153 defines transient storage as temporary data accessible only during the originating transaction, cleared automatically at the end of the transaction. This FIP adapts transient storage to the FEVM using `TLOAD` and `TSTORE` opcodes, ensuring compatibility with Ethereum contracts and libraries that rely on this feature. Lifecycle validation mechanisms enforce the transaction-scoped behavior of transient storage, achieving functional equivalence with Ethereum’s implementation and seamless integration with Filecoin’s architecture.
```
rvagg, line 18: should probably same something about it also being scoped _to a contract_, such that nested contract calls don't have access to this memory space - scoped strictly to transaction + contract

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+## Change Motivation
+Transient storage offers developers temporary, transaction-scoped storage for managing intermediate states. One key benefit of this feature is its ability to improve the implementation of reentrancy locks, enhancing security by mitigating risks associated with multiple calls to a function within sub-transactions. By introducing transient storage, the FEVM aligns with Ethereum’s tooling and Solidity’s modern features, providing a seamless developer experience while supporting advanced contract use cases and secure computing.
```
rvagg, line 22: note about availability in Ethereum - I believe this went live in March with the Decun (Cancun) upgrade, so we're trailing already by 7 months

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+  - `key`: Location to store the value  
+  - `value`: Value to store  
+- **Output:** None  
+- **Description:** Stores `value` at the specified `key` in the transient storage for the current transaction.
+
+### Lifecycle Management
+Transient storage is valid only within the context of a single transaction. A lifecycle mechanism tracks transaction metadata (`origin` and `nonce`) to enforce lifecycle validation. 
+
+### Implementation Details
+The FEVM implements transient storage using a lifecycle validation mechanism to ensure data remains accessible only during the originating transaction. This validation enforces the same behavior as Ethereum’s ephemeral transient storage. Internally, transient storage relies on permanent storage to manage lifecycle data and state while ensuring functional adherence to Ethereum’s behavior.
```
rvagg, line 50: note about isolation between contracts that may get stacked during the transaction

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+---
+
+## Design Rationale
+
+The design adheres to the intent of EIP-1153 while adapting to Filecoin's architecture. The use of lifecycle validation ensures transient storage behaves as expected within the scope of a single transaction. This approach balances compatibility with Ethereum contracts and simplifies implementation within the existing FEVM architecture. 
+
+Alternative designs, such as purely in-memory storage, were considered but deemed impractical due to technical implementation difficulties.
+
+## Backwards Compatibility
+The addition of transient storage is fully backward-compatible. Existing contracts remain unaffected unless they utilize `TLOAD` and `TSTORE`. Contracts compiled with older Solidity versions will continue to execute without changes.
```
rvagg, line 61: note about solidity compiler and mode in here at which point compatibility breaks -  v0.8.24 with "Cancun" I believe? Antyhing before that should remain unaffected but after that will support TLOAD and TSTORE once this goes live

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Test Cases
+
+### Essential Tests
+1. **Basic Functionality:**
+   - Verify `TLOAD` retrieves the correct value.
+   - Verify `TSTORE` writes data to the transient storage correctly.
+
```
rvagg, line 69:
```suggestion
   - Verify `TSTORE` writes data to the transient storage correctly.
   - Verify `TLOAD` from an unitialized location returns the zero value.

```

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+2. **Lifecycle Validation:**
+   - Verify that transient storage is automatically cleared and becomes inaccessible after the transaction ends.
+   - Verify that transient storage is properly cleared at the end of each transaction and any out-of-lifecycle data does not interfere with subsequent transaction operations.
+
```
rvagg, line 73:
```suggestion
   - Verify that transient storage is properly cleared at the end of each transaction and any out-of-lifecycle data does not interfere with subsequent transaction operations.
   - Verify that nested contracts do not have access to shared memory spaces and can read and write independently.
   - Verify that memory remains accessible and stable after contract re-entry.

```

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,102 @@
...
+---
+
+## Security Considerations
+Transient storage introduces minimal additional risk compared to existing state storage. Lifecycle validation ensures storage is inaccessible outside the originating transaction. Security measures include:
+- Preventing out-of-bounds memory access.
+- Ensuring transient storage clears properly after a transaction ends.
+
```
rvagg, line 80:
```suggestion
- Ensuring transient storage clears properly after a transaction ends.
- Ensuring nested contracts do not have access to eachother's memory spaces.

```

rvagg (changes requested):
:+1: excellent; just some inline notes about some things I think are missing from here

FIPS/fip-draft_transient_storage.md
snissn: thanks! I added changes to e0f94d2 to reflect your comment. I think that they add helpful clarity!

FIPS/fip-draft_transient_storage.md
snissn: c7d03e3

FIPS/fip-draft_transient_storage.md
snissn: To add further clarity on transient storage and its integration with Solidity:

The implementation of transient storage in the Ethereum Virtual Machine (EVM), as introduced in EIP-1153, relies on two new opcodes: `TLOAD` and `TSTORE`. These opcodes enable direct interaction with transient storage slots, which are scoped to the transaction and the contract. Solidity supports these opcodes starting from version `v0.8.24`, introduced with the Cancun upgrade.

However, it’s important to note that Solidity's support for transient storage remains assembly-based. Developers must explicitly use inline assembly to interact with transient storage, as high-level abstractions (like mappings or arrays) are not currently available for transient data. For example, accessing transient storage in Solidity might look like this:

```solidity
uint256 val;
assembly {
    val := tload(slot) // Load value from transient storage at `slot`
    tstore(slot, 42)   // Store value 42 in transient storage at `slot`
}
```

A good example of how transient storage is currently being used can be found in [[Uniswap V4’s Lock implementation](https://github.com/Uniswap/v4-core/blob/b619b6718e31aa5b4fa0286520c455ceb950276d/src/libraries/Lock.sol)](https://github.com/Uniswap/v4-core/blob/b619b6718e31aa5b4fa0286520c455ceb950276d/src/libraries/Lock.sol). This demonstrates a lock implemented in assembly using `TLOAD` and `TSTORE`.

This means that while transient storage functionality is available in newer Solidity versions, its usage is not deeply integrated into the language's high-level constructs. Contracts compiled with older versions of Solidity (prior to `v0.8.24`) will remain unaffected unless they are explicitly updated to use `TLOAD` and `TSTORE`.

This reinforces the point that the introduction of transient storage does not break compatibility for existing contracts. The feature is optional and only affects contracts that are explicitly written to use it with newer Solidity versions.

snissn: To provide additional context, the notion of transient storage compatibility being closely tied to specific Solidity versions, such as `v0.8.24`, may not fully capture the broader picture. While Solidity added support for the `TLOAD` and `TSTORE` opcodes in `v0.8.24` following the Cancun upgrade, these opcodes are fundamentally EVM-level features. Developers can interact with them directly at the opcode level, independent of the Solidity version, as long as the underlying EVM implementation supports these opcodes.

For instance, transient storage can be accessed in assembly using the raw opcodes for `TLOAD` (`0x5c`) and `TSTORE` (`0x5d`), even with Solidity versions earlier than `v0.8.24`:

```solidity
uint256 val;
assembly {
    // Load value from transient storage at `slot`
    mstore(0x00, slot)      // Prepare the slot address
    pop(staticcall(gas(), 0x5c, 0x00, 0x20, 0x00, 0x20)) // Raw opcode for TLOAD
    val := mload(0x00)      // Retrieve the value from memory

    // Store value 42 into transient storage at `slot`
    mstore(0x00, slot)      // Prepare the slot address
    mstore(0x20, 42)        // Prepare the value
    pop(staticcall(gas(), 0x5d, 0x00, 0x40, 0x00, 0x00)) // Raw opcode for TSTORE
}
```

This shows that while Solidity `v0.8.24` introduces a more streamlined way to use `TLOAD` and `TSTORE` opcodes, the underlying functionality is an EVM feature and not strictly tied to the compiler version. This means developers could theoretically use transient storage with older Solidity versions by working directly with these opcodes, as long as the runtime environment supports them.

That said, this approach comes with some challenges:
- **Readability and Maintenance**: Using raw opcodes makes code harder to read and maintain compared to the cleaner `TLOAD` and `TSTORE` mnemonics introduced in Solidity.
- **Version Interpretation**: Associating transient storage solely with Solidity `v0.8.24` might create confusion, as its implementation is ultimately dependent on the EVM rather than the compiler version.
- **Risk of Errors**: Directly handling opcodes increases the risk of bugs and introduces additional complexity, particularly if the EVM evolves in the future.

To summarize, while Solidity `v0.8.24` provides high-level support for transient storage, the functionality itself exists at the EVM level. Discussions of compatibility should focus on the EVM’s support for `TLOAD` and `TSTORE` opcodes rather than being solely tied to specific Solidity compiler versions.

snissn: To add to the discussion, the backwards compatibility section as written is accurate—contracts compiled with older Solidity versions (pre-`v0.8.24`) will remain unaffected, as they do not include `TLOAD` and `TSTORE`. These opcodes, introduced in the EVM with the Cancun upgrade, are supported in Solidity starting with `v0.8.24`. However, it’s a fun idea to explore that the compatibility of transient storage is fundamentally an EVM-level feature and not strictly tied to a specific Solidity version.

For example, developers can interact with transient storage directly using the raw opcodes (`TLOAD = 0x5c` and `TSTORE = 0x5d`) in assembly, regardless of the Solidity version used, as long as the underlying EVM supports these opcodes. The cancun Ethereum hardfork definitely itroduces the two opcodes associated with transient storage, but these can be accessed by developers working with different versions of Solidity:

```solidity
uint256 val;
assembly {
    // Load value from transient storage at `slot`
    mstore(0x00, slot)      // Prepare the slot address
    pop(staticcall(gas(), 0x5c, 0x00, 0x20, 0x00, 0x20)) // Raw opcode for TLOAD
    val := mload(0x00)      // Retrieve the value from memory

    // Store value 42 into transient storage at `slot`
    mstore(0x00, slot)      // Prepare the slot address
    mstore(0x20, 42)        // Prepare the value
    pop(staticcall(gas(), 0x5d, 0x00, 0x40, 0x00, 0x00)) // Raw opcode for TSTORE
}
```

This flexibility means that while Solidity `v0.8.24` introduces a more user-friendly way to access transient storage, it is not strictly required for its use. The feature itself is native to the EVM, so compatibility depends more on the runtime environment supporting these opcodes than on the Solidity compiler version.

In summary, the backwards compatibility section captures the key points well. For developers working with transient storage, the association with Solidity `v0.8.24` is primarily about convenience, as the opcodes themselves are an EVM feature accessible in any version of Solidity that supports inline assembly.

FIPS/fip-draft_transient_storage.md
snissn: 12556d9 and f576eab

FIPS/fip-draft_transient_storage.md
snissn: f576eab I think this adds the requirement in a way that accommodates your comment

snissn:
Thanks for the comments @rvagg ! Your feedback helps clarify many important aspects of this FIP and leaves to a more robust proposal overall! I got a little "nerd sniped" by the solidity version relationship to this FIP, and hope I didn't get too nuanced and abstract! Please share any additional comments and feedback to ensure all of your requests and ideas have been addressed and accommodated! Thanks again!

FIPS/fip-draft_transient_storage.md
rvagg: hah, that's all quite interesting! thanks for the braindump. I'm not sure how much of that to suggest putting into the doc, maybe at least the thing about it being an `assembly {}` feature even in newer versions of Solidity—the important point being that it's _explicitly_ opt-in, it's not something the compiler is going to pop out for you when you compile because it makes sense for something you're doing. Because that's the concern with compatibility - will a newer compiler produce bytecode that it thinks is optimal because there's a new feature but we're not able to run it. But in this case, from what you're saying, that's not even a possibility. That seems like something worth noting.

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,105 @@
...
+   - Verify that nested contracts have independent transient storage spaces can read and write independently.
+   - Verify that memory remains accessible and stable after contract reentry.
+
+---
+
+## Security Considerations
+Transient storage introduces minimal additional risk compared to existing state storage. Lifecycle validation ensures storage is inaccessible outside the originating transaction. Security measures include:
+- Preventing out-of-bounds memory access.
+- Ensuring transient storage clears properly after a transaction ends.
+- Ensuring nested contracts do not have access to eachother's memory spaces.
```
jsoares, line 82:
```suggestion
- Ensuring nested contracts do not have access to each other's memory spaces.
```
I'm all for archaic and nonstandard spellings, but maybe not here :)

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,105 @@
...
+
+---
+
+## Incentive Considerations
+Transient storage supports the management of temporary, transaction-scoped data and enables features outlined in **EIP-1153**, such as reentrancy locks, temporary approvals, on-chain address computation, and enhanced proxy call metadata handling. These capabilities align with Filecoin’s goal of enabling scalable and reliable decentralized applications.
+
+By implementing transient storage, the FEVM ensures compatibility with Ethereum’s tooling and supports developers accustomed to this feature. It also facilitates seamless integration of existing projects utilizing EIP-1153, enriching the Filecoin ecosystem and expanding its appeal to the broader Ethereum developer community.
```
jsoares, line 89: These read like product, rather than incentive, considerations. It's unclear if there are significant incentive considerations at play; I'm not convinced there are.

jsoares (approved):
Nothing substantive to add to @rvagg's excellent feedback. Provided the outstanding comments are addressed, this looks good to go.

jsoares:
Assigning FIP-0097. Please update filename, front matter, and readme.

FIPS/fip-draft_transient_storage.md
snissn: thanks! The github UI is currently unwilling to commit the change but I added it here: 79d0086

FIPS/fip-draft_transient_storage.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Change Motivation
+Transient storage offers developers temporary, transaction-scoped storage for managing intermediate states. One key benefit of this feature is its ability to improve the implementation of reentrancy locks, enhancing security by mitigating risks associated with multiple calls to a function within sub-transactions. This feature was introduced on Ethereum mainnet in March 2024 as part of the Cancun (Decun) upgrade, making adopting this FIP important to align the FEVM with Ethereum’s evolving tooling and Solidity’s modern features. By implementing transient storage, the FEVM ensures a seamless developer experience while supporting advanced contract use cases and secure computing.
```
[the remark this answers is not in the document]
snissn: thanks! 06ced1d

snissn:
@luckyparadise thanks for your question! Reviewing the writing I see that the word usage of ephemeral can be creating confusion witihout adding much to the document. 

Transient storage is the technical term that has technical meaning. My intention behind my usage of ephemeral in this FIP write up was to be a descriptive term that doesn't insist on a precise technical purpose per se. I think it's an attempt I made to create a category for memory data and transient data that would be separate from storage data..

 I think I can just delete the three instances of the word ephemeral in the FIP write up and it removes ambiguity and nothing is lost. Changed here b69533d

FIPS/fip-draft_transient_storage.md
snissn: Thanks @rvagg what you are suggesting does sound helpful! I'll try to synthesize that into an addition this week.

FIPS/fip-0097.md
```diff
@@ -0,0 +1,105 @@
...
+---
+
+## Design Rationale
+
+The design adheres to the intent of EIP-1153 while adapting to Filecoin's architecture. The use of lifecycle validation ensures transient storage behaves as expected within the scope of a single transaction. This approach balances compatibility with Ethereum contracts and simplifies implementation within the existing FEVM architecture. 
+
+Alternative designs, such as purely in-memory storage, were considered but deemed impractical due to technical implementation difficulties.
+
+## Backwards Compatibility
+The addition of transient storage is fully backward-compatible. Existing contracts remain unaffected unless they utilize `TLOAD` and `TSTORE`. Contracts compiled with older Solidity versions will continue to execute without changes.
```
rvagg, line 60:
```suggestion
The addition of transient storage is fully backward-compatible. Existing contracts remain unaffected unless they utilize `TLOAD` and `TSTORE`. Contracts compiled with older Solidity versions will continue to execute without changes.

At the time of writing, support for `TLOAD` and `TSTORE` in current versions of Solidity is only available using inline-assembly. Thus, it is an explicitly opt-in feature, rather than a feature that contract authors may unknowngly be enabling when bumping compiler version and EVM target.
```

FIPS/fip-draft_transient_storage.md
rvagg: left a suggestion for wording

FIPS/fip-0097.md
snissn: Thanks @rvagg your addition is much appreciated!

FIPS/fip-draft_transient_storage.md
snissn: Thanks! I have revised this section here: ddf9501

rvagg (approved):

anorth (approved):
Editorial 👍

PR #1117, opened 2025-02-10
Removal of the ProveCommitAggregate method from the miner actor

rvagg:
**Discussion**: https://github.com/filecoin-project/FIPs/discussions/1116

## Simple Summary
This FIP proposes the removal of the `ProveCommitAggregate` method (Method 26) from the miner actor as it is no longer actively used due to the introduction of `ProveCommitSectors3`, which offers the same functionality in a more flexible manner.

## Abstract
The `ProveCommitAggregate` method was implemented for batch processing of pre-committed sectors using aggregate proofs. However, the subsequent introduction of `ProveCommitSectors3` allows greater flexibility and efficiency in handling this task. Given that `ProveCommitAggregate` has been deprecated in practical use, it adds unnecessary complexity to the system and generates technical debt. This FIP proposes to remove this method from the miner actor to simplify the protocol.


jsoares (approved):

rvagg:
I guess since I'm an editor and there are a couple of approvals now that I'll assign a number to this so I'm claiming **0101** and have updated the PR accordingly.

PR #1123, opened 2025-02-14
Move FIP-0100 to Last Call

momack2:
This is a proposal to move FIP-0100 to last call in order to make the nv25 finalization and code freeze timeline. 
Last call would last for the normal 2-week period from Feb 14-Feb 28th.


TippyFlitsUK (approved):
🚀

dannyob (approved):
lgtm!

PR #1128, opened 2025-02-20
Address changes to FIP 98 while in last call

Schwartz10:
Updates to FIP 98 while in last call:

- Added 3 special cases to consider when computing termination fees: when termination fee < fault fee, when a sector's age is < 140 days, and when a sector's termination fee is less than some minimum % of its initial pledge
- Added more test cases to address the special cases, as well as a test case for computing aggregate fault fees across multiple sectors
- Define new built-in actor methods to help callers (including FEVM callers by exporting built-in actor methods) to compute termination fees more easily

### DIFF
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -126,7 +126,7 @@ In addition to the revised termination fee formulas, a few built-in actor method
 |--------------|-------------------|-------------|
 | `miner_power` | `miner_power() -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
 | `miner_initial_pledge` | `miner_initial_pledge() -> TokenAmount` | Returns the miner's total initial pledge amount |
-| `termination_fee_percentage` | `termination_fee_percentage() -> f64` | Returns the network's termination fee percentage (8.5%) |
+| `termination_fee_percentage` | `termination_fee_percentage() -> (num u32, denom u32)` | Returns the network's termination fee percentage  (8.5%) |
 | `max_termination_fee` | `max_termination_fee(initial_pledge: TokenAmount, power: QAPower) -> TokenAmount` | Returns the maximum termination fee calculation for a given initial pledge and power amount |
 
 ## Design Rationale
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -198,7 +198,7 @@ It is important that Storage Providers are incentivized to honor their commitmen
 
 The termination fee for sectors committed a long time ago will immediately and significantly decrease, but only to match the fee:pledge ratio of new sectors. This will increase the incentive for some SPs with high-pledge sectors to terminate and re-onboard to gain more power for the same amount of pledge (but more hardware).
 
-In contrast to the current system, the termination penalty for a sector will be constant regardless of its age. Currently the termination penalty starts low and increases with age up to a max. This presents a change in the incentive structure for short-term-view SPs.
+In contrast to the current system, the termination penalty for a sector will be constant regardless of its age (after the 140 day linear ramp-up). Currently the termination penalty starts low and increases with age up to a max. This presents a change in the incentive structure for short-term-view SPs.
 
 ## Product Considerations
 
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -25,7 +25,7 @@ The termination fee for any given sector is calculated as a fixed percentage of 
 
 Today, it is overly sophisticated and computationally expensive to compute termination fees for miner actors. As a result, DeFi applications that leverage miner actors must make major security and/or performance sacrifices in order to operate.
 
-This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 8.5%. There are two special cases to consider where the termination fee must be tweaked to maintain current network conditions, which are addressed in the technical specification and design rationale.
+This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 8.5%. There are three special cases to consider where the termination fee must be tweaked to maintain current network conditions, which are addressed in the technical specification and design rationale.
 
 As a result, DeFi applications on Filecoin can operate with significantly better performance, UX, and economic security. Additionally, Filecoin economics and code implementations will be significantly simplified, as well as state bloat removed.
 
@@ -39,14 +39,13 @@ Today, it is: (1) computationally expensive, (2) economically sophisticated, and
 
 The two primary motivations to this FIP proposal are:
 
-1. Enabling a more efficient termination penalty calculation such that termination penalties can be computed (or returned by an FEVM precompile) in an FEVM runtime. This is important because it enables FEVM protocols that use Miner Actors as collateral to operate with better economic security because they can precisely estimate the value of Miner Actor collateral on-chain. 
-2. Changing the formula for calculating termination penalties to be simpler to compute. This is important because it allows both Storage Providers and FEVM actors that use Miner Actors as collateral to easily predict their economic riskiness. 
+1. Enabling a more efficient termination penalty calculation such that termination penalties can be computed (or returned by an FEVM precompile) in an FEVM runtime. This is important because it enables FEVM protocols that use Miner Actors as collateral to operate with better economic security because they can precisely estimate the value of Miner Actor collateral on-chain.
+2. Changing the formula for calculating termination penalties to be simpler to compute. This is important because it allows both Storage Providers and FEVM actors that use Miner Actors as collateral to easily predict their economic riskiness.
 
 Additionally, this FIP intends to maintain the original motivations behind termination fees:
 
 1. Create a more resilient storage network for storage clients, such that if an SP stores a client's data, the client isn't just suddenly left out to dry and their data lost.
 2. Network stability - disincentivize large power swings caused by rapid on/off-boarding of power and pledge.
-
 
 ## Specification
 
@@ -54,12 +53,12 @@ Additionally, this FIP intends to maintain the original motivations behind termi
 
 Three new constants should be created, and an existing one will be reused. Each is fixed for the whole network and configurable by future governance:
 
-| Constant | Description | Amount |
-|----------|-------------|--------|
-| `TERM_FEE_PLEDGE_MULTIPLE` | This constant will be used to compute termination fees in the base case by multiplying against initial pledge. | 0.085x initial pledge |
-| `TERM_FEE_MAX_FAULT_FEE_MULTIPLE` | This constant will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector. | 1.05x |
-| `TERM_FEE_MIN_PLEDGE_MULTIPLE` | This constant will ensure the termination fee for young sectors is not arbitrarily low. | 0.02x initial pledge |
-| `TERMINATION_LIFETIME_CAP` | _(Existing constant)_ This constant maintains the linear increase in termination fee for young sectors. | 140 days |
+| Constant                          | Description                                                                                                                                 | Amount                |
+| --------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- | --------------------- |
+| `TERM_FEE_PLEDGE_MULTIPLE`        | This constant will be used to compute termination fees in the base case by multiplying against initial pledge.                              | 0.085x initial pledge |
+| `TERM_FEE_MAX_FAULT_FEE_MULTIPLE` | This constant will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector. | 1.05x                 |
+| `TERM_FEE_MIN_PLEDGE_MULTIPLE`    | This constant will ensure the termination fee for young sectors is not arbitrarily low.                                                     | 0.02x initial pledge  |
+| `TERMINATION_LIFETIME_CAP`        | _(Existing constant)_ This constant maintains the linear increase in termination fee for young sectors.                                     | 140 days              |
 
 Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
 
@@ -93,16 +92,16 @@ termination_fee = max(base_termination_fee, minimum_fee)
 
 This ensures that:
 
-* The fee is never less than 2% of the initial pledge
-* The fee is never less than 105% of the fault fee
-* For sectors younger than 140 days, the fee scales linearly with age (subject to the bounds)
-* For sectors older than 140 days, the fee is 8.5% of the initial pledge (subject to the bounds)
+- The fee is never less than 2% of the initial pledge
+- The fee is never less than 105% of the fault fee
+- For sectors younger than 140 days, the fee scales linearly with age (subject to the bounds)
+- For sectors older than 140 days, the fee is 8.5% of the initial pledge (subject to the bounds)
 
 The complete formula in one expression is:
 
 ```
 termination fee = max(
-  MIN_TERMINATION_FEE * initial_pledge, 
+  MIN_TERMINATION_FEE * initial_pledge,
   max(
     TERM_PENALTY_PLEDGE_PERCENTAGE * initial_pledge * min(1, sector_age_days / TERMINATION_LIFETIME_CAP),
     FAULT_FEE_MULTIPLE * sector_fault_fee
@@ -127,11 +126,7 @@ In addition to the revised termination fee formulas, a few built-in actor method
 |--------------|-------------------|-------------|
 | `miner_power` | `miner_power(miner_id: ActorID) -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
 | `miner_initial_pledge` | `miner_initial_pledge(miner_id: ActorID) -> TokenAmount` | Returns the miner's total initial pledge amount |
-| `network_projection` | `network_projection(projection_period_start: ChainEpoch, projection_period_end: ChainEpoch) -> (Power, TokenAmount)` | Returns the network's projected power and reward for the given projection period |
 | `termination_fee_percentage` | `termination_fee_percentage() -> f64` | Returns the network's termination fee percentage (8.5%) |
-  
-(the below 2 methods could also be built as a FEVM native oracle contract instead of as a built-in actor method with exports to access from FEVM):
-| `single_fault_fee` | `single_fault_fee(qa_power: QAPower) -> TokenAmount` | Returns the fault fee calculation for a given amount of QA power |
 | `max_termination_fee` | `max_termination_fee(initial_pledge: TokenAmount, power: QAPower) -> TokenAmount` | Returns the maximum termination fee calculation for a given initial pledge and power amount |
 
 ## Design Rationale
@@ -142,7 +137,7 @@ There are a few different approaches one could take to accomplish the same end r
 
 1. Use a "multiple of daily rewards" approach, which computes a termination fee based on a % or multiple of expected daily rewards for a period of time based on the current block rewards
    - Problems:
-     - Less predictable - in order to estimate the collateral value of a Miner Actor, the protocol needs to be able to estimate what the future daily block reward will be for the network. If the network grows quickly and unexpectedly, this could dangerously impact a DeFi protocol as the daily block rewards may increase due to increasing baseline, thus increasing the termination penalty and decreasing the collateral value for a given Miner Actor. 
+     - Less predictable - in order to estimate the collateral value of a Miner Actor, the protocol needs to be able to estimate what the future daily block reward will be for the network. If the network grows quickly and unexpectedly, this could dangerously impact a DeFi protocol as the daily block rewards may increase due to increasing baseline, thus increasing the termination penalty and decreasing the collateral value for a given Miner Actor.
 2. Optimize data structures based on the current methodology to make an aggregate termination sector method feasible in constant time lookup
    - Problems:
      - Unsure if this is technically feasible, and if so, most likely a lot of sophistication and ugly accounting
@@ -150,10 +145,10 @@ There are a few different approaches one could take to accomplish the same end r
 
 The %-of-pledge method is:
 
- 1. simple to calculate
- 2. predictable
- 3. fixed while the sector's pledge is constant
- 4. will not decay towards zero as block rewards decline, thus more effectively protecting against churn
+1.  simple to calculate
+2.  predictable
+3.  fixed while the sector's pledge is constant
+4.  will not decay towards zero as block rewards decline, thus more effectively protecting against churn
 
 Additionally, there are two other considerations to make with respect to the design rational:
 
@@ -173,6 +168,7 @@ This FIP requires a network upgrade because it intends to change built-in actor 
 <!--Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
 
 - When terminating a single sector:
+
   - Not considering fault fees, for a sector where its age >= `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_FEE_PLEDGE_MULTIPLE * initial pledge`
   - Not considering fault fees, for a sector where its age < `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_FEE_PLEDGE_MULTIPLE * of initial pledge * sector age in days / TERMINATION_LIFETIME_CAP`
   - Considering fault fees, for a sector with a termination fee that is less than the associated sector's fault fee, termination fee should equal `TERM_FEE_MAX_FAULT_FEE_MULTIPLE * fault fee`
@@ -192,7 +188,7 @@ The security concern with introducing this FIP is that Storage Providers will no
 3. It seems appropriate to charge an exit fee for breaking a commitment to the network, but we also do not want to charge a capture fee. Storage Providers who do not wish to use Filecoin anymore shouldn't have prohibitively high expenses for leaving the network early - ultimately for the network, burning some of this SPs tokens through sector terminations is better than 18 months of selling.
 4. As Filecoin matures, the average Storage Provider will likely be accepting paid deals in some form according to an SLA in addition to the SLA guaranteed by PoRep. These additional SLAs can/will enforce their own termination clauses, which should provide adequate motivation and verification to achieve a good experience for the average storage client.
 5. The 140 day linear termination fee increase is kept in tact to ensure no existing loan / collateral position in the Filecoin capital markets flips undercollateralized.
-6. There can be theoretical security issues when the termination fee is less than the fault fee for the same sector, however, the proposal here does not allow termination fees to drop below fault fees. 
+6. There can be theoretical security issues when the termination fee is less than the fault fee for the same sector, however, the proposal here does not allow termination fees to drop below fault fees.
 
 ## Incentive Considerations
 
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -196,7 +196,7 @@ The security concern with introducing this FIP is that Storage Providers will no
 
 It is important that Storage Providers are incentivized to honor their commitments to the network - this prevents major network volatility. It makes sense to maintain a disincentive for breaking commitments to the network, but this FIP doesn't change the incentive to provide useful storage to the network, it makes it somewhat easier.
 
-The termination fee for sectors committed a long time ago will immediately and significantly decrease, but only to match the fee:pledge ratio of new sectors. This will increase the incentive for some SPs with high-pledge sectors to terminate and re-onboard to gain more power for the same amount of pledge (but more hardware).
+The termination fee for sectors committed a long time ago will immediately and significantly decrease, but only to match the fee:pledge ratio of new sectors. This will increase the incentive for some SPs with high-pledge sectors to terminate and re-onboard to gain more power for the same amount of pledge (but more hardware). See [this discussion comment](https://github.com/filecoin-project/FIPs/discussions/1036#discussioncomment-10215706) for the latest data on live sector fee:pledge data from August, 2024.
 
 In contrast to the current system, the termination penalty for a sector will be constant regardless of its age (after the 140 day linear ramp-up). Currently the termination penalty starts low and increases with age up to a max. This presents a change in the incentive structure for short-term-view SPs.
 
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -64,16 +64,55 @@ Three new constants should be created, and an existing one will be reused. Each 
 Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
 
 1. Take the sector's `initial_pledge` `TokenAmount` as an argument
-2. Calculate a `termination_fee` value of `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
-3. If the sector's age is less than `TERMINATION_LIFETIME_CAP` days, multiply the `termination_fee` by `sector age in days / TERMINATION_LIFETIME_CAP`
-4. Compute the per sector fault fee using the method `pledge_penalty_for_continued_fault` [link](https://github.com/filecoin-project/builtin-actors/blob/8a757341d4658461e70161d1f20f9b02991e8a5f/actors/miner/src/monies.rs#L148), if the computed termination fee is less or equal to fault fee, return `FAULT_FEE_MULTIPLE * fault_fee`
-5. If `termination_fee` is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
+1. Calculate the **base termination fee**:
 
-Note that the `pledge_penalty_for_continued_fault` method is not refactored in this FIP, but it must be verified to work for aggregate power numbers above the possible QA power for any single sector. For instance, an aggregate of 10 sectors' power should return the same end result as summing the `pledge_penalty_for_continued_fault` of each sector individually. This allows a caller to compute accurate maximum termination fees for any given aggregate power number (without looping through each sector).
+```
+base_termination_fee = initial_pledge * TERM_FEE_PLEDGE_MULTIPLE  // (0.085 * initial_pledge)
+```
 
-The formula written out is:
+2. Apply the **age adjustment** for young sectors:
 
-`termination fee = max(MIN_TERMINATION_FEE, max{TERM_PENALTY_PLEDGE_PERCENTAGE * initial_pledge, FAULT_FEE_MULTIPLE * sector_fault_fee} * min(1, sector_age_days / TERMINATION_LIFETIME_CAP))`
+```
+age_factor = min(1, sector_age_days / TERMINATION_LIFETIME_CAP)
+age_adjusted_fee = base_termination_fee * age_factor
+```
+
+3. Calculate the **minimum allowed fee**:
+
+```
+minimum_fee = initial_pledge * TERM_FEE_MIN_PLEDGE_MULTIPLE  // (0.02 * initial_pledge)
+```
+
+4. Calculate the **maximum allowed fee** (based on fault fee, using the [existing `pledge_penalty_for_continued_fault` method]((https://github.com/filecoin-project/builtin-actors/blob/8a757341d4658461e70161d1f20f9b02991e8a5f/actors/miner/src/monies.rs#L148))):
+
+```
+fault_fee = pledge_penalty_for_continued_fault(sector_power)
+maximum_fee = fault_fee * TERM_FEE_MAX_FAULT_FEE_MULTIPLE  // (1.05 * fault_fee)
+```
+
+5. Apply the fee bounds:
+
+```
+termination_fee = max(minimum_fee, min(age_adjusted_fee, maximum_fee))
+```
+
+This ensures that:
+
+* The fee is never less than 2% of the initial pledge
+* The fee is never more than 105% of the fault fee
+* For sectors younger than 140 days, the fee scales linearly with age
+* For sectors older than 140 days, the fee is 8.5% of the initial pledge (subject to the bounds)
+
+The complete formula in one expression is:
+
+```
+termination_fee = max(
+    initial_pledge * TERM_FEE_MIN_PLEDGE_MULTIPLE,
+    min(
+        initial_pledge * TERM_FEE_PLEDGE_MULTIPLE * min(1, sector_age_days / TERMINATION_LIFETIME_CAP),
+        fault_fee * TERM_FEE_MAX_FAULT_FEE_MULTIPLE
+    )
+)
 
 This change would immediately apply to all sectors, including historical sectors.
 
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -135,7 +135,7 @@ In addition to the revised termination fee formulas, a few built-in actor method
 
 There are a few different approaches one could take to accomplish the same end result:
 
-1. Use a "multiple of daily rewards" approach, which computes a termination fee based on a % or multiple of expected daily rewards for a period of time based on the current block rewards
+1. Use a "multiple of daily rewards" approach, which computes a termination fee based on a % or multiple of expected daily rewards for a period of time, based on the current block rewards
    - Problems:
      - Less predictable - in order to estimate the collateral value of a Miner Actor, the protocol needs to be able to estimate what the future daily block reward will be for the network. If the network grows quickly and unexpectedly, this could dangerously impact a DeFi protocol as the daily block rewards may increase due to increasing baseline, thus increasing the termination penalty and decreasing the collateral value for a given Miner Actor.
 2. Optimize data structures based on the current methodology to make an aggregate termination sector method feasible in constant time lookup
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -60,10 +60,10 @@ Three new constant variables should be created, each variable is fixed for the w
 Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
 
 1. Take the sector's `initial_pledge` `TokenAmount` as an argument
-2. Return `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
-3. If the sector's age is less than 140 days, multiply the result computed above (step 2) by a fraction of (`sector age in days / 140`)
-4. If the computed termination fee is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee` 
-5. If the computed termination fee is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
+2. Calculate a `termination_fee` value of `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+3. If the sector's age is less than 140 days, multiply the `termination_fee` by `sector age in days / 140`
+4. If `termination_fee` is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee`
+5. If `termination_fee` is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
 
 The formula written out is:
 
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -152,7 +152,7 @@ The %-of-pledge method is:
 
 Additionally, there are two other considerations to make with respect to the design rationale:
 
-1. According to certain network participants, some Filecoin lenders provide loans to Storage Providers based on lower termination fees for sectors that are younger than 140 days old. The proposed design maintains a linear increasing termination fee over the first 140 days of the sectors life
+1. According to certain network participants, some Filecoin lenders provide loans to Storage Providers based on lower termination fees for sectors that are younger than 140 days old. The proposed design maintains a linear increasing termination fee over the first 140 days of the sector's life
 2. There are theoretical economic security issues when `termination fee < fault fee` for a sector (see [here](https://github.com/filecoin-project/FIPs/discussions/1036#discussioncomment-11984562) for more info). The proposed design ensures this security issue does not happen by forcing the termination fee to be greater than the fault fee if that edge case ever occurs.
 
 ## Backwards Compatibility
--- a/FIPS/fip-0098.md
+++ b/FIPS/fip-0098.md
@@ -17,7 +17,7 @@ created: 2024-09-26
 
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
 
-The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector. The proposed penalty as a percentage of initial pledge is 8.5%.
+The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector, except in 2 special cases. The proposed penalty as a percentage of initial pledge is 8.5%. 
 
 ## Abstract
 
@@ -25,7 +25,7 @@ The termination fee for any given sector is calculated as a fixed percentage of 
 
 Today, it is overly sophisticated and computationally expensive to compute termination fees for miner actors. As a result, DeFi applications that leverage miner actors must make major security and/or performance sacrifices in order to operate.
 
-This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 8.5%.
+This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 8.5%. There are two special cases to consider where the termination fee must be tweaked to maintain current network conditions, which are addressed in the technical specification and design rationale.
 
 As a result, DeFi applications on Filecoin can operate with significantly better performance, UX, and economic security. Additionally, Filecoin economics and code implementations will be significantly simplified, as well as state bloat removed.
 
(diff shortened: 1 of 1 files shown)


FIPS/fip-0098.md
```diff
@@ -17,15 +17,15 @@ created: 2024-09-26
 
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
 
-The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector. The proposed penalty as a percentage of initial pledge is 8.5%.
+The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector, except in 2 special cases. The proposed penalty as a percentage of initial pledge is 8.5%.
```
rvagg, line 20:
```suggestion
The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector, except in 2 special cases. The proposed penalty as a percentage of initial pledge is 8.5%.
```

very minor nit

FIPS/fip-0098.md
```diff
@@ -52,12 +52,22 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
+2. `FAULT_FEE_MULTIPLE` - this variable will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector.
+3. `MIN_TERMINATION_FEE` - this variable will ensure the termination fee for young sectors is not arbitrarily low.
 
 Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
 
 1. Take the sector's `initial_pledge` `TokenAmount` as an argument
 2. Return `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+3. If the sector's age is less than 140 days, multiply the result computed above (step 2) by a fraction of (`sector age in days / 140`)
+4. If the computed termination fee is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee` 
+5. If the computed termination fee is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
```
rvagg, line 66:
```suggestion
2. Calculate a `termination_fee` value of `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
3. If the sector's age is less than 140 days, multiply the `termination_fee` by `sector age in days / 140`
4. If `termination_fee` is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee`
5. If `termination_fee` is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
```
?

FIPS/fip-0098.md
```diff
@@ -52,12 +52,22 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
+1. `TERM_PENALTY_PLEDGE_PERCENTAGE` - this variable will be used to compute termination fees in the base case by multiplying against initial pledge.
+2. `FAULT_FEE_MULTIPLE` - this variable will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector.
+3. `MIN_TERMINATION_FEE` - this variable will ensure the termination fee for young sectors is not arbitrarily low.
 
 Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
 
 1. Take the sector's `initial_pledge` `TokenAmount` as an argument
 2. Return `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+3. If the sector's age is less than 140 days, multiply the result computed above (step 2) by a fraction of (`sector age in days / 140`)
+4. If the computed termination fee is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee`
```
rvagg, line 65: Looking through the code I'm unsure what this is derived from, I see `pledge_penalty_for_continued_fault` but that seems to be `FF(t), but we want it total FF, for .. the remainder of the term, so FF(t)*term or something? This doesn't necessarily need clarification in the FIP text because maybe it's clear to others precisely what this value is, I'm just not grokking it in the code so wonder if this is obvious to others.

FIPS/fip-0098.md
Schwartz10: Yeah thank you for making this clearer

FIPS/fip-0098.md
irenegia: @rvagg 
what do you meant with " total FF, for .. the remainder of the term"?

FF(t) is the fee that is paid when a faulty sector is not recovered in time for the next widowPoSt. 
SPs pay one FF(t) for each windowPoSt they miss for that sectors (up to 14 post, after that he sector gets terminated).
Iirc, you pay at deadline and you pay 100% of ite at each deadline, so there is not "remainder to pay".
There is a "t" in the terminology because FF(t) value can change over time because it depends on the per sector block reward. 

Given this, i think bullet pint 3 in the fip text is correct. Maybe we can simply make it more explicit in the following way:

3. Compute the per sector `fault fee` using the method `pledge_penalty_for_continued_fault` ([link](https://github.com/filecoin-project/builtin-actors/blob/8a757341d4658461e70161d1f20f9b02991e8a5f/actors/miner/src/monies.rs#L148)), if the computed termination fee is less or equal to `fault fee`, return `FAULT_FEE_MULTIPLE * fault_fee`

LesnyRumcajs:
I believe the test cases section should be expanded as well.

FIPS/fip-0098.md
```diff
@@ -52,12 +52,22 @@ Additionally, this FIP intends to maintain the original motivations behind termi
 
 <!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any of the current Filecoin implementations. -->
 
-A new constant variable called `TERM_PENALTY_PLEDGE_PERCENTAGE` should be created, which is fixed for the whole network and configurable by future governance.
+Three new constant variables should be created, each variable is fixed for the whole network and configurable by future governance:
+1. `TERM_PENALTY_PLEDGE_PERCENTAGE` - this variable will be used to compute termination fees in the base case by multiplying against initial pledge.
+2. `FAULT_FEE_MULTIPLE` - this variable will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector.
+3. `MIN_TERMINATION_FEE` - this variable will ensure the termination fee for young sectors is not arbitrarily low.
```
LesnyRumcajs, line 58: Personally, I'd use `constant` instead of `constant variable`.

Also, what are the proposed values of those constants? `TERM_PENALTY_PLEDGE_PERCENTAGE` is 8.5%, what about others? Should they differ between different networks?

FIPS/fip-0098.md
```diff
@@ -17,15 +17,15 @@ created: 2024-09-26
 
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
 
-The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector. The proposed penalty as a percentage of initial pledge is 8.5%.
+The termination fee for any given sector is calculated as a fixed percentage of the initial pledge held by that sector, except in 2 special cases. The proposed penalty as a percentage of initial pledge is 8.5%.
```
LesnyRumcajs, line 20: Isn't it 3 cases?
1. `termination fee < fault fee`
2. 140 days ramp
3. clamping to `MIN_TERMINATION_FEE`

FIPS/fip-0098.md
```diff
@@ -52,12 +52,22 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
+1. `TERM_PENALTY_PLEDGE_PERCENTAGE` - this variable will be used to compute termination fees in the base case by multiplying against initial pledge.
+2. `FAULT_FEE_MULTIPLE` - this variable will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector.
+3. `MIN_TERMINATION_FEE` - this variable will ensure the termination fee for young sectors is not arbitrarily low.
 
 Refactor the [`pledge_penalty_for_termination`](https://github.com/filecoin-project/builtin-actors/blob/12d9af8a00d0909598c67e1a18dc1577e0833137/actors/miner/src/monies.rs#L179) method to:
 
 1. Take the sector's `initial_pledge` `TokenAmount` as an argument
-2. Return `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+2. Calculate a `termination_fee` value of `initial_pledge * TERM_PENALTY_PLEDGE_PERCENTAGE`
+3. If the sector's age is less than 140 days, multiply the `termination_fee` by `sector age in days / 140`
```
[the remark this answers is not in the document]
rvagg: yes

FIPS/fip-0098.md
```diff
@@ -52,12 +52,22 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
+4. If `termination_fee` is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee`
```
rvagg, line 65:
```suggestion
4. If `termination_fee` is less than the fault fee computed for the same sector, return `FAULT_FEE_MULTIPLE * fault_fee`, where `fault_fee` is calculated per the existing [`pledge_penalty_for_continued_fault()`](https://github.com/filecoin-project/builtin-actors/blob/8a757341d4658461e70161d1f20f9b02991e8a5f/actors/miner/src/monies.rs#L148-L159) function.
```

FIPS/fip-0098.md
rvagg, line 58:
```suggestion
Three new constants should be created, each is fixed for the whole network and configurable by future governance:

1. `TERM_PENALTY_PLEDGE_PERCENTAGE = 8.5`: used to compute termination fees in the base case by multiplying against initial pledge.
2. `FAULT_FEE_MULTIPLE = 1.05`: used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector.
3. `MIN_TERMINATION_FEE = XXX`: used to ensure the termination fee for young sectors is not arbitrarily low.

In addition, the existing lifetime cap constant will continue to be used in the new formulation:

```rust
// Maximum number of lifetime days penalized when a sector is terminated.
pub const TERMINATION_LIFETIME_CAP: ChainEpoch = 140;
```

`MIN_TERMINATION_FEE` needs defining @Schwartz10. IIUC it comes from [here](https://github.com/filecoin-project/FIPs/discussions/1036#discussioncomment-12253154), but in that formulation it's 2% of current. Maybe you intend to come up with a rough estimate of what the average 2% is today and use that?

FIPS/fip-0098.md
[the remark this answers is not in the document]
rvagg:
```suggestion
3. If the sector's age is less than `TERMINATION_LIFETIME_CAP` days, multiply the `termination_fee` by `sector age in days / TERMINATION_LIFETIME_CAP`
```

FIPS/fip-0098.md
```diff
@@ -52,12 +52,22 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
+5. If `termination_fee` is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
+
+The formula written out is:
+
+`termination fee = max(MIN_TERMINATION_FEE, max{TERM_PENALTY_PLEDGE_PERCENTAGE * initial_pledge, FAULT_FEE_MULTIPLE * sector_fault_fee} * min(1, sector_age_days / 140))`
```
rvagg, line 70:
```suggestion
`termination fee = max(MIN_TERMINATION_FEE, max{TERM_PENALTY_PLEDGE_PERCENTAGE * initial_pledge, FAULT_FEE_MULTIPLE * sector_fault_fee} * min(1, sector_age_days / TERMINATION_LIFETIME_CAP))`
```

FIPS/fip-0098.md
```diff
@@ -71,6 +81,16 @@ For all new sectors or snapped sectors, these three values are set to zero.
 
 These redundant fields can be removed from state in a future migration.
 
+In addition to the revised termination fee formulas, a few built-in actor methods needs to be created and exported so they are accessible from FEVM:
```
rvagg, line 84: Not critical for passing this FIP but before this goes live we need to have blanks filled in here with rough method signatures - names, param types, return types. We can't neglect the opportunity to bikeshed about naming of course!

rvagg: @LesnyRumcajs would you mind doing a sync with @Schwartz10 to get these figured out, we're running a bit low on time?

LesnyRumcajs: @rvagg Happy to help once @Schwartz10 is back (I believe he's a bit busy on a conference). In the meantime, I assume it's about https://github.com/filecoin-project/builtin-actors/blob/f8b60df640f64530f39e569d2afc39aaa473278e/actors/miner/src/lib.rs#L5994-L6033, correct?

rvagg: Yes, that would be it (I believe, tbh I haven't written one of these exported-to-fevm methods myself so I'm no expert), we just need to get the ball moving on this because I think we're looking at an actors code freeze for nv25 in ~10 days.

LesnyRumcajs: I mentioned it on Slack, but I'll forward some of my concerns here. I might be missing something, though, so I'll appreciate correcting me.

> adding those methods seems very loosely related to the FIP itself, i.e., simplifying the termination fee calculation. Am I missing something? It might make sense to omit those from the proposal (but implement them in the builtin actors regardless).

rvagg: I think they are _desired_ to also make it practical to handle this calculation in a smart contract so this logic can be built properly into lending contracts. At the moment it's too complex to properly account for, but once you can account for it in a contract then you can build logic on top of it.

I'm concerned that we're running short on time and we should ship this FIP in nv25. If we have to trim scope to do it then so be it, there may be pieces to pick up in nv26, like integrating the daily fee introduced in FIP-0100 if that ships with it.

jsoares (commented):
Probably best for Rod's comments to be addressed before further reviews, but a more general ask: when making changes during last call, it would be advisable to use the PR description to explain what changes are proposed and why.

FIPS/fip-0098.md
Schwartz10: I included a more well defined spec with method names and signatures here https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

FIPS/fip-0098.md
Schwartz10: Included change https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

FIPS/fip-0098.md
Schwartz10: Included a change here - https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

This was reported by @diogohavendp in the comment you tagged

FIPS/fip-0098.md
Schwartz10: Addressed here with Irene's suggestion in this commit https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

FIPS/fip-0098.md
Schwartz10: Fixed here https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

FIPS/fip-0098.md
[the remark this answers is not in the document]
Schwartz10: Fixed here https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

FIPS/fip-0098.md
Schwartz10: Sorry - my mistake, all fixed here https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

FIPS/fip-0098.md
Schwartz10: Added Irene's suggestion here: https://github.com/filecoin-project/FIPs/pull/1128/commits/e28e24c56309956c6ddd29d9d15bf64bd33f650f

Schwartz10:
@rvagg all outstanding comments have been addressed! @jsoares i've updated the description to highlight the changes that were made in last call

FIPS/fip-0098.md
```diff
@@ -52,12 +52,27 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
 
-A new constant variable called `TERM_PENALTY_PLEDGE_PERCENTAGE` should be created, which is fixed for the whole network and configurable by future governance.
+Three new constants should be created, each variable is fixed for the whole network and configurable by future governance:
+| Variable | Description | Amount |
+|----------|-------------|--------|
+| `TERM_PENALTY_PLEDGE_PERCENTAGE` | This constant will be used to compute termination fees in the base case by multiplying against initial pledge. | 8.5% of initial pledge |
+| `FAULT_FEE_MULTIPLE` | This constant will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector. | 1.05x |
+| `MIN_TERMINATION_FEE` | This constant will ensure the termination fee for young sectors is not arbitrarily low. | 2% of initial pledge |
+| `TERMINATION_LIFETIME_CAP` | This constant maintains the linear increase in termination fee for young sectors. | 140 days |
```
rvagg, line 61:
```suggestion
Three new constants should be created, and an existing one will be reused. Each is fixed for the whole network and configurable by future governance:

| Variable | Description | Amount |
|----------|-------------|--------|
| `TERM_FEE_PLEDGE_MULTIPLE` | This constant will be used to compute termination fees in the base case by multiplying against initial pledge. | 0.085x initial pledge |
| `TERM_FEE_MAX_FAULT_FEE_MULTIPLE` | This constant will be used to compute termination fees when the termination fee of a sector is less than the fault fee for the same sector. | 1.05x |
| `TERM_FEE_MIN_PLEDGE_MULTIPLE` | This constant will ensure the termination fee for young sectors is not arbitrarily low. | 0.02x initial pledge |
| `TERMINATION_LIFETIME_CAP` | _(Existing constant)_ This constant maintains the linear increase in termination fee for young sectors. | 140 days |
```

@LesnyRumcajs how do these sound? `_MULTIPLE` gets transformed to the `_NUM`+`_DENOM` form used in actors but I think these names make it clearer what their purpose is.

FIPS/fip-0098.md
```diff
@@ -52,12 +52,27 @@ Additionally, this FIP intends to maintain the original motivations behind termi
...
+4. Compute the per sector fault fee using the method `pledge_penalty_for_continued_fault` [link](https://github.com/filecoin-project/builtin-actors/blob/8a757341d4658461e70161d1f20f9b02991e8a5f/actors/miner/src/monies.rs#L148), if the computed termination fee is less or equal to fault fee, return `FAULT_FEE_MULTIPLE * fault_fee`
+5. If `termination_fee` is less than `MIN_TERMINATION_FEE`, return `MIN_TERMINATION_FEE`
+
+Note that the `pledge_penalty_for_continued_fault` method is not refactored in this FIP, but it must be verified to work for aggregate power numbers above the possible QA power for any single sector. For instance, an aggregate of 10 sectors' power should return the same end result as summing the `pledge_penalty_for_continued_fault` of each sector individually. This allows a caller to compute accurate maximum termination fees for any given aggregate power number (without looping through each sector).
+
+The formula written out is:
+
+`termination fee = max(MIN_TERMINATION_FEE, max{TERM_PENALTY_PLEDGE_PERCENTAGE * initial_pledge, FAULT_FEE_MULTIPLE * sector_fault_fee} * min(1, sector_age_days / TERMINATION_LIFETIME_CAP))`
```
rvagg, line 75: I'm just going to try and rework this and space it out a bit for clarity and using the suggested const names from above. _(This suggestion is meant to also replace line 65, I just can't include it here due to the way GH works)_:

````suggestion
1. Calculate the **base termination fee**:

```
base_termination_fee = initial_pledge * TERM_FEE_PLEDGE_MULTIPLE  // (0.085 * initial_pledge)
```

2. Apply the **age adjustment** for young sectors:

```
age_factor = min(1, sector_age_days / TERMINATION_LIFETIME_CAP)
age_adjusted_fee = base_termination_fee * age_factor
```

3. Calculate the **minimum allowed fee**:

```
minimum_fee = initial_pledge * TERM_FEE_MIN_PLEDGE_MULTIPLE  // (0.02 * initial_pledge)
```

4. Calculate the **maximum allowed fee** (based on fault fee, using the [existing `pledge_penalty_for_continued_fault` method]((https://github.com/filecoin-project/builtin-actors/blob/8a757341d4658461e70161d1f20f9b02991e8a5f/actors/miner/src/monies.rs#L148))):

```
fault_fee = pledge_penalty_for_continued_fault(sector_power)
maximum_fee = fault_fee * TERM_FEE_MAX_FAULT_FEE_MULTIPLE  // (1.05 * fault_fee)
```

5. Apply the fee bounds:

```
termination_fee = max(minimum_fee, min(age_adjusted_fee, maximum_fee))
```

This ensures that:

* The fee is never less than 2% of the initial pledge
* The fee is never more than 105% of the fault fee
* For sectors younger than 140 days, the fee scales linearly with age
* For sectors older than 140 days, the fee is 8.5% of the initial pledge (subject to the bounds)

The complete formula in one expression is:

```
termination_fee = max(
    initial_pledge * TERM_FEE_MIN_PLEDGE_MULTIPLE,
    min(
        initial_pledge * TERM_FEE_PLEDGE_MULTIPLE * min(1, sector_age_days / TERMINATION_LIFETIME_CAP),
        fault_fee * TERM_FEE_MAX_FAULT_FEE_MULTIPLE
    )
)
````

File: FIPS/fip-0098.md [hunk 12]
```diff
@@ -71,6 +86,18 @@ For all new sectors or snapped sectors, these three values are set to zero.
 
 These redundant fields can be removed from state in a future migration.
 
+In addition to the revised termination fee formulas, a few built-in actor methods need to be created and exported so they are accessible from FEVM:
+| Function Name | Function Signature | Description |
+|--------------|-------------------|-------------|
+| `get_miner_power` | `get_miner_power(miner_id: ActorID) -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
+| `get_miner_initial_pledge` | `get_miner_initial_pledge(miner_id: ActorID) -> TokenAmount` | Returns the miner's total initial pledge amount |
+| `get_network_projection` | `get_network_projection(projection_period_start: ChainEpoch, projection_period_end: ChainEpoch) -> (Power, TokenAmount)` | Returns the network's projected power and reward for the given projection period |
+| `get_termination_fee_percentage` | `get_termination_fee_percentage() -> f64` | Returns the network's termination fee percentage (8.5%) |
```
rvagg, line 95: @LesnyRumcajs this probably means this value should be in the actors policy framework rather than just a plain constant.

FIPS/fip-0098.md
LesnyRumcajs: LGTM. In builtin-actors we also use the `_factor`, e.g., `LOCK_TARGET_FACTOR`, but no reason to bikeshed.

FIPS/fip-0098.md
LesnyRumcajs: Is this going to be variable across different networks? I'm not opposed to this, but I'd like to understand what makes some constants get promoted to the actors policy from plain constants.

LesnyRumcajs, line 95: It's name bikeshedding time! To my understanding, the naming is to be implemented in the power actor, around [here](https://github.com/filecoin-project/builtin-actors/blob/f8b60df640f64530f39e569d2afc39aaa473278e/actors/power/src/lib.rs#L54-L58)

```rust
    // Method numbers derived from FRC-0042 standards
    CreateMinerExported = frc42_dispatch::method_hash!("CreateMiner"),
    NetworkRawPowerExported = frc42_dispatch::method_hash!("NetworkRawPower"),
    MinerRawPowerExported = frc42_dispatch::method_hash!("MinerRawPower"),
    MinerCountExported = frc42_dispatch::method_hash!("MinerCount"),
    MinerConsensusCountExported = frc42_dispatch::method_hash!("MinerConsensusCount"),
```

It seems that existing methods don't have the `Get` prefix, so for consistency, I'd also avoid them.

FIPS/fip-0098.md
```diff
@@ -104,8 +136,14 @@ This FIP requires a network upgrade because it intends to change built-in actor
...
+- When terminating a single sector:
+  - Not considering fault fees, for a sector where its age >= `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_PENALTY_PLEDGE_PERCENTAGE * initial pledge`
+  - Not considering fault fees, for a sector where its age < `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_PENALTY_PLEDGE_PERCENTAGE * of initial pledge * sector age in days / TERMINATION_LIFETIME_CAP`
+  - Considering fault fees, for a sector with a termination fee that is less than the associated sector's fault fee, termination fee should equal `FAULT_FEE_MULTIPLE * fault fee`
+  - Given all test cases above, if the termination fee computed is less than `MIN_TERMINATION_FEE * initial pledge`, termination fee should equal `MIN_TERMINATION_FEE * initial pledge`
+
+- Additional test case for the `pledge_penalty_for_continued_fault` method:
+  - `pledge_penalty_for_continued_fault` should work for aggregate power numbers above the possible QA power for any single sector. For instance, an aggregate of 10 sectors' power should return the same end result as summing the `pledge_penalty_for_continued_fault` of each sector individually.
```
LesnyRumcajs, line 146: @Schwartz10 I contrived a test for this https://github.com/filecoin-project/builtin-actors/pull/1639/commits/d854b85b9a5b2b3e181e3fbe850a1e2a99e28092

Do we allow some wiggle room for this assertion, or may there be a bug in the logic somewhere? 

```
  left: TokenAmount(69673276.724698680084565705) (aggregate)
 right: TokenAmount(69673276.7246986800845657) (individual * 10)
```

rvagg: heh; so playing with your test, it looks like a .57 attofil per sector overreach when doing aggregates. I think this all comes down to the magic in `extrapolated_cum_sum_of_ratio` and the fact that it's operating on very small variances so does a shift by `PRECISION` to try and get as accurate as possible given the need to divide & floor at at least one point. I'm finding a very similar problem with the FIP-0100 and end asserting an absolute difference of 1 attofil in a couple of cases.
I'm not sure there's much we can do about this, unless perhaps we want to insist that you only ever call it for an individual amount of QAP and multiply it out by what you have, but that doesn't seem ideal.

FIPS/fip-0098.md
Schwartz10: Just deleted line 65 in https://github.com/filecoin-project/FIPs/pull/1128/commits/7f0ddd20165bb2cd008d53a3c80a991cbb941d72

FIPS/fip-0098.md
Schwartz10: Thinking out loud here:

- From EVM perspective at least, <1 atto imprecision can be ignored. the smallest denomination we care about is 1 atto. Not sure we have precision limits in built-in actors, so this may only be applicable in the context of FEVM 
- We should double check this with very large numbers of aggregates - for instance, test some of the miners on the biggest miners list here https://filfox.info/en/ranks/power - how far off do we get on an aggregate calculation of the biggest miner? We could even hypothetically 10x duplicate each sector to really push the imprecision to the max and see how bad it gets

FIPS/fip-0098.md
Schwartz10: Fixed here! https://github.com/filecoin-project/FIPs/pull/1128/commits/1869dfa37175ecbfacf91b679433d5a712e4ec78

Schwartz10: I'm not the best person to answer where these variables should live, I don't think it's necessary to have per network values for these variables. Actually seems more useful to always keep them in sync.

FIPS/fip-0098.md
LesnyRumcajs: @Schwartz10 It's 5 atto, not 705.
```
  left: 69673276724698680084565705
 right: 69673276724698680084565700
```
On 100 PiB QAP * 10 it's 3 atto.
```
  left: 114152696585746317450552452276413
 right: 114152696585746317450552452276410
```
I tinkered with different knobs but didn't see the difference go above nine atto.

FIPS/fip-0098.md
```diff
@@ -104,8 +175,14 @@ This FIP requires a network upgrade because it intends to change built-in actor
...
 
-- When terminating a single sector, the realized termination fee should equal the expected % of pledge termination fee
-- When terminating all sectors on a miner, the realized termination fee should equal the expected % of pledge termination fee
+- When terminating a single sector:
+  - Not considering fault fees, for a sector where its age >= `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_PENALTY_PLEDGE_PERCENTAGE * initial pledge`
+  - Not considering fault fees, for a sector where its age < `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_PENALTY_PLEDGE_PERCENTAGE * of initial pledge * sector age in days / TERMINATION_LIFETIME_CAP`
+  - Considering fault fees, for a sector with a termination fee that is less than the associated sector's fault fee, termination fee should equal `FAULT_FEE_MULTIPLE * fault fee`
+  - Given all test cases above, if the termination fee computed is less than `MIN_TERMINATION_FEE * initial pledge`, termination fee should equal `MIN_TERMINATION_FEE * initial pledge`
```
LesnyRumcajs, line 182: Constants should be adjusted to the ones from the table defined in the beginning.
```suggestion
  - Not considering fault fees, for a sector where its age >= `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_FEE_PLEDGE_MULTIPLE * initial pledge`
  - Not considering fault fees, for a sector where its age < `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_FEE_PLEDGE_MULTIPLE * of initial pledge * sector age in days / TERMINATION_LIFETIME_CAP`
  - Considering fault fees, for a sector with a termination fee that is less than the associated sector's fault fee, termination fee should equal `TERM_FEE_MAX_FAULT_FEE_MULTIPLE * fault fee`
  - Given all test cases above, if the termination fee computed is less than `TERM_FEE_MIN_PLEDGE_MULTIPLE * initial pledge`, termination fee should equal `TERM_FEE_MIN_PLEDGE_MULTIPLE * initial pledge`
```

FIPS/fip-0098.md
```diff
@@ -104,8 +175,14 @@ This FIP requires a network upgrade because it intends to change built-in actor
 
 <!--Test cases for an implementation are mandatory for FIPs that are affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
 
-- When terminating a single sector, the realized termination fee should equal the expected % of pledge termination fee
-- When terminating all sectors on a miner, the realized termination fee should equal the expected % of pledge termination fee
+- When terminating a single sector:
+  - Not considering fault fees, for a sector where its age >= `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_PENALTY_PLEDGE_PERCENTAGE * initial pledge`
+  - Not considering fault fees, for a sector where its age < `TERMINATION_LIFETIME_CAP`, termination fee should equal `TERM_PENALTY_PLEDGE_PERCENTAGE * of initial pledge * sector age in days / TERMINATION_LIFETIME_CAP`
+  - Considering fault fees, for a sector with a termination fee that is less than the associated sector's fault fee, termination fee should equal `FAULT_FEE_MULTIPLE * fault fee`
```
LesnyRumcajs, line 181: Is this correct? The currently proposed bounds in this PR suggest it's the other way around; if the termination fee is larger than the `fault_fee * TERM_FEE_MAX_FAULT_FEE_MULTIPLE`, it's bound to to that max. If the termination fee is smaller, it gets limited only by the `TERM_FEE_MIN_PLEDGE_MULTIPLE`
```
fault_fee = pledge_penalty_for_continued_fault(sector_power)
maximum_fee = fault_fee * TERM_FEE_MAX_FAULT_FEE_MULTIPLE  // (1.05 * fault_fee)
termination_fee = max(minimum_fee, min(age_adjusted_fee, maximum_fee))
```

FIPS/fip-0098.md
```diff
@@ -71,6 +125,18 @@ For all new sectors or snapped sectors, these three values are set to zero.
...
+|--------------|-------------------|-------------|
+| `miner_power` | `miner_power(miner_id: ActorID) -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
+| `miner_initial_pledge` | `miner_initial_pledge(miner_id: ActorID) -> TokenAmount` | Returns the miner's total initial pledge amount |
+| `network_projection` | `network_projection(projection_period_start: ChainEpoch, projection_period_end: ChainEpoch) -> (Power, TokenAmount)` | Returns the network's projected power and reward for the given projection period |
+| `termination_fee_percentage` | `termination_fee_percentage() -> f64` | Returns the network's termination fee percentage (8.5%) |
+  
+(the below 2 methods could also be built as a FEVM native oracle contract instead of as a built-in actor method with exports to access from FEVM):
+| `single_fault_fee` | `single_fault_fee(qa_power: QAPower) -> TokenAmount` | Returns the fault fee calculation for a given amount of QA power |
+| `max_termination_fee` | `max_termination_fee(initial_pledge: TokenAmount, power: QAPower) -> TokenAmount` | Returns the maximum termination fee calculation for a given initial pledge and power amount |
```
LesnyRumcajs, line 138: Is there any existing example of such an approach? Is this required?

FIPS/fip-0098.md
```diff
@@ -71,6 +125,18 @@ For all new sectors or snapped sectors, these three values are set to zero.
 
 These redundant fields can be removed from state in a future migration.
 
+In addition to the revised termination fee formulas, a few built-in actor methods need to be created and exported so they are accessible from FEVM:
```
LesnyRumcajs, line 128: It might make sense to add another column - which actor this method belongs to. This would also affect signatures. As I see it:
- `power`: `miner_power`, 
- `miner`: `miner_initial_pledge`,
- ?: `termination_fee_percentage`, it's just a constant defined in the `miner`, so I guess there?
- ?: `network_projection`, could you please point me to a more specific implementation of this?

Schwartz10: cc @anorth for help / assistance on this

FIPS/fip-0098.md
Schwartz10: Are you asking about the max_termination_fee method? I'm not sure I understand what you're asking, but there wouldn't be an example as this FIP is now creating a new way to compute that

From my perspective, this is the single most important part of the fip - it allows FEVM actors to get a collateral value for a miner

FIPS/fip-0098.md
Schwartz10: Whoops my mistake around 705 number. 9 atto seems safu from my perspective, but it might also make sense to test with many RBP sectors (without multipliers) as presumably the precision issue will be the worst on the smallest amount of power?

FIPS/fip-0098.md
LesnyRumcajs: I mean an example of `FEVM native oracle contract`.

Schwartz10: Oooh, not off the top of my head, but i think the name is fancier than the concept - just a singleton FEVM actor that:

1. Exposes methods like `max_termination_fee`
2. When called, makes the associated calls to exported built-in actor methods to gather the building blocks to compute the answer
3. Computes the answer and returns it to the caller

In my book, this is "less good" than just including these methods in built-in actors because it will be harder / messier to maintain. Also potentially harder to upgrade in the future, but it is possible

FIPS/fip-0098.md
Schwartz10: @LesnyRumcajs really good catch here - my tired eyes failed me last night. I fixed in https://github.com/filecoin-project/FIPs/pull/1128/commits/a31032e3fc1a070dc7ce279e9098f1bf09c153e2

Noting that the fault_fee and min_termination_fee are not actually discounted by sector day age. Should we be applying the young-sector-age discount to fault fees? I don't _think_ so, but @irenegia might be a good person to weigh in on this

FIPS/fip-0098.md
Stebalien: Thoughts below but my main question is: how are you planning on using these methods? On-chain? Off-chain? Per-sector? For batches of sectors? That'll impact exactly what the final design should look like.

I'd much prefer to just have a built-in method so we can change it on network upgrades. I'd prefer not to expose any `network_projection` or `termination_fee_percentage` methods, instead exposing _just_ a `max_termination_fee` method (and maybe a `single_fault_fee` method). I don't want to expose too many of the internals of how these fees are calculated in case they change in the future.

I'd prefer to expose these methods in terms of "given a set of sectors X/Y/Z, what is the max termination fee?", but I think it's _probably_ also safe to do this in terms of power/pledge (we just have to guarantee that we'll never take any other per-sector parameters into account, but I think that's safe).

In terms of accuracy, the `max_termination_fee` isn't going to be 100% accurate if called for a batch of sectors (sum of power, sum of pledge). It'll still be correct when called sector-by-sector.

Schwartz10: @Stebalien thanks for chiming in here - this will be used on-chain to get the collateral value of a miner. It would look something like this in pseudo code:

```
var minerAddress = 0x0000000000....xxx
var minerID = xxx

var minerBalance = address(minerAddress).balance
var minerIP = precompiles.miner_initial_pledge(minerID)
var minerPower = precompiles.miner_power(minerID)
var minerMaxTermFee = precompiles.max_termination_fee(minerIP, minerPower)

var minerCollateralValue = minerBalance - minerMaxTermFee

var minerDebt = fetchFromContractToGetDebt(minerID)

var isUnderCollateralized = minerDebt > minerCollateralValue

if isUnderCollateralized { ... do something ... }
else { ... do something ...}
```

Something like that. 

It is quite useful to be able to get power and initial pledge for other reasons, but ultimately the most important thing to come from this fip is a supported method to get the max termination fee that is supported by the core devs, upgraded as it needs to, and reports reliable _enough_ info to rely on. It would be easier UX if the method took the minerID instead of IP and power, but IP and power are more composable primitives, per @anorth point in a previous message

FIPS/fip-0098.md
```diff
@@ -17,15 +17,15 @@ created: 2024-09-26
...
 
 ## Abstract
 
 <!--A short (~200 word) description of the technical issue being addressed.-->
 
 Today, it is overly sophisticated and computationally expensive to compute termination fees for miner actors. As a result, DeFi applications that leverage miner actors must make major security and/or performance sacrifices in order to operate.
 
-This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 8.5%.
+This FIP addresses the issue by simplifying the calculation of the miner actor termination fee. The proposed new termination fee calculation uses a "percentage of pledge" strategy - where `termination fee = initial pledge * termination penalty %`. The proposed new termination penalty % is 8.5%. There are two special cases to consider where the termination fee must be tweaked to maintain current network conditions, which are addressed in the technical specification and design rationale.
```
anorth, line 28: "two special cases" here inconsistent with "3 special cases" above

FIPS/fip-0098.md
anorth: The network projection couldn't easily be a single call, as different actors have the state for each part:
- Reward actor has the reward estimator
- Power actor has the power estimator

AFAIK the only possible projection_period_start is the current epoch, so that cannot be a parameter to either.

But I think these are probably too low level. The discussion below suggests a maxTerminationPenalty method which sounds better. Whatever actor it's placed on, it will have to call to the reward and power actors to get the estimators.

We don't have prior art for where to put a constant, since actors all just have them compiled in.
Perhaps put the termination fee % on the power actor, as a well-defined singleton. But if @Stebalien  or @ZenGround0 suggest something else, go with that. Or just don't expose this constant, abstract over it with a method for computing termination fees.

FIPS/fip-0098.md
anorth: I concur with @Stebalien I'd rather not expose the primitive methods, just the useful higher level ones (apologies if I suggested otherwise elsewhere, @Schwartz10). I do recall that I suggested the max_termination_fee method take pledge and power as parameters though; I think that makes it more useful since it can be used for one or more sectors, as well an entire miner, so long as the caller knows or can get the associated power and pledge.

But i'm not wedded to that – happy to let @Stebalien make a final decision.

anorth: Unasked question about what actor to put these two final methods on. It'll need to call to both power and reward actor. The power actor would be a reasonable singleton place IMO, but could be convinced otherwise.

FIPS/fip-0098.md
LesnyRumcajs: In general, the diff seems to be always somewhere below 1 atto per sector https://github.com/filecoin-project/builtin-actors/pull/1639/commits/b4f78b46d1eca17cc31a7e54f1ec2dfa96904e75

On thousands of sectors it will go up to thousands of atto, but I guess at this scale it should be okay?

FIPS/fip-0098.md
LesnyRumcajs: Let's wait for @irenegia, then.

FIPS/fip-0098.md
Schwartz10: Yeah, so if we assume 1 atto imprecision per sector, even at 1m sectors, we'd be off by `0.000000000001 FIL` - that seems safu to me

FIPS/fip-0098.md
Schwartz10: Good catch

Schwartz10:
@anorth @Stebalien @LesnyRumcajs I removed some methods that were "low level" in favor of the necessary ones from an FEVM perspective. You could also get rid of `termination_fee_percentage` here but i think this will become useful to invalidate assumptions if things change. I.e. - if a lending app checks this percentage every so often and it comes back _different_, we know something in our assumptions has changed, and we might issue an update on our end for handling logic differently. Less important than the others, but seems easy enough to export? Whatever y'all think

FIPS/fip-0098.md
```diff
@@ -71,6 +121,14 @@ For all new sectors or snapped sectors, these three values are set to zero.
 
 These redundant fields can be removed from state in a future migration.
 
+In addition to the revised termination fee formulas, a few built-in actor methods need to be created and exported so they are accessible from FEVM:
+| Function Name | Function Signature | Description |
+|--------------|-------------------|-------------|
+| `miner_power` | `miner_power(miner_id: ActorID) -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
+| `miner_initial_pledge` | `miner_initial_pledge(miner_id: ActorID) -> TokenAmount` | Returns the miner's total initial pledge amount |
+| `termination_fee_percentage` | `termination_fee_percentage() -> f64` | Returns the network's termination fee percentage (8.5%) |
```
LesnyRumcajs, line 129: If we include this method, should the return type be `f64`? I'm not sure how big of a deal this is, but it's a floating point type, so it might not return exactly 0.085. I'm not sure we have a better way of handling those anywhere. Perhaps it should return the numerator and denominator (how it is handled in the code).

LesnyRumcajs: @Schwartz10 I went ahead with my suggestion in https://github.com/filecoin-project/builtin-actors/pull/1639. Does this work for you? Is everyone in agreement that this method should be there at all?

FIPS/fip-0098.md
```diff
@@ -71,6 +121,14 @@ For all new sectors or snapped sectors, these three values are set to zero.
 
 These redundant fields can be removed from state in a future migration.
 
+In addition to the revised termination fee formulas, a few built-in actor methods need to be created and exported so they are accessible from FEVM:
+| Function Name | Function Signature | Description |
+|--------------|-------------------|-------------|
+| `miner_power` | `miner_power(miner_id: ActorID) -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
+| `miner_initial_pledge` | `miner_initial_pledge(miner_id: ActorID) -> TokenAmount` | Returns the miner's total initial pledge amount |
```
LesnyRumcajs, line 128: Given `miner_initial_pledge` lives under `miner`, it probably makes sense to rename it to `initial_pledge` and omit the parameters (the parameter is implicitly `this`/`self`)

Schwartz10: from the context of the fevm smart contracts we definitely want to be able to pass a parameter

FIPS/fip-0098.md
Schwartz10: Good point - i dont mind returning two values for num and denom, but whatever those values return will likely end up getting converted into `uint256` types in solidity, so ideally each returned number is convertible into `uint256`

LesnyRumcajs: They are `u32` for now, so it should be fine.

FIPS/fip-0098.md
Stebalien: I agree. So:

1. Add a method to the power actor to get a miner's QA power. We currently only export a method to get the miner's _raw_ byte power (i.e., the amount of space they're proving).
2. Add a method to the miner actor to get the initial pledge.
3. Add a `MaxTerminationFee` method to the power actor.

FIPS/fip-0098.md
Schwartz10: Actually, i think you are right - i think you would make a solidity wrapper method that passes the miner ID like we do [here](https://github.com/filecoin-project/filecoin-solidity/blob/c72ec4bf8213f38ca5cd6f553abff5cab0709488/contracts/v0.8/MinerAPI.sol#L287).   Fixed in https://github.com/filecoin-project/FIPs/pull/1128/commits/f5bd8e9a531c5062bb7c272a737c583d4b9f6318

FIPS/fip-0098.md
Schwartz10: Updated https://github.com/filecoin-project/FIPs/pull/1128/commits/074cc7deb84cf5484f72f81740b812ded41b18a7

FIPS/fip-0098.md
LesnyRumcajs: > Add a MaxTerminationFee method to the power actor.

@Stebalien Implementation question: given the current termination fee calculation lives in the [miner code](https://github.com/filecoin-project/builtin-actors/blob/3a963754234ea64e0585b3012b2fb023aa207614/actors/miner/src/monies.rs#L191), where would you move it? Alternatively, should the `MaxTerminationFee` live under the miner actor?

Also, given that you didn't list `termination_fee_percentage`, should it (or not) be included in the FIP and its implementation?

Stebalien: > @Stebalien Implementation question: given the current termination fee calculation lives in the [miner code](https://github.com/filecoin-project/builtin-actors/blob/3a963754234ea64e0585b3012b2fb023aa207614/actors/miner/src/monies.rs#L191), where would you move it? Alternatively, should the MaxTerminationFee live under the miner actor?

I'd prefer it if it lived under the power actor because it's not specific to any miner and putting it there will save some gas (one fewer external call).

The question is: where to put the actual code as it's shared between the power actor and the miner actor. FOr now, I'd put it in `runtime::builtin::power` (mirroring `runtime::builtin::reward`). I'm not a huge fan of that, but the alternatives are:

1. Define `MaxTerminationFee` _on_ the miner actor itself. We could do this, but it'll cost more gas (it'll have to call out to the power actor and the reward actor) and it's _supposed_ to be miner-independent. Although I guess one benefit of defining it on the miner actor is that it allows us to _possibly_ make this fee miner-dependent in the future.
2. Define some new `actors/miner/shared` crate (as we do with `actors/evm/shared`. But I'd rather not add another crate right now. Also, moving this logic out of the runtime crate in the future shouldn't be too difficult.

> Also, given that you didn't list termination_fee_percentage, should it (or not) be included in the FIP and its implementation?

I don't think it should be exposed unless users _really_ need it. Exposing it would lock us into keeping the fee as a strict percentage of the pledge and that's not accurate even given the current fee.

LesnyRumcajs: > I'd prefer it if it lived under the power actor because it's not specific to any miner and putting it there will save some gas (one fewer external call).

Great, thanks for the pointers, @Stebalien. I'll put it into the `ruintime::builtin::power`, then. I initially did the 1st alternative as I felt it conceptually belongs in that domain (as a, in a way, static method), but it introduces an additional call.

>I don't think it should be exposed unless users really need it. Exposing it would lock us into keeping the fee as a strict percentage of the pledge and that's not accurate even given the current fee.

@Schwartz10 What do you think?

Stebalien: I guess the real question is: how gas sensitive is this? Assuming this is only called in exceptional cases when terminating large batches of sectors, putting it in the miner actor isn't going to be a huge issue. And I agree it makes everything simpler.

@anorth or @rvagg can we get a third opinion on this? Specifically: put a `MaxTerminationFee` method on either (a) the miner actor (where all the logic lives now) or (b) the Power actor to save some gas (also because the method isn't currently miner-specific).

Stebalien: Basically, it likely doesn't matter much either which way and we might as well keep this as simple as possible. If nobody objects, we might as well leave it on the miner actor.

rvagg: I'm not too fussed, my standards on abstraction purity vs cost is a bit lower than some, but actors currently does have fairly nice separation of concerns so we ought to be careful not to break the barriers down too much. I'd like to see what happens to the constants and other building-blocks when we move this around.

Is `max_termination_fee` just `initial_pledge * TERM_FEE_MIN_PLEDGE_MULTIPLE`? So `TERM_FEE_MIN_PLEDGE_MULTIPLE` would need to move to the top-level to be reusable I guess, probably argues for at least this one to go into the `Policy` but maybe we should put more (or all, including `TERMINATION_LIFETIME_CAP`).

Stebalien: It's a little more involved. This is the function we're talking about:

https://github.com/filecoin-project/builtin-actors/pull/1639/files#diff-fbcb2ec1a9d82b18f146c728cafd643df0e7ae47a04d84be7644913fe89236e5R2224-R2253

rvagg: Oh, yuk, so that pulls in all of the `pledge_penalty_*` cruft which would need to be moved up to somewhere shared .. or something. Yeah, I don't know then, because doing it on the miner actor is pretty weird since it doesn't involve the actor at all.

Stebalien: It doesn't involve the miner actor but... it's still "miner actor logic". It's effectively a static function.

Also, this _does_ mean that we can _potentially_ introduce miner-specific parts to the fee (e.g., if we need to charge different fees for different proof types, etc). I can't imagine that ever being an issue, but it's _a_ reason this could make sense.

TippyFlitsUK (approved):

FIPS/fip-0098.md
```diff
@@ -126,7 +198,7 @@ It is important that Storage Providers are incentivized to honor their commitmen
 
 The termination fee for sectors committed a long time ago will immediately and significantly decrease, but only to match the fee:pledge ratio of new sectors. This will increase the incentive for some SPs with high-pledge sectors to terminate and re-onboard to gain more power for the same amount of pledge (but more hardware).
 
-In contrast to the current system, the termination penalty for a sector will be constant regardless of its age. Currently the termination penalty starts low and increases with age up to a max. This presents is a change in the incentive structure for short-term-view SPs.
+In contrast to the current system, the termination penalty for a sector will be constant regardless of its age. Currently the termination penalty starts low and increases with age up to a max. This presents a change in the incentive structure for short-term-view SPs.
```
momack2, line 201:
```suggestion
In contrast to the current system, the termination penalty for a sector will be constant regardless of its age (after the 140 day linear ramp-up). Currently the termination penalty starts low and increases with age up to a max. This presents a change in the incentive structure for short-term-view SPs.
```

FIPS/fip-0098.md
```diff
@@ -126,7 +198,7 @@ It is important that Storage Providers are incentivized to honor their commitmen
 
 The termination fee for sectors committed a long time ago will immediately and significantly decrease, but only to match the fee:pledge ratio of new sectors. This will increase the incentive for some SPs with high-pledge sectors to terminate and re-onboard to gain more power for the same amount of pledge (but more hardware).
```
momack2, line 199: Flag - I'd love this to be more specific if possible (ex "it will tend to decrease termination fee for sectors over X years/months old"). 

Also - I presume there is NOT an incentive to terminate and re-onboard sectors with deals and FIL+ verified datacap - because they would have to get net-new datacap granted. Correct?

momack2 (approved):

FIPS/fip-0098.md
Schwartz10: @momack2 thanks for pointing this out - re live network stats, I'm absolutely swamped with work this week with GLIF, so i don't have enough time to recalculate, but i've linked to the relevant data in the discussion for this fip in this commit https://github.com/filecoin-project/FIPs/pull/1128/commits/3b58d9f3de41f087b79e9c104253aae982f86feb

You're correct - presumably there is no incentive to repledge QAP as DC is apparently quite hard to come by.

momack2: Thanks! that helps a lot to assuage potential issues.

momack2:
@Schwartz10 @LesnyRumcajs - this has the requisite approvals now - is it ready to merge?

LesnyRumcajs:
> @Schwartz10 @LesnyRumcajs - this has the requisite approvals now - is it ready to merge?

I believe the `termination_fee_percentage` exported method should still be removed from the FIP per https://github.com/filecoin-project/FIPs/pull/1128#discussion_r1988151166 and the [Slack thread](https://filecoinproject.slack.com/archives/C01EU76LPCJ/p1741891466720849?thread_ts=1741775567.319219&cid=C01EU76LPCJ).

FIPS/fip-0098.md
```diff
@@ -71,26 +121,39 @@ For all new sectors or snapped sectors, these three values are set to zero.
 
 These redundant fields can be removed from state in a future migration.
 
+In addition to the revised termination fee formulas, a few built-in actor methods need to be created and exported so they are accessible from FEVM:
+| Function Name | Function Signature | Description |
+|--------------|-------------------|-------------|
+| `miner_power` | `miner_power() -> (QAPower, RawPower)` | Returns the miner's quality-adjusted and raw power |
+| `miner_initial_pledge` | `miner_initial_pledge() -> TokenAmount` | Returns the miner's total initial pledge amount |
+| `termination_fee_percentage` | `termination_fee_percentage() -> (num u32, denom u32)` | Returns the network's termination fee percentage  (8.5%) |
```
LesnyRumcajs, line 129:
```suggestion
```

rjan90: Since @Schwartz10 commented [here in Slack](https://filecoinproject.slack.com/archives/C01EU76LPCJ/p1741891646795239?thread_ts=1741775567.319219&cid=C01EU76LPCJ), this was not mission-critical. Jonathan flagged to me that he is currently overwhelmed with other tasks, so I don't think it's necessary to wait for him to click the button.

LesnyRumcajs:
I believe all the outstanding comments were addressed. Mergio?

PR #1130, opened 2025-02-25
FIP-0100: Add VestingFunds optimisation details

rvagg:
This is an addition to https://github.com/filecoin-project/FIPs/pull/1124 that's large enough and different enough in nature that it should be reviewed separately. It doesn't impact the fee mechanism, it's an internal detail for the miner actor, but it needs to be documented as a change in state representation, mutation and how we perform a migration.

Details being finalised in https://github.com/filecoin-project/FIPs/pull/1124.

### DIFF
--- a/FIPS/fip-0100.md
+++ b/FIPS/fip-0100.md
@@ -246,11 +246,89 @@ Sectors that were onboarded prior to the activation of this FIP, and therefore d
 | **Faults** | No change | No change | Update if sectors move between queues | No direct impact on fee calculation |
 | **Early terminations** | N/A (sector removed) | Reduce by terminated sector's `daily_fee` | N/A (sector removed) | Applies to both automatic and manual terminations |
 
+### Optimization of `VestingFunds`
+
+The `VestingFunds` data structure, associated with the miner actor's state, is currently implemented as a single, large array of vesting tuples: `(ChainEpoch, TokenAmount)`. Each tuple specifies the amount of tokens to be unlocked at a specific epoch. Vesting epochs are quantized, meaning that vesting occurs at regular intervals, but not necessarily every epoch or even every deadline. This array can become very large, leading to inefficiencies when it needs to be updated. These updates occur when new vesting block rewards are added, when tokens vest and are unlocked, and when locked rewards are used to cover fee debts arising from penalties.
+
+The introduction of a daily fee, paid using the fee debt mechanism, increases the frequency with which the `VestingFunds` data structure is mutated. To mitigate the performance impact of these frequent mutations, this FIP introduces an optimization to the storage of `VestingFunds`. The goal is to make the common operation of drawing funds from the _next_ vesting epoch more efficient, avoiding unnecessary reads and writes to the large `VestingFunds` array.
+
+**Existing state representation**
+
+```rust
+#[derive(Serialize_tuple, Deserialize_tuple)]
+pub struct State {
+    // ... existing fields
+
+    /// VestingFunds (Vesting Funds schedule for the miner).
+    pub vesting_funds: Cid,
+
+    // ... existing fields
+}
+
+/// Represents the vesting table state for the miner.
+/// It is a slice of (VestingEpoch, VestingAmount).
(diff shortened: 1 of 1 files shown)


FIPS/fip-0100.md
```diff
@@ -234,12 +234,94 @@ Sectors that were onboarded prior to the activation of this FIP, and therefore d
 
 **Early terminations**: both automatic and manual, will result in a reduction of a `Deadline`'s `daily_fee` value for the sectors being terminated.
 
+### Optimization of `VestingFunds`
+
+The `VestingFunds` data structure, associated with the miner actor's state, is currently implemented as a single, large array of vesting tuples: `(ChainEpoch, TokenAmount)`. Each tuple specifies the amount of tokens to be unlocked at a specific epoch. Vesting epochs are quantized, meaning that vesting occurs at regular intervals, but not necessarily every epoch or even every deadline. This array can become very large, leading to inefficiencies when it needs to be updated. These updates occur when new vesting block rewards are added, when tokens vest and are unlocked, and when locked rewards are used to cover fee debts arising from penalties.
```
Stebalien, line 239: note: very large -> 360. In fact, it'll pretty much always have 360 elements for active SPs. So, not huge, but definitely a few KiB (which is a lot in blockchain contexts).

Stebalien (commented):
Looks correct to me.

FIPS/fip-0100.md
BigLep: So maybe discuss what the upperbound is inline?

My biggest question from reading this is: is the increased complexity worth it?  (I don't know enough about this stuff to make that judgement, but I think getting some numbers in would give us more confidence that this is worth it.)

rvagg: @Stebalien @BigLep for fun I ran the numbers we were theorising about today, across the ~1.6k miners with power:

```
VestingFunds size statistics:
  Total VestingFunds blocks: 1671
  Total VestingFunds blocks size: 9,501,236 bytes
  Min size: 2 bytes
  Max size: 6096 bytes
  Average size: 5685.96 bytes
  Median size: 5778 bytes

Size distribution:
  2-611 bytes: 10 miners
  612-1221 bytes: 0 miners
  1222-1831 bytes: 0 miners
  1832-2441 bytes: 4 miners
  2442-3051 bytes: 17 miners
  3052-3661 bytes: 6 miners
  3662-4271 bytes: 9 miners
  4272-4881 bytes: 29 miners
  4882-5491 bytes: 77 miners
  5492-6096 bytes: 1519 miners
```

`9501236/1024/1024 * 48 = 435 MiB` of churn each day if we mutated all of these each deadline.

I'll include these stats in the update.

FIPS/fip-0100.md
```diff
@@ -246,12 +246,98 @@ Sectors that were onboarded prior to the activation of this FIP, and therefore d
...
+#[serde(transparent)]
+pub struct VestingFunds(Option<VestingFundsInner>);
+
+#[derive(Serialize_tuple, Deserialize_tuple)]
+struct VestingFundsInner {
+    // The "next" batch of vesting funds.
+    head: VestingFund,
+    // The rest of the vesting funds, if any.
+    tail: Cid, // Vec<VestingFund>
+}
```
Stebalien, line 313: For the sake of the FIP, I'd consider describing this as:

```suggestion
    pub vesting_funds: Option<VestingFunds>,

    // ... existing fields
}

/// Represents the vesting table state for the miner. It's composed of a `head` (stored inline) and
/// a tail (referenced by CID).
#[derive(Serialize_tuple, Deserialize_tuple)]
struct VestingFunds {
    // The "next" batch of vesting funds.
    head: VestingFund,
    // The rest of the vesting funds, if any.
    tail: Cid, // Vec<VestingFund>
}
```

Stebalien: The representation is the same. The fact that we have a `VestingFundsInner` is, IMO, an implementation detail.

Stebalien (commented):
Still LGTM, although I'd prefer to simplify the description of how we encode the state in the FIP.

PR #1150, opened 2025-05-16
FIP-0107: Implicit Messages in Block Receipts

rvagg:
# FIP Draft: Implicit Messages in Block Receipts

This FIP adds implicit messages (cron and reward) to block receipts, allowing their execution to be inspected and events generated by system operations to be properly captured. It primarily improves visibility into sector lifecycles and simplifies determination of executed messages across tipset blocks.

The proposal:
1. Formalises implicit messages so they can be linked in the chain
2. Includes links to executed messages in receipts
3. Adds codec information for proper return data interpretation
4. Makes system operations (especially sector terminations) visible through standard event mechanisms

Currently, storage providers and developers must resort to complex state diff analysis to understand sector lifecycle events. This change enables complete observability through the standard receipt and event system.

**Discussion thread for this FIP: https://github.com/filecoin-project/FIPs/discussions/1149**


lanzafame (approved):
This looks excellent! Really useful for the ecosystem

FIPS/fip-implicit-receipts.md
```diff
@@ -0,0 +1,336 @@
...
+   Value:      0,
+   GasFeeCap:  0,
+   GasPremium: 0,
+   GasLimit:   buildconstants.BlockGasLimit * 10000, // 100 Trillion
+   Method:     cron.Methods.EpochTick,               // 2
+   Params:     nil,
+}
+```
+
+The extremely high gas limit (100 trillion) ensures that cron operations, which include critical network maintenance functions like sector expiration and deal completion, are never constrained by gas limitations.
```
jsoares, line 112: Is it worthwhile to clarify the distinction here, i.e. "never fail" above vs "never constrained" here?

jsoares (approved):
This is a significant improvement. The FIP is editorially sound, but I'll love the detailed technical review to others.

rvagg:
Taking FIP number 0107 for this.

rvagg:
@Kubuxu & @ZenGround0 if you wouldn't mind having a look at commit c54c2cc here regarding the `GasLimit`s that we're using for the implicit messages just to make sure my understanding is correct regarding how they are applied; i.e. `BlockGasLimit` doesn't have anything to do with these, it's simply during execution that we run down from the limit to `0`, so we have a high limit to start with.

Kubuxu:
Yes, IIRC, it was "essentially unlimited" gas limit
It is not counted against BlockGasLimit

PR #1159, opened 2025-05-29
Move FIP-0077 to Last Call

ipfsforcezuofu:
Update status of FIP 0077 Add Cost Opportunity For New Miner Creation from Draft to Last Call


rvagg (approved):
Fine by me, I haven't had time yet to review the builtin-actors PR but I know it's been through at least one round of reviews; I think the outstanding question here is simply whether we want to impose this on the miner actor or not, so it's worth opening it up for making a decision.

PR #1164, opened 2025-06-02
Update recommendations on F3 rebroadcast timeout increments

masih:
After the F3 activation on mainnet the network experienced a lack of progress at an instance that ultimately entered more than 20 rounds.

What we observed was that there were extended periods of total radio silence in the network where none of the participants were exchanging messages. This was caused by two things:
 1. the FIP defines lack of progress as no change in instance, round or phase, and
 2. the rebroadcast timeout is always incremented relative to phase timeout.

The increase in rounds increased the phase timeout exponentially, and rebroadcast never triggered until the phase timeout expired.

Update the FIP to:
 * adjust advice on arbitrary rebroadcast timeout increment
 * rebroadcast independent of phase timeout at high rounds

The changes made here specifically avoid updating the pseudocode to keep it abstract. Because, the rebroadcast interval is not a consensus-critical process and is largely at the discretion of implementers.

For reference, the go-f3 implementation of this FIP has opted to:
* await phase rebroadcast for the first 3 rounds before rebroadcasting any messages.
* beyond 3 rounds rebroadcast with exponential backoff without waiting for phase timeout to expire.
* cap the maximum exponential backoff for rebroadcast

See:
* https://github.com/filecoin-project/go-f3/issues/983
* https://github.com/filecoin-project/go-f3/pull/1003

Fixes: https://github.com/filecoin-project/go-f3/issues/1009

### DIFF
--- a/FIPS/fip-0086.md
+++ b/FIPS/fip-0086.md
@@ -432,7 +432,7 @@ GossiPBFT(instance, inputChain, baseChain, committee) → decision, certificate:
 01:  round ← 0
 02:  proposal ← inputChain  // Chain that the participant locally believes should be decided.
 03:  timeout ← 2*Δ
-04:  timeout_rebroadcast ← timeout + 1 // Any value larger than timeout, at implementation discretion.
+04:  timeout_rebroadcast ← timeout + 1 // Any value larger than the initial timeout, at implementation discretion.
 05:  value ← proposal // Used to communicate the voted value to others (proposal or 丄)
 06:  evidence ← nil   // Used to communicate evidence for the voted value
 07:  C ← {baseChain}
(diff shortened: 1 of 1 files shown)


rvagg (approved):
"arbitrary" sounds a bit squishy, but I guess we're in squishy-world now with this?

masih:
> arbitrary

Happy to use a different term. That term is already used in [a couple of places in the FIP](https://github.com/filecoin-project/FIPs/blob/8e2fcff7e31b0e09eec3b600f85bf533b30a7f18/FIPS/fip-0086.md?plain=1#L502), hence to choice to continue to use it in alterations here.

lanzafame (approved):

PR #1169, opened 2025-06-24
frc: f3-augmented-snapshot

hanabi1224:
We propose extending the Filecoin CAR snapshot with an F3 snapshot as a raw data block, and changing CAR roots to be a CID that points to a CBOR-encoded Filecoin snapshot header struct.

### DIFF
--- a/FRCs/frc-f3-augmented-snapshot.md
+++ b/FRCs/frc-f3-augmented-snapshot.md
@@ -1,0 +1,138 @@
+---
+fip: "####"
+title: "Augment Filecoin Snapshot with F3 data"
+author: "Hailong Mu (@hanabi1224)"
+discussions-to: https://github.com/filecoin-project/go-f3/issues/480
+status: Draft
+type: "FRC"
+created: 2025-06-25
+---
+
+# FRC-####: Augment Filecoin Snapshot with F3 data
+
+## Simple Summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the FIP.-->
+
+Downloading F3 finality certificates from scratch takes a long time and increases the p2p network bandwidth usage.
+An F3 snapshot is proposed to be included in the Filecoin CAR snapshot to reduce F3 catchup time and p2p network bandwidth usage on bootstrapping
+a Filecoin node with a Filecoin snapshot.
+
+## Abstract
+<!--A short (~200 words) description of the technical issue being addressed.-->
+
+We propose extending the Filecoin CAR snapshot with an F3 snapshot as a raw data block, and changing CAR roots to be a CID that points to a CBOR-encoded Filecoin snapshot header struct.
+
+## Motivation
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+
+The time cost and the network bandwidth usage for a new Filecoin node to catch up with all F3 finality certificates grow over time, which delays the readiness of the F3-aware V2 RPC APIs. By embedding an F3 snapshot into the current Filecoin CAR snapshot, both can be vastly reduced at the cost of a slightly increased Filecoin CAR snapshot size.
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any current Filecoin implementations. -->
+
+We propose the blow changes to the Filecoin CAR snapshot format.
+
+- Change CAR roots to be a CID that points to a CBOR-encoded [`SnapshotMetadata`](#snapshotmetadata) struct that is stored as the first data block in the CAR.
+- Store the raw [`F3Snapshot`](#f3snapshot) bytes as the second data block in the CAR when `F3Data != nil` in the metadata.
+
+### SnapshotMetadata
+
+```go
+type SnapshotMetadata {
+    HeadTipsetKey []Cid // required
+    F3Data        *Cid  // optional
+}
+```
+
+### F3Snapshot
+
+An F3 snapshot contains one header block and N(N>0) data blocks in the below format:
+
+`[Header block] [Data block] [Data block] [Data block] ...`
+
+A header block is a CBOR-encoded [`F3SnapshotHeader`](#f3snapshotheader) with a length prefix in the below format:
+
+`[varint-encoded length] [CBOR-encoded F3SnapshotHeader]`
+
+### F3SnapshotHeader
+
+```go
+type SnapshotHeader struct {
+	Version           uint64
+	FirstInstance     uint64
+	LatestInstance    uint64
+	InitialPowerTable gpbft.PowerEntries
+}
+```
+
+A data block is a CBOR-encoded [`FinalityCertificate`](#finalitycertificate) with a length prefix in the below format:
+
+`[varint-encoded length] [CBOR-encoded FinalityCertificate]`
+
+### FinalityCertificate
+
+```go
+// FinalityCertificate represents a single finalized GPBFT instance.
+type FinalityCertificate struct {
+	// The GPBFT instance to which this finality certificate corresponds.
+	GPBFTInstance uint64
+	// The ECChain finalized during this instance, starting with the last tipset finalized in
+	// the previous instance.
+	ECChain *gpbft.ECChain
+	// Additional data signed by the participants in this instance. Currently used to certify
+	// the power table used in the next instance.
+	SupplementalData gpbft.SupplementalData
+	// Indexes in the base power table of the certifiers (bitset)
+	Signers bitfield.BitField
+	// Aggregated signature of the certifiers
+	Signature []byte
+	// Changes between the power table used to validate this finality certificate and the power
+	// used to validate the next finality certificate. Sorted by ParticipantID, ascending.
+	PowerTableDelta PowerTableDiff `json:"PowerTableDelta,omitempty"`
+}
+```
+
+Notes:
+- `FinalityCertificate`s should be ordered by `GPBFTInstance` in ascending order, thus they can be validated and intermediate power tables can be generated while data blocks are being read in a stream.
+- The first and last `FinalityCertificate` instances should match those in the header, respectively.
+
+## Backwards Compatibility
+<!--All FIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The FIP must explain how the author proposes to deal with these incompatibilities. FIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+
+- A filecoin node should try to read a snapshot CAR in the proposed format, and fallback to the old format to maintain backward compatibility.
+- CLI options should remain unchanged to make it transparent to the node users.
+- The code change in all Filecoin nodes should be shipped with a network upgrade, and the Filecoin snapshot providers should only start publishing with the new format after the mainnet upgrade finishes to avoid potential errors during snapshot import for node users.
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for FIPs affecting consensus changes. Other FIPs can choose to include links to test cases if applicable.-->
+
+## Security Considerations
+<!--All FIPs must contain a section that discusses the security implications/considerations relevant to the proposed change. Include information that might be important for security discussions, surfaces risks and can be used throughout the life cycle of the proposal. E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. FIP submissions missing the "Security Considerations" section will be rejected. A FIP cannot proceed to status "Final" without a Security Considerations discussion deemed sufficient by the reviewers.-->
+
+This change has minimal security implications as the additional F3 data are also stored in the node database, unencrypted. Key considerations:
+
+- **Integrity**: The F3 snapshot can be validated.
+- **Performance** The F3 snapshot data blocks can be read, validated and imported in a stream.
+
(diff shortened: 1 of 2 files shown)


FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+fip: "####"
+title: "Augment Filecoin Snapshot with F3 data"
+author: "Hailong Mu (@hanabi1224)"
+discussions-to: https://github.com/filecoin-project/go-f3/issues/480
+status: Draft
+type: "FRC"
+created: 2025-06-25
+---
+
+# FRC-####: Augment Filecoin Snapshot with F3 data
```
BigLep, line 11: @rvagg : this is definitely something we're going to do and ship.  Any concern with assigning this FRC-0108?

BigLep, line 11:
```suggestion
# FRC-####: Filecoin Snapshot Format
```

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+`[varint-encoded length] [CBOR-encoded F3SnapshotHeader]`
+
+### F3SnapshotHeader
+
+```go
+type SnapshotHeader struct {
+	Version           uint64
+	FirstInstance     uint64
+	LatestInstance    uint64
+	InitialPowerTable gpbft.PowerEntries
```
BigLep, line 64: Do we define this structure anywhere?

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+
+We propose the blow changes to the Filecoin CAR snapshot format.
+
+- Change CAR roots to be a CID that points to a CBOR-encoded [`SnapshotMetadata`](#snapshotmetadata) struct that is stored as the first data block in the CAR.
+- Store the raw [`F3Snapshot`](#f3snapshot) bytes as the second data block in the CAR when `F3Data != nil` in the metadata.
+
+### SnapshotMetadata
+
+```go
+type SnapshotMetadata {
```
BigLep, line 41: Why don't we have a version field?

BigLep (commented):
Thanks for starting on this @hanabi1224 .  I know I wasn't really engaged in previous conversations, so apologies if any of my comments were covered during previous conversations.  Feel free to point me to Google Doc comments, etc.  These comments are coming from looking at this with fresh eyes.

My highest-level feeback items are:
1. I think this FRC's scope should be "Filecoin Snapshots" which happens to include 2 versions.  The focus should be on v2 (making snapshots F3 aware), but I think we need to specify enough of the old/v1 format.
2. Why are we doing all the "varint" stuff vs. just relying on cbor encoding of the overall "f3data"?
3. Why version the F3SnapshotHeader rather the SnapshotMetadata itself?  That seems like the clear place to allow future extensibility even outside the "F3data" area.

For points 2 and 3, I created a commit drafting some alternative language: https://github.com/filecoin-project/FIPs/pull/1170 .  I couldn't target my PR to your fork.  If you like where that is going, maybe you can take that commit on top of the work in your fork?

And again, apologies if I'm wildly off or misunderstanding of things here...

rvagg:
As per comment @ https://github.com/filecoin-project/FIPs/pull/1169/files#r2170162191, we'll assign this one number 108. @hanabi1224 would you mind updating the various places the number needs to appear? filename, README, frontmatter, title.

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+<!--The motivation is critical for FIPs that want to change the Filecoin protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the FIP solves. FIP submissions without sufficient motivation may be rejected outright.-->
+
+The time cost and the network bandwidth usage for a new Filecoin node to catch up with all F3 finality certificates grow over time, which delays the readiness of the F3-aware V2 RPC APIs. By embedding an F3 snapshot into the current Filecoin CAR snapshot, both can be vastly reduced at the cost of a slightly increased Filecoin CAR snapshot size.
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for any current Filecoin implementations. -->
+
+We propose the blow changes to the Filecoin CAR snapshot format.
+
+- Change CAR roots to be a CID that points to a CBOR-encoded [`SnapshotMetadata`](#snapshotmetadata) struct that is stored as the first data block in the CAR.
```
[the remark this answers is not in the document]
rvagg: Not technically wrong since it's plural, `roots` and can be an array. A pedantic edit could be:

```suggestion
- Change CAR header's `roots` array to contain a single CID that points to a CBOR-encoded [`SnapshotMetadata`](#snapshotmetadata) struct that is stored as the first data block in the CAR.
```

FRCs/frc-f3-augmented-snapshot.md
rvagg: Not critical though, versioning can just be done by breaking the format and then checking schemas -- does this thing decode as a V1 format? No, then try V2. Just as long as they are different enough, like having a different number of fields.

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+	// Additional data signed by the participants in this instance. Currently used to certify
+	// the power table used in the next instance.
+	SupplementalData gpbft.SupplementalData
+	// Indexes in the base power table of the certifiers (bitset)
+	Signers bitfield.BitField
+	// Aggregated signature of the certifiers
+	Signature []byte
+	// Changes between the power table used to validate this finality certificate and the power
+	// used to validate the next finality certificate. Sorted by ParticipantID, ascending.
+	PowerTableDelta PowerTableDiff `json:"PowerTableDelta,omitempty"`
```
rvagg, line 91: worth defining, I think https://github.com/filecoin-project/go-f3/blob/d713c5a5ec94f0e6af7736f5c0b4bae625496acf/certs/certs.go#L18-L31 is relevant

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+// FinalityCertificate represents a single finalized GPBFT instance.
+type FinalityCertificate struct {
+	// The GPBFT instance to which this finality certificate corresponds.
+	GPBFTInstance uint64
+	// The ECChain finalized during this instance, starting with the last tipset finalized in
+	// the previous instance.
+	ECChain *gpbft.ECChain
+	// Additional data signed by the participants in this instance. Currently used to certify
+	// the power table used in the next instance.
+	SupplementalData gpbft.SupplementalData
```
rvagg, line 84: worth defining or linking

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+Notes:
+- `FinalityCertificate`s should be ordered by `GPBFTInstance` in ascending order, thus they can be validated and intermediate power tables can be generated while data blocks are being read in a stream.
+- The first and last `FinalityCertificate` instances should match those in the header, respectively.
+
+## Backwards Compatibility
+<!--All FIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The FIP must explain how the author proposes to deal with these incompatibilities. FIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+
+- A filecoin node should try to read a snapshot CAR in the proposed format, and fallback to the old format to maintain backward compatibility.
```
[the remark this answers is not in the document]
rvagg:
```suggestion
- A Filecoin node should try to read a snapshot CAR in the proposed format. A failure to successfully decode the blocked referenced as as the CAR's single root using the schema presented above, a snapshot reader may fallback to the old format and maintain backward compatibility. Additional failures to decode original snapshot format would indicate a fatal error.
```

FRCs/frc-f3-augmented-snapshot.md
```diff
@@ -0,0 +1,138 @@
...
+
+A header block is a CBOR-encoded [`F3SnapshotHeader`](#f3snapshotheader) with a length prefix in the below format:
+
+`[varint-encoded length] [CBOR-encoded F3SnapshotHeader]`
+
+### F3SnapshotHeader
+
+```go
+type SnapshotHeader struct {
+	Version           uint64
```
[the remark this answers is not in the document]
hanabi1224: That's a great idea

hanabi1224:
> Why are we doing all the "varint" stuff vs. just relying on cbor encoding of the overall "f3data"?

@BigLep we proposed the CAR-like block structure to facilitate stream reading from and writing to the snapshot. The current CBOR-gen tool seems to require holding all the blocks in memory in both reading and writing, please let me know if this is not a problem and it's be great to point me to the best CBOR package that supports stream reading and writing

BigLep:
> > Why are we doing all the "varint" stuff vs. just relying on cbor encoding of the overall "f3data"?
> 
> @BigLep we proposed the CAR-like block structure to facilitate reading block streams from and writing block streams to the snapshot, with low RAM requirement. The current CBOR-gen tool seems to require holding all the blocks in memory for both reading and writing, please let me know if this is not a problem and it'd be great to point me to the best CBOR package that supports stream reading and writing

ACK - got it.  I don't know CBOR libraries and which ones support streaming.  I had done a quick search to see that it's possible, but I didn't investigate it beyond that and whether it applies in our context.

If we aren't packaging the whole thing in CBOR (fine for the reasons discussed) the rationale of this design decision seems important to capture in this document.

hanabi1224:
> I wonder if it's worth either clarifying that this is what we're talking about ("with varint-encoded length prefix as per the CAR format"), or even just remove mention of the prefixes entirely since it's really just an implementation detail of the CAR format that we don't need to duplicate here.

Clarified that the F3 snapshot is in a CARv1-like format but not CAR. All prefixes thing mentioned in the FRC spec details of the F3 snapshot format

> how do I distinguish? 

The F3 data CID is defined in the `SnapshotMetadata` to distinguish from the chain IPLD blocks. The F3 data block should be just delegated to the F3 package for importing. A draft implementation: https://github.com/filecoin-project/go-f3/pull/1032

FRCs/frc-f3-augmented-snapshot.md
hanabi1224: Added to the FRC

FRCs/frc-0108.md
```diff
@@ -0,0 +1,244 @@
...
+
+Node users should experience faster F3 bootstrapping time and less network bandwidth usage.
+
+## Product Considerations
+<!--All FIPs must contain a section that discusses the product implications/considerations relative to the proposed change. Include information that might be important for product discussion. A discussion on how the proposed change will enable better storage-related goods and services to be developed on Filecoin. FIP submissions missing the "Product Considerations" section will be rejected. An FIP cannot proceed to status "Final" without a Product Considerations discussion deemed sufficient by the reviewers.-->
+
+Nodes starting from a snapshot should not rely on the certificate exchange protocol to catch up with the F3 data because we expect this will get slower over time. A slow F3 catchup time leads to, e.g.
+
+- delay in the readiness of F3-aware RPC APIs
```
rvagg, line 233:
```suggestion
### Start-up without initial F3 data

Nodes starting from a snapshot should not rely on the certificate exchange protocol to catch up with the F3 data because we expect this will get slower over time. One outcome of a slow F3 catchup time is a delay in the readiness of F3-aware RPC APIs.

### CAR format expectations

This change introduces a relatively novel use of the CAR format in that it contains one very large block, much larger than typical blocks found in most CAR containers for use with IPLD data. At the time of this proposal, this block size would be approximately TODO MiB and this will only grow over time. While this is not disallowed by the CAR specification, many CAR processing utilities are built on an assumption of classic IPFS style blocks of more more than approximately 1MiB each. Some CAR tooling may struggle to deal with the new proposed format, although handling CAR data outside of the narrow use-case of snapshort imports on Filecoin nodes is not typical or necessarily recommended.
```

See TODO in there, I don't know what size we're talking about here, do you have a rough guestimate to get us started? If it's too hard we could remove the specific reference but I wanted to provide a ballpark to point out how it's different to the max ~1MiB expectation that a lot of tools expect today. But also be clear that this isn't necessarily a bad thing, just bad expectations in existing software.

rvagg:
This reads really well now and it's pretty clear. Thanks for the clarification of "CARv1-like".
I've left some suggestions inline but otherwise this seems good to merge to me.

BigLep (commented):
Thanks @hanabi1224 for the quick turn around!

FYI that I did a few small cosmetic things in https://github.com/filecoin-project/FIPs/pull/1169/commits/7dac2e23cfa546a4a11975ebe9aefe9a780cdf4d .  I moved the V1 specification before the V2 specification and I did more section linking.

I also documented my understanding of some of the design considerations/decisions in https://github.com/filecoin-project/FIPs/pull/1169/commits/11a72919ead87e18d9b00046a682346b89cfd405.

(Please change anything I have wrong.  I'm fully expecting these commits to get squashed or removed...)

Concerning versioning `F3SnapshotHeader` as well, that seems fine to add as well.

hanabi1224:
> FYI that I did a few small cosmetic things

@BigLep Thanks a lot!

> Concerning versioning `F3SnapshotHeader` as well,

@BigLep design concerns are added in [commit](https://github.com/filecoin-project/FIPs/pull/1169/commits/8db56b386470afaf4b5b44c281899193012b90b5#diff-9180f2619c41417eb7d2f503718a4ea8ffbbf6ddd5eb6401ebfc58824e9540b4R91).

BigLep:
Looks good to me @hanabi1224 .  Once we get the official approval from @rvagg , we can work to get it merged

rvagg (approved):
lgtm!

PR #1200, opened 2025-09-27
fix(fip-55): f410 binary encoding

wjmelements:
Yes this FIP is final but it has a minor mistake worth fixing.

The `0x` prefix signals hexadecimal. This portion of the address binary is 10 in decimal which is 0x0a. The current FIP says 0x10 which is incorrect. I lost some time debugging this today.


Stebalien (approved):
This fix is correct. 10 here is in decimal and should be:

1. Varint encoded (a no-op here because 10 fits in a single varint byte).
2. Converted to hex.

PR #1209, opened 2025-10-31
FIPs: Add draft for EIP-7702 (Set Code for EOAs)

snissn:
EIP-7702 defines a new EIP-2718 transaction type (`0x04`) that includes an `authorization_list`. Each authorization is a signed tuple from an EOA (the "authority") designating a smart contract (the "delegate") to execute on its behalf.

This FIP adapts EIP-7702 to the FEVM architecture. The implementation introduces a new method, `ApplyAndCall`, in the EVM actor. This method atomically processes the authorization list—validating signatures, updating the delegation mapping, and incrementing authority nonces—and then executes the outer transaction call. The FEVM interpreter is updated to recognize these delegations: when a `CALL` targets a delegated EOA, the interpreter executes the delegate's code within the authority's context. This implementation ensures that delegation mappings and nonce increments persist even if the outer transaction call reverts, adhering to the EIP's specified atomicity model.

**Discussion:** https://github.com/filecoin-project/FIPs/discussions/1143

### DIFF
--- a/FIPS/fip-0111.md
+++ b/FIPS/fip-0111.md
@@ -1,5 +1,5 @@
 ---
-fip: "\<to be assigned\>"
+fip: "0111"
 title: Add Support for EIP-7702 (Set Code for EOAs) in the FEVM
 author: "Michael Seiler (@snissn), Aarav Mehta (@aaravm)"
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/1143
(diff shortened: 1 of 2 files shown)


rvagg:
@snissn let's assign this number `0111` - please update the filename, title, frontmatter and README link.

snissn:
thanks @rvagg updated here: 032ffea

jsoares (approved):

PR #1218, opened 2025-11-18
Add draft FIP for tipset gas reservations

snissn:
This FIP introduces tipset-scope gas reservations as a consensus rule for Filecoin to eliminate miner exposure to underfunded gas charges within a single tipset while preserving existing receipts and gas accounting. For each tipset, nodes derive the canonical set of explicit messages and construct a per-sender reservation plan equal to the sum, over the tipset, of each message’s `gas_limit * gas_fee_cap`. The execution engine maintains an internal reservation session ledger keyed by actor identifier; when a session is open, each sender’s free balance is defined as on-chain balance minus the remaining reserved amount recorded in this ledger.

https://github.com/filecoin-project/FIPs/discussions/1210

### DIFF
--- a/FIPS/fip-draft_tipset-reservations.md
+++ b/FIPS/fip-draft_tipset-reservations.md
@@ -1,0 +1,383 @@
+---
+fip: "<to be assigned>"
+title: Multi-Stage Tipset Gas Reservations
+author: "Michael Seiler (@snissn)"
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/1210
+status: Draft
+type: Technical (Core)
+category: Core
+created: 2025-11-17
+spec-sections:
+  - section-systems.filecoin_vm.gas_fee
+  - section-systems.filecoin_vm.runtime.receipts
+  - section-systems.filecoin_vm.interpreter.gas-payments
+  - section-systems.filecoin_vm.interpreter.duplicate-messages
+---
+
+## Simple Summary
+
+This FIP introduces multi-stage gas reservations across each Filecoin tipset. Before executing user messages, block-producing nodes reserve enough gas funds to cover each sender’s maximum possible gas spend across all of their messages in the tipset. Messages still execute in the same order and produce the same receipts as today, but gas charges are drawn from these reservations instead of depending solely on the sender’s live balance at execution time. This removes a class of intra-tipset “self-drain” scenarios where underfunded messages would otherwise fall back to miner penalties.
+
+## Abstract
+
+This FIP introduces tipset-scope gas reservations as a consensus rule for Filecoin to eliminate miner exposure to underfunded gas charges within a single tipset while preserving existing receipts and gas accounting. For each tipset, nodes derive the canonical set of explicit messages and construct a per-sender reservation plan equal to the sum, over the tipset, of each message’s `gas_limit * gas_fee_cap`. The execution engine maintains an internal reservation session ledger keyed by actor identifier; when a session is open, each sender’s free balance is defined as on-chain balance minus the remaining reserved amount recorded in this ledger.
+
+Messages execute in the same order and with the same gas metering and `GasOutputs` computation as today. Preflight checks assert that reserved funds cover each message’s maximum gas cost, and all value transfers are enforced against free balance rather than raw balance, so a sender cannot move away funds needed to pay for reserved gas. Settlement is rewritten so that, for each message, the engine net-charges the sender only for actual consumption (base-fee burn, overestimation burn, and miner tip) while releasing the full reserved amount, realizing refunds by increasing free balance instead of issuing explicit refund transfers. The reservation ledger is tipset-local and requires no new on-chain actors or state migrations. The new rules activate at network version 28 via a chain upgrade, with pre-activation feature gating allowing best-effort and strict modes while different client implementations roll out support.
+
+## Change Motivation
+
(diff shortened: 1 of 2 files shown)


snissn:
@ZenGround0

snissn:
@ZenGround0 

Thanks for the detailed analysis. You are absolutely correct that the previous "blind aggregation" model created an asymmetric attack vector where valid individual blocks could combine into an invalid tipset, effectively allowing a low-balance attacker to invalidate blocks and deny miner rewards.

We have implemented the fix you suggested: **Strict Sender Partitioning**.

### The Fix: Strict Sender Partitioning
We have updated the consensus logic (`ApplyBlocks` and `buildReservationPlan`) to enforce a strict rule:
**"If a sender's messages have been executed in a higher-precedence block within the current tipset, all messages from that same sender in any lower-precedence blocks are strictly ignored."**

This changes the behavior from "Merge and Execute" to "First-Block-Takes-All" for each sender identity.

### Why this resolves the security concerns
1.  **Eliminates Invalid Tipsets:**
    *   Under the old model, `Reserve(Block A) + Reserve(Block B)` could exceed `Balance`.
    *   Under the new model, the reservation plan logic mirrors the execution logic: if Block A is processed first, `Reserve(Sender)` is derived *solely* from Block A. The messages in Block B are dropped from both the reservation plan and the execution set.
    *   Therefore, if Block A is valid (affordable) and Block B is valid (affordable), the tipset is guaranteed to be valid because the aggregate cost for the sender never exceeds the cost of the winning block.

2.  **Neutralizes EIP-7702/AA Complexity:**
    *   The partitioning is based on the static `From` address in the message header.
    *   The reservation check happens *before* any execution.
    *   This creates a firewall where complex execution logic (account abstraction, internal sends) cannot retroactively invalidate the gas reservation or the tipset structure.

### The Trade-off: Partial Orphan Spam
We acknowledge the trade-off you mentioned: "no-one is paying for the gas to include messages on chain" for the ignored messages in lower-precedence blocks.
*   This is acceptable because it shifts the failure mode from **Consensus Failure** (invalid tipset) to **Miner Bandwidth Waste** (partial orphan).
*   It aligns with existing Filecoin incentives where miners already risk including duplicate messages that yield no reward.
*   Attacking this vector is not free; the attacker must still pay for the messages in the winning block to trigger the exclusion of the losing block.

This change is now implemented in the `multistage-execution` branch and documented in the FIP draft.

ZenGround0:
So I think the above is not correct.  Basefee calculation relies on total gas limit of messages.  So the above attack is possible with todays rules by using two messages with the same nonce and running the attack to pack the lower precedence blocks with a message with gas limit 100x that of the message sent to the highest precedence block.  

Overall I don't see an increase to the basefee manipulation attack surface beyond the situation today.  The same thing roughly applies to wasted message tree space as the duplicated message can be packed with large parameters.

@Kubuxu does this make sense?  Are there other scenarios I'm missing that make one-sender-per-block impact base fee manipulation leverage?

Kubuxu:
> miner pays the penalty for the M2 message 

This is not true, the same nonce messages don't get counted or penalised.

snissn:
Thanks for your analysis @Kubuxu! 

Is there anything you would suggest we change? 
Does this seem reasonable to you as it is currently?

FIPS/fip-draft_tipset-reservations.md
```diff
@@ -0,0 +1,406 @@
...
+
+Let `B_1, B_2, ..., B_n` be the blocks in `T`, ordered by ticket precedence (where `B_1` has the highest precedence/lowest ticket).
+Let `S_seen` be the set of senders whose messages have been accepted in previous blocks, initially empty.
+Let `M_seen` be the set of message CIDs accepted so far, initially empty.
+
+`M(T)` is constructed by iterating through `B_1` to `B_n`:
+- For each block `B_i`:
+  - Identify the set of candidate messages `C_i` in `B_i` (excluding messages with CIDs already in `M_seen`, per legacy deduplication rules).
+  - For each message `m` in `C_i`:
+    - If `from(m)` is in `S_seen`, the message `m` is **excluded** (it is ignored for execution and reservation purposes).
```
rvagg: "ignored for execution and reservation purposes" - do we still get a receipt for this though? does this need to be noted in the backward compatibility section, or am I missing that?

rvagg:
@snissn some review items:

* Missing "Incentive Considerations", "Product Considerations", "Implementation", "TODO" and "Copyright" sections.
  - Incentive and/or Product Considerations sections should probably address: SP concerns regarding revenue changes, mempool changes (discussed above), tooling changes needed to adapt to new execution/selection rules? Do block explorers or other chain analysis tools need to update to understand the new rules?
* The "Specification" is _really_ dense and it makes it hard to review, I've not been thorough in understanding it by any means but I see a lot of duplication in there which leads me to believe it's far more dense than it needs to be and being more concise and terse in many places would make this more straightforward to review. For example I see a bunch of invariants repeated in many places, `reserved[s] ≥ gas_cost(m)`, `consumption + refund = gas_cost` (one with `==`), and so on. `reserved[a] == 0` at "end" is repeated a few times too.
* "Design Rationale" seems to largely reframe/restate the spec?
* The "Backwards Compatibility" rehashes a lot of rules already covered in a very verbose way, can we be more selective, or just highlight useful items that specifically relate to backward compatibility or drop them entirely if it's not really a concern? 

tbh my concern is mostly the verbosity and repetition, cause it makes it very hard to review on tecnical merits, at least for me, but I'm guessing for others too.

---

Separately, what's the status of this FIP from your perspective? Does it depend strictly on a 7702 FIP landing at the same time (if so we should call it out more clearly as I mentioned), so therefore depends on help getting 7702 over the line and is essentially stalled until we collectively decide to make it happen? I'm wondering about whether we should assign a number for this but that depends on the liklihood of getting this landed any time soon. Having numbers in limbo for open FIPs makes the numbering hard to manage. It's fine if they're near-merge ready, or the path to merge is clear. It's not with this one, at least to me, so what do you think?

jsoares:
What's the status here? Is it awaiting editorial review or are there still ongoing discussions?

PR #1248, opened 2026-03-30
FRC-0116: JSON-RPC HTTP conventions

LesnyRumcajs:
Proposal on unifying the HTTP response code conventions with the JSON-RPC API.

[Discussions thread](https://github.com/filecoin-project/FIPs/discussions/1247)

### DIFF
--- a/FRCs/frc-0116.md
+++ b/FRCs/frc-0116.md
@@ -98,7 +98,6 @@ However, the pragmatic benefits of mapping errors to HTTP status codes outweigh 
 
 1. **Lightweight error detection.** An HTTP client can determine that an error occurred from the status code alone, without parsing the response body. This is especially valuable when the response body is large.
 2. **Infrastructure compatibility.** Proxies, load balancers, CDNs, and monitoring tools are built around HTTP status codes. Returning 200 for errors makes these tools blind to failure conditions.
-3. **Precedent.** The [JSON-RPC 2.0 Transport: HTTP](https://www.simple-is-better.org/json-rpc/transport_http.html) community draft defines this mapping. While not an official standard, it represents the most widely referenced specification for JSON-RPC over HTTP.
 
 ### Why not follow the Ethereum ecosystem?
 
--- a/FRCs/frc-0116.md
+++ b/FRCs/frc-0116.md
@@ -32,7 +32,7 @@ This FRC specifies the HTTP transport conventions for Filecoin JSON-RPC endpoint
 
 The key words "MUST", "MUST NOT", "SHOULD", "SHOULD NOT", and "MAY" in this document are to be interpreted as described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119).
 
-This specification is based on the [JSON-RPC 2.0 Transport: HTTP](https://www.simple-is-better.org/json-rpc/transport_http.html) community draft.
+This specification is derived from the [JSON-RPC 2.0 Transport: HTTP](https://www.simple-is-better.org/json-rpc/transport_http.html) community draft, with modifications to suit the Filecoin ecosystem.
 
 ### HTTP Method Requirements
 
--- a/FRCs/frc-0116.md
+++ b/FRCs/frc-0116.md
@@ -1,5 +1,5 @@
 ---
-fip: "XXXX"
+fip: "0116"
 title: JSON-RPC 2.0 HTTP Transport Conventions
 author: Hubert Bugaj (@LesnyRumcajs)
 discussions-to: https://github.com/filecoin-project/FIPs/discussions/1247
--- a/FRCs/frc-0116.md
+++ b/FRCs/frc-0116.md
@@ -8,7 +8,7 @@ type: FRC
 created: 2026-03-30
 ---
 
-# FRC-XXXX: JSON-RPC 2.0 HTTP Transport Conventions
+# FRC-0116: JSON-RPC 2.0 HTTP Transport Conventions
 
 ## Simple Summary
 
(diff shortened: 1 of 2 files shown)


jennijuju (approved):
editorial review ✅  (v well written!

FRCs/frc-draft_jsonrpc_http_transport.md
```diff
@@ -0,0 +1,166 @@
+---
+fip: "XXXX"
```
jennijuju: plz take 116!

rvagg:
I should have commented in the discussion so I moved my text there: https://github.com/filecoin-project/FIPs/discussions/1247#discussioncomment-16421365

<details>
<summary>original comment</summary>

I introduced -32005 in https://github.com/filecoin-project/lotus/pull/13561, with the intention of trying to align more with the Ethereum ecosystem on jsonrpc patterns, at least for our Ethereum calls. In this case, it aligns with [EIP-1474](https://eips.ethereum.org/EIPS/eip-1474). Unfortunately this is not a uniformly adopted EIP so Ethereum RPC providers are all over the place with this. But if you look at the codes described there they are not all application logic problems and don't belong as `500`s. Currently the majority of error codes Lotus returns are going to be `200`'s, and I'd read that as "valid json in response, deal with it". A quick survey of the various public Ethereum nodes show that a "range exceeded", while not universally -32005, are universally http 200 responses (e.g. Cloudflare, PublicNode, 1RPC, and Merkle all return HTTP 200 for block range exceeded errors), parse errors are the one place where some providers return 400. From what I can tell, 200-for-everything is Ethereum standard. My 2c is that I'd be happy with mostly 200, maybe some 500's for some "something went badly wrong" and 4xx's for "not here" type errors.

One challenge for Go is that the go-jsonrpc client rejects non-200/400/500 responses entirely, not even parsing the body, it doesn't even accept them as expected & valid. So we'd have to introduce a change there and cross our fingers that every Go consumer application (mainly miner software) updates to that before we go about introducing new error codes.

How about 200 for all handler/application errors (matching Ethereum ecosystem practice), 400 for malformed requests (-32700, -32600, -32602), and 500 only for genuine server failures (-32603)?
</details>

FRCs/frc-0116.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Why map errors to HTTP status codes?
+
+The alternative approach — returning HTTP 200 for all JSON-RPC responses, including errors — maintains a clean separation between the transport layer (HTTP) and the application layer (JSON-RPC). Some implementations and libraries (e.g., Parity's `jsonrpsee`) follow this convention, and it is a defensible design choice.
+
+However, the pragmatic benefits of mapping errors to HTTP status codes outweigh the theoretical purity of layer separation:
+
+1. **Lightweight error detection.** An HTTP client can determine that an error occurred from the status code alone, without parsing the response body. This is especially valuable when the response body is large.
+2. **Infrastructure compatibility.** Proxies, load balancers, CDNs, and monitoring tools are built around HTTP status codes. Returning 200 for errors makes these tools blind to failure conditions.
+3. **Precedent.** The [JSON-RPC 2.0 Transport: HTTP](https://www.simple-is-better.org/json-rpc/transport_http.html) community draft defines this mapping. While not an official standard, it represents the most widely referenced specification for JSON-RPC over HTTP.
```
jsoares: This does not read accurate. The linked draft specifically states:

```
200 OK
for responses (both for Response and Error objects)
```

And provides an error example for an invalid request that returns status 200 and error code -32600:

```
HTTP/1.1 200 OK
Connection: close
Content-Length: ...
Content-Type: application/json
Date: Sat, 08 Jul 2006 12:04:08 GMT

{
    "jsonrpc": "2.0",
    "error": {
        "code": -32600,
        "message": "Invalid Request.",
        "data": "'method' is missing"
        },
    "id": 6
    }
}
```

This is at odds with the behaviour define in this FRC, where an invalid request would return 400 / -32600. 

I'm not commenting on the merits of either approach, just noting that the reference does not support the argument. @rvagg made a broader argument with respect to the use of 200 codes, which I think it worth considering despite being, IMHO, of a substantive (i.e. FIP discussion) rather than editorial (i.e. FIP review) nature.

jsoares (approved):
The goal is standardising response codes across implementations is a valuable one. People can disagree on the ideal standard, but this proposal is well-drafted and internally consistent, so 👍 on advancing.

FRCs/frc-0116.md
LesnyRumcajs: Yes, it's not 1:1. The original draft of this FRC was more in line with https://www.simple-is-better.org/json-rpc/transport_http.html. I'll continue the discussion around actual behaviour in the Discussions, but for now I reworded this to match the current status.

> This specification is derived from the [JSON-RPC 2.0 Transport: HTTP](https://www.simple-is-better.org/json-rpc/transport_http.html) community draft, with modifications to suit the Filecoin ecosystem.


Does this work?

jsoares: I think the reference above in the Specification section, line 35, is perfect. Your changes made it better, but it was fine as it was.

My comment was strictly on referencing the document in the `Why map errors to HTTP status codes?` section, in the context of explaining why you're proposing this mapping instead of the alternative of using 200 status code for all JSON-RPC responses -- when that's arguably what the document prescribes (kind of, depending on your definition of `all`). 

So my suggestion would be to remove bullet item nr 3 entirely as tangential to the argument.

LesnyRumcajs: I see, this makes sense. Thanks for flagging. Removed it as per your suggestion.

FRCs/frc-0116.md
```diff
@@ -0,0 +1,165 @@
...
+fip: "0116"
+title: JSON-RPC 2.0 HTTP Transport Conventions
+author: Hubert Bugaj (@LesnyRumcajs)
+discussions-to: https://github.com/filecoin-project/FIPs/discussions/1247
+status: Draft
+type: FRC
+created: 2026-03-30
+---
+
+# FRC-XXXX: JSON-RPC 2.0 HTTP Transport Conventions
```
jsoares:
```suggestion
# FRC-0116: JSON-RPC 2.0 HTTP Transport Conventions
```

jsoares:
The FRC already has a number and two approving reviews. While the process for FRCs has always been a bit more malleable, I'm fine with merging if no forthcoming changes.

cc @rvagg given previous thoughts expressed in discussion

PR #1250, opened 2026-03-31
Update FIP-0112: reorder SectorStatusCode enum

wjmelements:
Reviewer @zenground0
This documents a pending change in builtin-actors.
I suggested this change, and it will be likely merged in https://github.com/filecoin-project/builtin-actors/pull/1720
The reason to change it is that Dead, especially Never Committed, makes sense for a default value, and we should use falsy values for defaults.
#### Changes
* reorder SectorStatusCode enum


rvagg (approved):
looks like that got integrated

jsoares (approved):

PR #1251, opened 2026-03-31
Update FIP-0115: document unchanged behavior for null rounds

wjmelements:
Reviewer @rvagg
I have verified that this behavior is not changing in any of the implementations,
but I think this detail should be documented in the FIP since it would otherwise be ambiguous.
#### Changes
* mention that null round base fee behavior is unchanged


LinZexiao (approved):

rvagg (approved):
@TippyFlitsUK @jennijuju @jsoares ping for review to merge pls

jsoares (approved):

PR #1259, opened 2026-04-14
Fixing up old error model

ZenGround0:
@wjmelements gave some good feedback [here](https://github.com/filecoin-project/builtin-actors/pull/1720#issuecomment-4165800650) on how this text has drifted from the implementation as we've cleaned up the error semantics.  This PR brings the FIP text in line with the latest implementation


FIPS/fip-0112.md
```diff
@@ -73,15 +73,17 @@ pub struct GenerateSectorLocationReturn {
...
+This method takes in a sector number, generic serialized bytes encoding a sector location and the status of the sector, either `Active`, `Dead`, or `Faulty`.  It returns true if the correct sector status is passed.  Note that the sector location input parameter is specified to be a generic upgradeable byte slice format. The first implementation should use the standard tuple based cbor encoding of the `SectorLocation` type, i.e. `[i64, i64]`.  Note that the three possible statuses form a partition over all sectors, i.e. for each sector there is one and only one valid status.  Further note some relevant terminology from builtin-actors: `live` sectors are those that have not been terminated, including `faulty` sectors and `active` sectors which are not faulty.  `dead` sectors are either terminated or never activated and are therefore never `faulty`.
```
wjmelements, line 76: > Note that the three possible statuses form a partition over all sectors

I wouldn't reuse the word partition here since it's also used in the next paragraph in a different way. Here I think you intend to convey that every sector has exactly one status. Alternative ways to convey this:

* trifurcate
* tripartite

or you can drop the first statement and just say

> for each sector there is exactly one valid status

FIPS/fip-0112.md
```diff
@@ -73,15 +73,17 @@ pub struct GenerateSectorLocationReturn {
...
+For all statuses this method uses the sector location to look up the sector's partition and determine if the `sectors` bitfield includes the specified sector number, i.e. if this is the true location of the sector.  If the sector is found at this location the method returns true if the partition status matches the input sector status and false otherwise.  If the sector is not at the declared location the method reverts with `USR_NOT_FOUND` error. 
 
-When validating a `Dead` sector with a partition and deadline index >= 0 this method performs the same lookup and verifies that the sector number is present in the `Terminated` bitfield.  If the sector is not found at the provided index the method errors.  If the the partition and deadline index are NO_PARTITION and NO_DEADLINE then the method verifies that the sector number is not assigned to any member of the miner actor's sectors AMT.
+To validating a declared `Active` sector this method checks that the `terminated` and `faults` bitfield do not include the sector number.
```
wjmelements, line 80:
```suggestion
To validate an `Active` sector, this method checks that the `terminated` and `faults` bitfields do not include the sector number.
```

FIPS/fip-0112.md
```diff
@@ -73,15 +73,17 @@ pub struct GenerateSectorLocationReturn {
...
+
+When validating a `Dead` sector this method verifies that the sector number is present in the `terminated` bitfield.
```
wjmelements, line 82: I thought dead is also a default value, but a sector that never existed would not be present in the bitfield.

wjmelements: > `dead` sectors are either terminated or never activated

so I think this paragraph should describe how to validate never activated

FIPS/fip-0112.md
```diff
@@ -73,15 +73,17 @@ pub struct GenerateSectorLocationReturn {
...
 
-There are edge cases around NO_PARTITION, NO_DEADLINE.  If only one of these is provided this is an invalid sector location error.  When sector numbers are not found matching a sector location this is a not found error.  If `(NO_PARTITION, NO_DEADLINE)` is provided as a location this location is always "found".  If it is passed as a location to validate `Active` or `Faulty` statuses these are trivially never found in this location and the validation method returns false.
+If only one of NO_DEADLINE, NO_PARTITION are provided within the sector location this is an invalid sector location error.  When both NO_DEADLINE, NO_PARTITION are provided, this "null" location behaves exactly as any other sector location.  If the sector is not found at this null location, i.e. it is found to exist in the sectors AMT, then the method returns a NOT_FOUND error.  If a sector is found in this location, i.e. it does not exist in the sectors AMT, then the only valid status is `Dead`.  Validation `Live` and `Faulty` statuses at this location will return false.
```
rvagg, line 86: Catching up here and this is not a critique but a question - what is the purpose of handling the `NO_DEADLINE` + `NO_PARTITION` case here? Is there an important case that this solves? It looks to me like a potential footgun here if I blindly take from get-location and pass that to here and only `Dead` works?

wjmelements: I'm not sure I understand the question, but `NO_DEADLINE` and `NO_PARTITION` are the correct sector location for `Dead` only after `CompactPartitions`; before, it's still in the sectors AMT.

rvagg: Ahhh! ok that's actually a very interesting detail and I'm not sure I see that in the FIP text. This comment is on the inputs to ValidateSectorStatus, but I don't see anything on GenerateSectorLocation that explains how they both could be -1 and what might lead to that. Could we add that little detail in there to be clear how we might arive at -1 for both? Any other way at ariving at -1 for either would also be interesting.

PR #1260, opened 2026-04-16
Update fip-0115.md to Accepted Status

ThatGuyLLC:
Updatign status to accepted for FIP 115


rvagg (approved):
@ThatGuyLLC it would be really good for these status change PRs to reference some source or at least mention an event where this happened and was agreed upon. My understanding is that this was in the last core-devs, so a link to those meeting notes or even just a mention of the date of such a meeting would be helpful for provenance in future.

ThatGuyLLC:
Correct, sorry forgot to drop link: https://github.com/filecoin-project/core-devs/blob/master/Core%20Dev%20Meetings/Meeting%200090.md

lanzafame (approved):

PR #1267, opened 2026-06-23
fip-0107: GasLimit=0 for canonical implicit msgs

rvagg:
The current numbers encoded in the original form of this FIP come out of Lotus' construction of the messages. But internally Lotus overwrites those values before execution anyway. So I'm taking this opportunity to just zero them out in their canonical form for the purpose of making deterministic CIDs. It's effectively "uncapped".

Ref: https://github.com/filecoin-project/lotus/blob/25a03f58bff5491f047b5c16eedad2d434f33158/chain/vm/fvm.go#L484


lanzafame (approved):

jennijuju (approved):
LGTM

PR #1268, opened 2026-06-25
fip-0107: clarify return data links

LesnyRumcajs:
Clarify on return data links in FIP-0107.

### DIFF
--- a/FIPS/fip-0107.md
+++ b/FIPS/fip-0107.md
@@ -142,7 +142,7 @@ Messages will be ordered as they currently are for tipset execution, and this or
 
 The FVM will maintain codec information from the execution return value and this will be included in the `MessageReceipt`. This allows external systems to properly decode the return data without requiring prior knowledge of the actor implementation.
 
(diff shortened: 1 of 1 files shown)


rvagg (approved):
From a discussion in Slack. For the interested observer here's my 2c from there:

> We allow raw, cbor and dag-cbor codecs in fvm. In practice we don’t use raw for return values although it does appear in a few places that are not returns. We use cbor almost everywhere but there’s one place we use dag-cbor which is an evm actor’s GetBytecode method (3). In that case, we return an Option<Cid> https://github.com/filecoin-project/builtin-actors/blob/ecb889fcded2ededa70bf5fccca0b5e5c8f33b08/actors/evm/src/lib.rs#L288-L299; so it’s just a block that encodes a nullable CID. That CID takes you to a raw block that’s got the actor’s bytecode.
>
> BUT for the purpose of 0107, we’re not creating CIDs here, we’re just showing metadata. The Return is just a byte string, and the codec is listed as an interesting fact that may or may not be of interest to parties reading it. There’s no way in which Return would automatically be decoded according to the codec unless someone wires up a decoder that cares about the codec.

Clarifying it for anyone looking at this struct is probably worthwhile, don't treat this codec as if it were an inline CID with more data to explore.

lanzafame (approved):

jennijuju (approved):

PR #1269, opened 2026-06-26
FIP-0117: Repeated SnapDeals with Piece-Level Integrity

lucaniz:
SnapDeals (FIP-0019) lets a storage provider (SP) embed deal data into a CC-sector replica exactly once, by adding a randomized, unpredictable version of the data, and prove it in one message. This FIP makes that update repeatable ("Re-Snap") so an SP can swap data into and out of an already-snapped sector over its lifetime, reusing committed capacity rather than re-sealing.

Repetition raises an integrity requirement the one-shot case did not have: when an SP updates a sector holding several pieces and preserves a subset, a verifier must be able to confirm those pieces survived unchanged. This FIP specifies that mechanism so the check can run on-chain. It records an opt-in commitment to the sector's unsealed data (CommD), stored in a new SectorOnChainInfo.UnsealedCID field at update time, against which preserved pieces are checked via Merkle sub-tree (PoDSI) inclusion proofs. Because an FVM actor can read state but not historical events, storing CommD is what makes on-chain verification possible; where only off-chain verification is needed, the same CommD is already carried by the FIP-0083 sector-updated event and no new state is required.

As in FIP-0019, this FIP is scoped to CC sectors. Sectors sealed natively with data remain out of scope, for the same reason FIP-0019 excludes them (see Limitations), and are noted only as Future Work.

**Discussion:** [#538](https://github.com/filecoin-project/FIPs/discussions/538)


ThatGuyLLC:
pinged some FIP editors for review for you :)

rvagg:
Let's give this one the number FIP-0117. Please update the filename, frontmatter and README to link to it.

ThatGuyLLC:
@lucaniz see above comments from Rod

jennijuju:
@ThatGuyLLC Luca is on leave for the next couple weeks;

ThatGuyLLC:
thansk for update, ill add the numbers for them then :D

FIPS/fip-resnap.md
```diff
@@ -0,0 +1,251 @@
+# FIP-XXXX: Repeated SnapDeals with Piece-Level Integrity
```
jsoares, line 1: This should be below the frontmatter, not above

FIPS/fip-resnap.md
```diff
@@ -0,0 +1,251 @@
...
+
+| fip | <to be assigned> |
+| --- | --- |
+| title | Repeated SnapDeals (Re-Snap) with Piece-Level Integrity |
+| author | @lucaniz (kuba/irene to be added if they want to) |
+| discussions-to | [#538](https://github.com/filecoin-project/FIPs/discussions/538) |
+| status | Draft |
+| type | Core |
+| created | 2026-06-18 |
+| requires | FIP-0019, FIP-0076, FIP-0083 |
```
jsoares, line 11: Formatted as a markdown table, should be yaml frontmatter

FIPS/fip-resnap.md
```diff
@@ -0,0 +1,251 @@
...
+
+As in FIP-0019, this FIP is scoped to CC sectors. Sectors sealed natively with data remain out of scope, for the same reason FIP-0019 excludes them (see Limitations), and are noted only as Future Work.
+
+## Change Motivation
+
+FIP-0019 unlocked committed capacity for a single round of data onboarding. Re-Snap addresses what FIP-0019 explicitly deferred (its Future Work anticipates terminating a deal in place for a new one, and updates to sectors that already hold data). Two concrete needs motivate it:
```
jsoares, line 27:
```suggestion
FIP-0019 unlocked committed capacity for a single round of data onboarding. Re-Snap addresses what FIP-0019 explicitly deferred (its Future Work anticipates terminating a deal in place for a new one and updates to sectors that already hold data). Two concrete needs motivate it:
```

FIPS/fip-resnap.md
```diff
@@ -0,0 +1,251 @@
...
+
+### One-message Update Protocol
+
+The protocol is that of FIP-0019, with the difference that it may be initiated on a sector that has already been snapped. The encoding step is unchanged:
+
+```
+newReplica[i] = Enc(sectorKey[i], data[i], EncodingRand(i))
+```
+
+For a first snap, `sectorKey` is the CC replica, exactly as in FIP-0019. For a Re-Snap, the sector is first decoded back to its CC key (`CommR_D → CommR_zero`) and the new data is then encoded against that same recovered CC key (`→ CommR_{D'}`). The CC key is invariant across re-snaps; the encoding sees the same `sectorKey` it would on a first snap. No step of the FIP-0019 protocol is removed; one precondition is relaxed, as detailed in the actor-method spec below.
```
jsoares, line 68: Would be nice to turn references to other FIPs into (relative) links.

jsoares (approved):
The spec might need a little more detail, but I'll leave that assessment to implementers.

ThatGuyLLC:
@anorth Are you abel to review this one

PR #1270, opened 2026-07-14
FIP-0118: (Solstice) Deprecate FIL+ and Fund Services via Block-Reward Split

irenegia:
This PR adds a draft FIP proposing to deprecate FIL+ datacap and fund
services via a governed block-reward split (consensus / service
streams).

Discussion: https://github.com/filecoin-project/FIPs/discussions/1249

Notes for editors:
- FIP number to be assigned
- Backwards Compatibility, Test Cases, and Implementation sections still TODO


TippyFlitsUK:
Let's give this one the number FIP-0118. 0116 and 0117 are already allocated to #1248 and #1269 respectively, though neither has landed on master yet, so the README table still stops at 0115.

Please update:

- `FIPS/fip-solstice.md` → `FIPS/fip-0118.md`
- frontmatter: `fip: "to be assigned"` → `fip: "0118"`
- the H1 heading: `FIP-XXXX:` → `FIP-0118:`
- `resources/fip-solstice/` → `resources/fip-0118/`, and the two image links in the doc that point at it (the image filenames themselves are fine as they are)
- add a row to the README table linking to the new file

The number doesn't imply a merge — assigning it now so the proposal can be referred to by number while the draft is still in progress.

decentramike (approved):

irenegia:
> Let's give this one the number FIP-0118. 0116 and 0117 are already allocated to #1248 and #1269 respectively, though neither has landed on master yet, so the README table still stops at 0115.
> 
> Please update:
> 
> * `FIPS/fip-solstice.md` → `FIPS/fip-0118.md`
> * frontmatter: `fip: "to be assigned"` → `fip: "0118"`
> * the H1 heading: `FIP-XXXX:` → `FIP-0118:`
> * `resources/fip-solstice/` → `resources/fip-0118/`, and the two image links in the doc that point at it (the image filenames themselves are fine as they are)
> * add a row to the README table linking to the new file
> 
> The number doesn't imply a merge — assigning it now so the proposal can be referred to by number while the draft is still in progress.

done!

kelsey0907:
May I ask whether all the updates will be completed by June 2027? This includes the deprecation of Filecoin Plus (FIL+). The community is looking forward to seeing all these updates implemented by June 2027.

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+propose a change to the block reward and sector valuation mechanism to
+support the network's evolution from a storage-capacity onboarding phase
+to a service-oriented economy. Two main changes:
+
+1.  **Deprecate datacap minting and the FIL+ allocation system** and
+    treat all sectors as having the same QAP at the L1 (consensus)
+    level. All sectors onboarded from the activation epoch
+    automatically gain 10× QAP for the sector lifetime, realized
+    immediately at onboarding regardless of content. The initial
+    pledge formula is unchanged.
```
rvagg:
```suggestion
    immediately at onboarding regardless of content. Provisions are also
    made for upgrading existing sectors to 10× QAP. The initial pledge
    formula is unchanged.
```

to be clear; the "all" in the first sentence isn't quite right, but it's almost right with enough caveats.

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+      - w1 x BR to the winning miner (consensus stream),
+
+      - and the residual w0 x BR, with w0 = 1 − w2 − w1, to the burn actor.
+
+The weights are per-epoch values that f02 computes from a set of cached, in-state scheduler parameters and applies to each block reward at issuance; since the split is linear, this is equivalent to a per-epoch division. BR is the newly minted block reward only: gas rewards remain paid in full to the winning miner, and penalties are assessed as today (Section 2.4).The split is unconditional and applies to all new block rewards.
+
+The **Stream Weights Actor (SWA)** is the new governed contract that
+computes all of the weight-schedule parameters and can update them over
+time. Each update writes the new parameters into the L1 Reward Actor
+(f02) and is subject to the activation timelock mechanism (see Section
```
rvagg:
```suggestion
(f02) and is subject to an activation timelock mechanism (see Section
```

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+    issuance, the built-in L1 Reward Actor (f02) divides the full block
+    reward BR by weight:
+
+      - w2 x BR to the *Service Orchestrator wallets* (service stream),
+        
+      - w1 x BR to the winning miner (consensus stream),
+
+      - and the residual w0 x BR, with w0 = 1 − w2 − w1, to the burn actor.
+
+The weights are per-epoch values that f02 computes from a set of cached, in-state scheduler parameters and applies to each block reward at issuance; since the split is linear, this is equivalent to a per-epoch division. BR is the newly minted block reward only: gas rewards remain paid in full to the winning miner, and penalties are assessed as today (Section 2.4).The split is unconditional and applies to all new block rewards.
```
rvagg:
```suggestion
The weights are per-epoch values that f02 computes from a set of cached, in-state scheduler parameters and applies to each block reward at issuance; since the split is linear, this is equivalent to a per-epoch division. BR is the newly minted block reward only: gas rewards remain paid in full to the winning miner, and penalties are assessed as today (Section 2.4). The split is unconditional and applies to all new block rewards.
```

rvagg: only mention of penalties in the whole FIP; we'll need to make sure it's covered in 2.4; but we might just be able to treat 2.4 as "to be updated with more detail" based on f02 change design (WIP)

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+## Specification
+
+### 1. Datacap Deprecation
+
+**Deprecation principle.** After the upgrade epoch, the miner actor
+never calls the verified registry (f06) or DataCap (f07) actors: every
+QAP decision is computed from state already recorded on the sector.
+Datacap is deprecated in full at the upgrade epoch: no minting, no new
+allocations, no claims, no claim-term extensions. Complete removal is
+preferred over a wind-down because new sectors and SnapDeals receive 10×
```
rvagg:
```suggestion
allocations, no claims, no claim-term extensions. Completely disabling these actors is
preferred over a wind-down because new sectors and SnapDeals receive 10×
```

"removal" .. I think we're not removing them in this upgrade, removal and deprecation are two different things and IIRC purging that state is a future thing.

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+  - Rule: carry forward the verified ratio already recorded on the
+    sector. Compute verified_space = verified_deal_weight /
+    current_duration, then set new verified_deal_weight =
+    verified_space × new_duration.
+
+  - Effect: a sector keeps its multiplier when extended. Legacy CC
+    extends at 1×, a fully verified legacy sector at 10×, a partially
+    verified one at its recorded fraction, and all post-upgrade
+    sectors (fully verified by default) at 10×. One algorithm covers
+    legacy and new sectors, with no special cases.
```
rvagg, line 342: I think this was my suggestion, but sitting beside allowing snaps to upgrade _and_ introducing a ExtendSectorExpiration3 makes this seem unnecessary; we should probably just make every extension do the 10x thing and not bother with the ratio thing. I don't recall if there was a good reason to do this special casing of the ratio here (tbh my original suggestion came from believing that the current design was "old sectors stay, new sectors get 10x", but that's not the case in this doc).

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+        than a few entries will make it storage-wasteful).
+
+    2.  **Weight record**: the bundle of scheduler parameters for one
+        stream weight, { v_start, slope, t_start, floor, cap }.
+        Written only by the SWA via SetWeightRecords (see later).
+
+        1.  **v_start**: the weight's starting value
+
+        2.  **slope**: change per epoch (negative for w1's ramp, 0 for
+            a constant weight like w2)
```
rvagg:
```suggestion
        2.  **slope**: the weight's change per epoch over this
            segment; negative for a declining ramp (w1), positive
            for a rising ramp (w2 during the Q1 bootstrap), and 0
            for a weight held flat over the segment (w2 from Q2
            onward, between gate steps)
```

otherwise this is a contradiction with the rest of the FIP, particularly as we get down around line 950 where it's fairly clear about this

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+    2.1). FPV_i(Q) counts only settlement events emitted by admitted
+    addresses. As new Filecoin Pay versions are deployed, SRA
+    Governance updates the admitted list; no protocol change is
+    required.
+
+#### 2.4 L1 actor changes (f02)
+
+Add the following structures and methods:
+
+1)  **Stream**: a named share of the block reward Stream = { id,
```
rvagg:
```suggestion
1)  **Stream**: a named, weighted portion of the block reward. Stream = { id,
```

trying to stomp out confusion over "weights" and " shares" which are easy to confuse (from experience!)

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+
+We do **not** adopt the word “subnet.” In Filecoin, “subnet” already has
+an established and different meaning: an IPC (InterPlanetary Consensus)
+subnet, a Layer-2 / child network with its own consensus. Reusing the
+term for “a named share of the block reward” would collide with that
+meaning and mislead readers into expecting a separate consensus network,
+sequencing, or settlement layer. None of that is involved here.
+
+We therefore use “**stream**”: each stream is a named, weighted share of
+every block reward at issuance, defined by its recipients (who it pays) and its weight (how large the share is). The naming avoids a term already taken in this ecosystem while keeping the design’s lineage explicit.
```
rvagg:
```suggestion
We therefore use “**stream**”: each stream is a named, weighted portion of
every block reward at issuance, defined by its recipients (who it pays) and its
weight (how large the portion is). The naming avoids a term already taken in
this ecosystem while keeping the design’s lineage explicit.
```

stamping out "weight" / "share" confusion

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+  - **Don’t spend without proof**. The service weight (w2) can
+    only step up when on-chain Filecoin Pay volume clears a verifiable
+    target. If volume falls short, w2 stays unchanged and the
+    registered Orchestrator addresses continue receiving at the
+    previously proven rate.
+
+  - **Burn what isn’t allocated.** The gap between what leaves
+    consensus (1 − w1) and what is actually allocated (w2) is burned
+    as w0 = 1 − w2 − w1. Why burn rather than leaving it in the
+    consensus share? Because continuing to reduce the consensus share
```
rvagg:
```suggestion
    consensus weight? Because continuing to reduce the consensus weight
```

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+    the compromised wallet keeps receiving its share of w2 until
+    freeze or removal by the SRA.
+
+## Incentive Considerations
+
+**Client incentives:** clients gain access to subsidised, verifiable,
+real-time deal infrastructure without waiting for datacap allocation.
+
+**SP incentives:** all PoRep SPs see their direct block-reward income reduced
+by the non-consensus share (1 − w1). SPs who serve clients recover this
```
rvagg:
```suggestion
by the non-consensus portion (1 − w1). SPs who serve clients recover this
```

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+    recomputation (Section 2.1). Unlike the share misallocation, which is confined to one quarter, a wrongly executed step is standing: the gate only moves w2 up, so reversing it takes a discretionary, FIP-backed SWA write lowering the w2 record (Section 4.2).
+
+  - **Sustainability of consensus mining.** The initial ceiling (1 −
+    w1 = 5%) is not expected to materially threaten L1-PoRep SP
+    profitability, and the continuous ramp provides predictability. At
+    higher ceilings the burn maintains discipline: unproven funds burn
+    rather than accumulate, and the consensus-share reduction is
+    predictable and can be halted if governance is needed.
+
+  - **Baseline minting coupling.** The split sets each stream’s share
```
rvagg:
```suggestion
  - **Baseline minting coupling.** The split sets each stream’s portion
```

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+to make attacks prohibitively expensive; beyond sufficiency, each
+marginal FIL spent on consensus buys little additional security and no
+additional demand. Spending nearly all issuance on consensus was the
+right call during bootstrapping, when the reward had to build capacity
+and security from zero. That phase is complete: raw power today far
+exceeds any realistic attack budget, and the binding constraint on
+Filecoin's future is now demand for paid services. The ramp therefore
+reallocates spend from where marginal returns are exhausted to where the
+network is actually constrained: proven service demand. And where demand
+is not yet proven, the freed share is burned, reducing dilution for all
```
rvagg:
```suggestion
is not yet proven, the freed portion is burned, reducing dilution for all
```

FIPS/fip-0118.md
rvagg:
```suggestion
    consensus portion? Because continuing to reduce the consensus weight
```

rvagg:
@kelsey0907 this FIP is on the governance path, and those involved are keen to see it pass ASAP; so as long as that governance process is successful then we could be adopting this before the end of this year and FIL+ will be dead.

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+  - Termination methods: no changes required.
+
+### 2. Block-Reward Split
+
+#### 2.1 Architecture overview
+
+The two-level split is performed by the f02 built-in actor and
+parameterised by **two governed contracts off the consensus path**. This
+separates the **split among streams** from the **split among
+orchestrators** (see Figure 1).
```
rvagg:
```suggestion
separates the **split among streams** ("weights") from the **split among
orchestrators** ("shares") (see Figure 1).
```

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+
+2.  **Stream Weights Actor (SWA): the split AMONG streams.** It
+    registers or removes streams and sets every stream’s weight as a *Weight record*; each stream’s
+    designated writer sets the wallets and shares within its own
+    stream; and f02 itself enforces the hard guardrails (Σ_i w_i = 1, caps
+    on the number of streams and recipients, and an activation
+    timelock).
+
+3.  **Service Rewards Actor (SRA): the split AMONG orchestrators.** It
+    computes the orchestrator shares s_i and holds the registry
```
rvagg:
```suggestion
    computes the orchestrator J *Shares* s_i and holds the registry
```

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+    orchestrator). Each quarter it computes the orchestrator shares
+    from verifiable volume and writes the wallet-to-share map into
+    f02; it never receives or holds value.
+
+#### 2.2 Quarterly cycle
+
+Quarters are consecutive EPOCHS_PER_QUARTER-epoch windows counted from
+the activation epoch.
+
+![Timeline of one quarter: posting period, verification window, values bind, conversion, gate check, shares update, timelock](../resources/fip-0118/quarter-timeline.png)
```
rvagg, line 431: "USD at the 30-day TWAP" needs updating here

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+        can only under-count, never over-count.
+
+      - *Parameters.* MIN_LOT (minimum lot face value for a
+        qualifying print), PRICE_BAND (a posted print whose implied
+        rate deviates from the last bound qualifying print by more
+        than the band is rejected at posting; the thin-auction
+        guard), and MAX_PRICE_PERIODS (upper bound on pricing periods
+        applied per quarter; if more prints qualify, adjacent periods
+        are merged and the merged slice is converted at the
+        lot-weighted mean of its prints). All three are SRA state,
```
rvagg, line 638: Unspecified edges in here, although unlikely we'll go over MAX_PRICE_PERIODS unless we have wild swings in bursty activity and volume, how do you calculate which "adjacent periods" get merged? Without clear specification the "anyone can verify" rule is broken. Perhaps this is one of those cases where _"we had to make a call, here's the call we made, you can verify that it resolves according to our call and if you want to argue let's go"._

@martapiekarska pls note. I think the FIL conversion mechanics in here are probably going to be the hardest part of volume calculation, compounded by the lack of events (~solved, see https://github.com/filecoin-project/solstice/issues/7, but not easy).

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+        governance-settable like the admitted lists (Section 3.2);
+        proposed initial values are maintained in the governance
+        repository (MIN_LOT of a few hundred USD, sized so that
```
rvagg, line 641: "few hudred USD" wew, that's a long way from where we are now; the auctions only operate on the FP fee value, so a small fraction of settlement values.

Also this value needs to be defined in the FIP I think. Perhaps we should just make this value proportional to some settled USD volume - settled volume of one or more of the tokens in a quarter, divided by .. 32, divided by fee (0.5%) divided again by some other number to hit a reasonable minimum for the period. For low FIL settlement we're still in "dust" territory subjectively.

Currently the higest auction we've seen cleared was ~$11, that was when we were burning the PDP sybill fee via FP. Currently sitting above $1 each auction (growing slightly).

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+Note that Weight and Distribution are orthogonal: a stream picks each
+independently, so registering a new stream means giving it a Weight and
+a Distribution (and, when EXPLICIT, naming its single writer). f02 never
+computes a distribution; the writer contract computes the shares off the
+per-epoch path and pushes the map through SetShares, exactly as the SRA
+does for the service stream, so a new stream needs no new L1 logic. Note
+that f02 stores only the writer's address and the current map, never the
+rules: how the shares are computed (for the service stream, SplitRule
+over FPV volumes) lives inside the writer contract, which carries its
+own governance (the SRA under SRA Governance, Section 4).
```
rvagg:
```suggestion
a Distribution (and, when EXPLICIT, naming its single writer). f02 is not
involved in calculating the stream share distribution; instead the responsibility lies
with the the writer contract which computes according to its
rules and uses SetShares to push the map to f02 to handle the mechanics of
dividing the shares. f02 only needs to store the writer's address and the current
share distribution map, never the rules. For the service stream in particular, the
SplitRule over FPV volumes lives inside the SRA contract and governance process,
leaving f02 agnostic.
```

not sure I've made it much clearer but "never computes a distribution" isn't really correct, it does, but according to the shares given; I'm attempting to minimise confusion here.

FIPS/fip-0118.md
```diff
@@ -0,0 +1,1647 @@
...
+    to SetShares(id, map) and stays callable only by that stream’s
+    designated writer.
+
+9)  **SetDistribution**(id, distribution) re-points a stream’s
+    Distribution, in particular the designated writer of an EXPLICIT
+    stream. Callable only by the SWA and queued under SWA_TIMELOCK
+    like every SWA write. The stream’s current wallet-toshare map
+    remains in force until the new writer overwrites it via SetShares,
+    so payments continue across the transition. Converting a stream to
+    IMPLICIT is not permitted (IMPLICIT is consensus-only).
```
rvagg:
```suggestion
    IMPLICIT is not permitted.
```

burn is kind of implicit too? maybe leave this unsaid since 2.4.1 goes to so much effort to be abstract about it

rvagg:
Addressing the bulk of my comments in https://github.com/filecoin-project/FIPs/pull/1271 and getting it into a mergable state so we can iterate on the details in this repo rather than lingering in a huge single doc PR.

FIPS/fip-0118.md
irenegia: why the "all" is not right? I am not seeing it

FIPS/fip-0118.md
irenegia: I think this still earns its place, for a reason that's now about pledge rather than "old sectors stay": the two paths that grant the extra 9× QAP (snap and ExtendSectorExpiration3) both require a pledge top-up, so the multiplier is always paid for. If plain ExtendSectorExpiration2 also granted 10×, it would hand out the same QAP for free; and if we made every extension require the top-up instead, we'd recreate the forced "pay up or retire" decision point for legacy SPs that this section avoids. So: plain extension = economically neutral, upgrade to 10× = explicit opt-in with pledge top-up.

FIPS/fip-0118.md
irenegia: done

FIPS/fip-0118.md
irenegia: Good catch, that number is not calibrated with today's auctions or the entry gate target ($3.5K/quarter). 

Agree it belongs in the FIP as a formula, along your lines: MIN_LOT(Q) = max(FLOOR, α × BoundVolume(Q−1) × FEE_RATE / MAX_PRICE_PERIODS), with FLOOR a small multiple of the gas cost to claim, and α, FLOOR governance-settable.  Should I add this to section 2.3?

FIPS/fip-0118.md
irenegia: Right, the selection rule was unspecified. Proposal for 2.3: order the quarter's N qualifying prints by epoch. If N ≤ MAX_PRICE_PERIODS (M), nothing merges. Otherwise partition the N periods into M contiguous groups by index, group k = periods floor(kN/M) through floor((k+1)N/M) − 1; each group is one pricing period converted at the lot-weighted mean of its prints. Fully determined by the ordered print list, so anyone recomputes the same result.

FIPS/fip-0118.md
rvagg: The "all" in the first sentence: _"and treat all sectors as having the same QAP at the L1 (consensus) level"_ - it's not "all", it's only new ones and then upgraded ones, but there's still lingering <10x sectors.

FIPS/fip-0118.md
rvagg: yes, but let's do this one as a follow-up on top of https://github.com/filecoin-project/FIPs/pull/1271 when we get that landed; so we can do focused, small PRs and get them in quicker

rvagg:
Inherited all of the new fixes from here into https://github.com/filecoin-project/FIPs/pull/1271 and added some additional clarity about ExtendSectorExpiration and ExtendSectorExpiration2, which I didn't realise until the thread above - pledge will need to change but we don't want to change the caller contract for those messages.

Let's work on getting #1271 merged and then we can do smaller, targetted updates that are easier to review & merge.

irenegia:
closed because superseded by #1271

PR #1272, opened 2026-07-30
FIP-0118: specify f02 mechanics: pull settlement, queue, validation

rvagg:
FIP-0118 update: the f02 (reward actor) mechanics are now fully specified (Section 2.4/2.5 rewrite). This comes out of work done in https://github.com/filecoin-project/solstice/pull/13.

- Payouts are now claim-based. f02 holds each orchestrator's earnings and anyone can trigger the withdrawal to the named wallet(s). Nothing new is pushed out per epoch, and the old "failed payment burns your share" rule is gone.
- Share updates take effect immediately when the SRA submits them; the quarterly rhythm comes from the SRA process, not the reward actor. Earnings are always settled under the shares in force while they were earned.
- The gate's step-up (via QuarterlyGateCheck) is now its own uncancellable operation (separate to the discretionary ProposeWrite). Every other governance write still queues under the 7-day hold and either Safe can cancel it.
- Bad weight schedules are rejected up front, when queued, instead of being worked around at runtime. The split can never pay out more than the block reward.
- Nothing strands: a removed stream's unpaid balances stay claimable until drained.
- Hard caps: 8 streams, 64 recipients per stream. Raising either takes a new FIP and network upgrade.
- Total issuance and circulating-supply reporting are untouched; only who receives issuance changes.

Update 1:

- Codifies consts `SIMPLE_TOTAL` = 330,000,000 FIL and `BASELINE_TOTAL` = 768,335,872,210,768,889,362,796,814 attoFIL as per live mainnet measurements.

### DIFF
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -1149,12 +1149,19 @@ goes.
 ##### 2.4.6 Stream lifecycle: removal, re-pointing, tombstones
 
 `RemoveStream` and `SetDistribution` first settle the current period
-exactly as `SetShares` does: each recipient's earned-but-unclaimed
-amount is banked as payable under the outgoing shares. On removal
-those balances move to a **tombstone** for the stream's id, so claims
-stay addressable after removal: `Claim` pays from the tombstone, rows
-delete as they are claimed, and a drained tombstone deletes entirely,
-so nothing about a removed stream is permanent. `RemoveStream` admission
+exactly as `SetShares` does. Earned-but-unclaimed balances remain
+`payable` on a map or writer change and move to a tombstone for the
+stream id on removal. `Claim` pays and deletes rows; a drained tombstone
+deletes.
+
+Surviving balances are capped at `2 * MAX_RECIPIENTS = 128` payable rows
+per live stream and `MAX_TOMBSTONE_ROWS = 256` rows across tombstones.
+`SetShares` rejects a fold whose resulting `payable` table exceeds the
+per-stream cap. A hostile designated writer can therefore fill only its
+own stream's table; a global cap would let one stream block another's
+`SetShares`. Because `Claim` is permissionless and accepts
+`MAX_RECIPIENTS` wallets, a blocked writer can drain one full share map's
+stale rows in one call and retry. `RemoveStream` admission
 requires existing tombstone payable rows plus one
 `max(MAX_RECIPIENTS, |payable ∪ shares|)` reservation for each pending
 removal, including the candidate, to be at most `MAX_TOMBSTONE_ROWS`.
@@ -1162,8 +1169,7 @@ For each removal, `|payable ∪ shares|` counts the distinct recipient IDs
 in that target stream's two tables; an ID present in both counts once. Any
 `SetShares` while a removal is pending recomputes every reservation from
 the post-fold `payable ∪ shares` and rejects the change if the aggregate
-exceeds the cap. Removal revalidates at application. Each successful
-`Claim` immediately frees capacity by deleting paid rows. A removed
+exceeds the cap. Removal revalidates at application. A removed
 stream's weight reverts to the burn residual until reallocated. Id
 non-reuse after the tombstone is gone is SWA discipline (f02 still rejects any
 duplicate it can see); violating it risks event-history ambiguity
@@ -1394,20 +1400,20 @@ outcomes require replay or state comparison.
 
 **`MAX_STREAMS = 8`, `MAX_RECIPIENTS = 64`,
 `MAX_TOMBSTONE_ROWS = 256`,
-`MAX_PENDING_WRITES = 3 * MAX_STREAMS + 2 = 26`.** `MAX_STREAMS`
+`MAX_PENDING_WRITES = 26`.** `MAX_STREAMS`
 includes Consensus. `MAX_RECIPIENTS` bounds each share map and
 `Claim.wallets`; `MAX_TOMBSTONE_ROWS` caps payable rows across all
-tombstones. The queue bound is its exact slot space: three per-stream
-operations and two schedule-wide operations.
+tombstones. The imposed queue bound, sized for all per-stream operations
+across a full stream table plus two schedule-wide slots, is enforced as a
+cap by an explicit length check rather than derived from slot space.
 
-These bounds limit configured award and mutation work. Live-stream
-payable rows have no aggregate bound and drain through `Claim`. At the
-tombstone cap, operators remove until reserved capacity is full, drain
-rows through permissionless `Claim`, then continue. The shared cap also
-couples governance tiers: each pending removal reserves at least 64 rows,
-so at most four can be in flight, and the lower-tier SRA writer can reduce
-that concurrency by growing recipient unions; it cannot block removals,
-which serialize under the drainage cadence above. Raising a limit requires
+These bounds limit configured award and mutation work. Payable rows drain
+through `Claim`. At the tombstone cap, operators remove until reserved
+capacity is full, drain rows through permissionless `Claim`, then continue.
+The shared cap also couples governance tiers: each pending removal reserves
+at least 64 rows, so at most four can be in flight. The lower-tier SRA
+writer can reduce that concurrency by growing recipient unions but cannot
+block removals, which serialize under the drainage cadence above. Raising
 a future FIP and network upgrade that also reshapes the affected structures
 for the new scale.
 
@@ -1695,7 +1701,7 @@ VOL_TARGET_ENTRY = 3_500  // USD. First gate (10% -> 15%); calibrated to current
                           // Filecoin Pay actuals, at FIL = $1
 VOL_TARGET_RATIO = 2.7    // per-step escalation; back-loads the volume requirement
 
-// Back-loaded escalation: every 5pp step demands ~2.7x the previous volume;
+// Back-loaded escalation: every 5pp step demands ~2.7x the previous target;
 // preserves a low entry barrier while tightening the terminal requirement.
 // Frozen schedule (USD): 3.5K, 9.4K, 25.5K, 68.9K, 186K, 502K, 1.36M, 3.66M
 // for gates 10->15 through 45->50. Fixed at FIP time.
@@ -2242,9 +2248,10 @@ The implementation test suite MUST cover at least:
     types, including explicit failure rollback and implicit best effort;
 
   - stream registration and distribution changes, share-map validation
-    and immediate binding, period folds and dust, positional claims and
-    send rollback, full explicit-stream decommissioning, removal around
-    pending gate writes, repeated awards with no explicit stream,
+    and immediate binding, payable-cap rejection and claim relief, period
+    folds and dust, positional claims and send rollback, full explicit-stream
+    decommissioning, removal around pending gate writes, repeated awards with
+    no explicit stream,
     consensus-stream removal, and stream-ID reuse only after its
     tombstone drains;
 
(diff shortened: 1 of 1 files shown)


rvagg:
Added in the work I was doing in #1273 and a bunch more after iterating in the f02 implementation and its interactions with the SWA.

Full FIP in this branch can be viewed here: https://github.com/filecoin-project/FIPs/blob/rvagg/0118-f02/FIPS/fip-0118.md

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+        1.  **shares**: the current wallet-to-share map, at most
+            `MAX_RECIPIENTS` entries, written via `SetShares`;
+
+        2.  **payable**: per-wallet amounts carried over from closed
+            periods, not yet claimed;
+
+        3.  **`claimed_period`**: per-wallet amounts already claimed
+            against the current period's accrual.
+
+        A future stream paying one fixed wallet needs no separate
```
irenegia: "A future stream paying one fixed wallet needs no separate form" should be "A future stream paying any fixed set of wallets needs no separate form", no?

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+tuple (`TokenAmount` counters and an epoch, both small on the wire);
+everything else lives behind one CID, written only at governance,
+quarterly or claim cadence.
+
+| Field | Change |
+|---|---|
+| `total_minted_reward` (position 9) | Renamed from `total_storage_power_reward`, position kept. Accrues the full block reward, all streams (Section 2.5). |
+| `simple_total`, `baseline_total` (positions 10, 11) | Deleted: retired to the code constants below, as the existing state comment anticipates ("can be moved from state back into a code constant in a subsequent upgrade"). |
+| `total_burn_minted` | New counter: cumulative w0 residual, accrued at award. |
+| `total_service_minted` | New counter: cumulative accrual across all `EXPLICIT` streams. |
```
irenegia:
```suggestion
| `total_explicit_minted` | New counter: cumulative accrual across all `EXPLICIT` streams. |
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+| `service_accrued[id]` | New, one per `EXPLICIT` stream: the current period's gross accrual. |
```
irenegia:
```suggestion
| `accrued[id]` | New, one per `EXPLICIT` stream: the current period's gross accrual. |
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+    effective_network_time: ChainEpoch,     //  3
+    effective_baseline_power: StoragePower, //  4
+    this_epoch_reward: TokenAmount,         //  5
+    this_epoch_reward_smoothed: FilterEstimate, // 6
+    this_epoch_baseline_power: StoragePower, // 7
+    epoch: ChainEpoch,                      //  8
+    total_minted_reward: TokenAmount,       //  9  renamed
+    total_burn_minted: TokenAmount,         // 10  new
+    total_service_minted: TokenAmount,      // 11  new
+    service_accrued: Vec<StreamAccrual>,    // 12  new; ascending id
```
irenegia:
```suggestion
    total_explicit_minted: TokenAmount,      // 11  new
    accrued: Vec<StreamAccrual>,    // 12  new; ascending id
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+    if epoch >= next_transition_epoch: apply due queued writes (Section 2.4.7)
+    BR = this_epoch_reward * win_count / 5      // minted reward, as today
+    paid = 0
+    for each active stream s:                   // in list order
+        portion = floor(ComputeWeight(s.weight_record, epoch) * BR)
+        if s.distribution is IMPLICIT:          // consensus stream
+            miner_reward = portion
+        else:                                   // EXPLICIT: accrue
+            service_accrued[s.id] += portion
+            total_service_minted += portion
```
irenegia:
```suggestion
            accrued[s.id] += portion
            total_explicit_minted += portion
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+    service_accrued[id] = 0; clear claimed_period
+    install new_map
+```
 
-  - f02 adds **one cumulative counter per non-consensus stream**,
-    `total_stream_reward[id]`, incremented at accrual in `AwardBlockReward`.
+A wallet dropped from the new map keeps its payable balance until it
+claims. The residue is rounding dust only, because shares sum to
+exactly one, and it burns without touching either counter:
+`total_service_minted` already counted it at accrual, so the
```
irenegia:
```suggestion
`total_explicit_minted` already counted it at accrual, so the
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+    require caller == the stream's designated writer
+    require Σ new_map shares == DENOM, every share positive
+    resolve each recipient to an ID address, rejecting on failure
+    pool = service_accrued[id]
+    for each (wallet, share) in the OLD map:
+        earned = floor(share * pool)
+        payable[wallet] += earned - claimed_period[wallet]
+    residue = pool - Σ earned               // rounding dust only
+    send(f099, residue)                     // burned; neither counter moves
+    service_accrued[id] = 0; clear claimed_period
```
irenegia:
```suggestion
    accrued[id] = 0; clear claimed_period
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+
+  - **`total_burn_minted`** (the w0 residual) and
+    **`total_service_minted`** (gross service accrual) are stored
+    counters, otherwise hard to recover independently: each needs the
+    full weight and reward history. Fold dust burns from the service
+    side and is attributed there (Section 2.4.4), keeping the residual
+    exact.
+
+  - What miners have received is the derived residual
+    `M = total_minted_reward − total_burn_minted − total_service_minted`;
```
irenegia:
```suggestion
    **`total_explicit_minted`** (gross service accrual) are stored
    counters, otherwise hard to recover independently: each needs the
    full weight and reward history. Fold dust burns from the service
    side and is attributed there (Section 2.4.4), keeping the residual
    exact.

  - What miners have received is the derived residual
    `M = total_minted_reward − total_burn_minted − total_explicit_minted`;
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+prospective.
 
-  - **`total_storage_power_reward` keeps its current meaning**: from
-    activation onward it accumulates only the consensus stream.
+```
+SetShares(id, new_map):     // designated writer only; never queued
+    require caller == the stream's designated writer
+    require Σ new_map shares == DENOM, every share positive
+    resolve each recipient to an ID address, rejecting on failure
+    pool = service_accrued[id]
```
irenegia:
```suggestion
    pool = accrued[id]
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+miner residual stays exact; f099's receipts exceed `total_burn_minted`
+by cumulative dust. A recipient that does not resolve to an ID address
+rejects the whole call: a backstop against typos and stranded funds,
+with the SRA validating recipients at registration.
+
+##### 2.4.5 Claim
+
+`Claim(id, wallets[]) -> amounts[]` is permissionless and batched. Each
+wallet's entitlement is its live portion of the current period,
+`floor(share * service_accrued[id])` minus what it has already claimed
```
irenegia:
```suggestion
`floor(share * accrued[id])` minus what it has already claimed
```

FIPS/fip-0118.md
```diff
@@ -705,193 +721,609 @@ Once per quarter, off the per-epoch path:
...
+  - Minting is counted at accrual: each award adds
+    `BR = miner + service + burn` exactly and bumps the three totals by
+    their parts, so the decomposition holds by construction. The burn
+    is minted and sent to f099: counted in `FilMined`, captured by
+    `FilBurnt`, net zero on circulating supply.
+
+The claim state carries no supply duty: the counters account for FIL
+ever minted, while the claim state accounts only for the slice still
+inside f02. At every epoch f02's balance covers
+`Σ (service_accrued − Σ claimed_period + Σ payable)` over streams and
```
irenegia:
```suggestion
`Σ (accrued − Σ claimed_period + Σ payable)` over streams and
```

FIPS/fip-0118.md
irenegia:
```suggestion
        A future stream paying any fixed set of wallets needs no separate
```

irenegia (commented):
Renamed two mechanism fields that borrowed the service stream's name: total_service_minted → total_explicit_minted and service_accrued → accrued.

jsoares (approved):

rvagg:
Added a new commit "`FIP-0118: specify reward accounting, recovery, and events`" with the proposed "service" -> "explicit" rename and some additional details that come out of implementation, and consideration of more edge cases.

- actor events and their DAG-CBOR fields/order
- derived explicit-stream liability and low-balance reward cap (satisfy existing f02 liability invariant)
- malformed/overweight schedule fallback plus timelocked repair
- pending-write bounds, cancellation/drop, and stream decommission semantics
- SubmitShares latest-bound, zero-volume, and standing-gate consequences

rvagg:
Added some tightening of reward accounting and state invariants and also added handling of corruption and invalid found-state - what to do when we try to award rewards and our state is inconsistent.

rvagg:
Two additional commits added here:

1. putting the SWA actor ID in state rather than a constant - so we can define it per-network; we trade off a few bytes in the churny f02 state block for that flexibility but having a fixed constant across all our networks makes it really hard/impossible.
2. Clearing up the slight ambiguity in the document about whether [SWA and SRA are upgpradable contracts](https://github.com/filecoin-project/solstice/issues/5) - the FIP clearly says that changes to those shouldn't require a network upgrade but some other language seems to conflict with pure upgradability. The governance model doesn't really work without upgradability so that's now spelled out and the trade-offs and shape of security concerns are cleared up.

rvagg:
Another update that deals with limits from some discovered (minor) flaws: the payable queue needs to be bounded to account for the rogue-SRA case where they could balloon the list. Also the pending writes queue limit number is asserted rather than derived because the queue can actually be larger than the derived number. So we just cap it and test.

FIPS/fip-0118.md
```diff
@@ -959,57 +1572,125 @@ quarter boundaries, and only if on-chain Filecoin Pay Volume clears a
...
 
 ```
-func QuarterlyGateCheck(Q uint64):
+// SWA state: steps = gate steps executed so far, indexing the target
+// schedule directly; last_checked_quarter = the last quarter consumed
+// by a check, initialized to 1 (the ungated bootstrap quarter).
+func QuarterlyGateCheck():
+    // Sequential once-per-quarter latch: quarters are checked in
+    // order and each quarter's bound volume is evaluated exactly once
+    Q = last_checked_quarter + 1
```
wjmelements, line 1613: I'm likely to track nextCheckedQuarter instead of lastCheckedQuarter in the SWA.

nalacateria (approved):

momack2 (approved):

PR #1273, opened 2026-08-05
FIP-0118: specify w2 gate mechanics; SWA tracks position as state

rvagg:
_(Stacked PR for now, Draft to get feedback)_

- SWA gate position is explicit state: steps counter + sequential once-per-quarter latch; the min(w2 + W2_STEP, 1 - w1) clamp and w2_level are removed
- Targets are step-indexed: a missed gate leaves the target unchanged; no catch-up; w2 moves only on the 5pp grid
- The counter is authoritative: discretionary w2 writes must carry a matching steps adjustment, taking effect at the same epoch
- Quarter numbering and window state belong to the SRA; the SWA does no window arithmetic
- `QuarterlyGateCheck()` and `SubmitShares()` take no quarter argument; shares can never regress to an older quarter
- Q1 runs the full reporting pipeline; only the gate check is skipped
- `GetState` removed; f02 exposes no read methods

### DIFF
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -115,10 +115,10 @@ streams are as follows. The weight w1 ramps down from 95% to 50% over
 roughly 9 quarters. During the first quarter (the bootstrap phase),
 `w2 = 1 − w1` so no w0 share is scheduled and w2 reaches 10% by the
 end of the quarter. From Q2 onward, w2 steps up by 5 percentage
-points per quarter,
-capped at the ceiling `1 − w1`, and only when on-chain aggregated Filecoin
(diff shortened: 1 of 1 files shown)


rvagg:
View the FIP text as it is on this branch: https://github.com/filecoin-project/FIPs/blob/rvagg/0118-swa-state/FIPS/fip-0118.md

rvagg:
Diff from master to the top of this branch: https://github.com/filecoin-project/FIPs/compare/master...rvagg/0118-swa-state

rvagg:
I'm folding this in to #1273 as the canonical change-set for FIP-0118

PR #1277, opened 2026-08-19
FIP-0118: Last Call clarifications

irenegia:
This PR batches the pending Last Call clarifications. Most items write down behavior that was already agreed in review discussions or in the Aug 17 SRA re-design doc; the binding-timing items in topic 5 change SRA measurement behavior and are flagged for review.

Main changes per topic:

#### 1. ~ExtendSectorExpiration3~ UpgradeSectorQuality mechanics
*Specification, Section 1 (Datacap Deprecation)*
Replaces the one-line mention of this method with its full mechanics: a method that upgrades any sector below 10x QAP (including partially verified sectors), optional extension in the same call, pledge raised at most once (sectors at 10x are skipped, not a failure), the shared SnapDeals/extension daily-fee rule (pre-FIP-0100 sectors get a fresh fee, so an upgraded sector is treated exactly like a post-FIP-0100 sector of the same power). Fee debt must clear in the same message or the whole call aborts, and the batch is atomic. SnapDeals recompute the pledge only on a QAP increase, so "raised at most once" holds in either order of the two paths.

#### 2. SetGateParams and volume targets
*Specification, Section 3.1.1 (volume gate)*
Removes the rounding ambiguity in the volume targets, as agreed: the table and the formula `entry * ratio^steps` did not quite agree because the table was rounded. The rounded table is replaced by the exact values, and a new Precision paragraph states that the formula is exact arithmetic, coinciding with 18-decimal fixed point for these parameters; a future retune that breaks this exactness must state its rounding rule. The "fixed schedule" wording becomes the escalation map defined by (`VOL_TARGET_ENTRY`, `VOL_TARGET_RATIO`).

*Specification, Section 3.1.1*
Writes down the validation rules raised in slack: `steps` is bounded by the gate count, the write's effective epoch must be later than the last executed gate check, and a retune applies to the next check after its effective epoch. Please @rvagg check the wording matches what was intended.

#### 3. MIN_LOT formula and pricing parameters
*Specification, Section 2.3; SRA state list in Section 3.2.*
Replaces the "few hundred USD" `MIN_LOT` guidance with the formula agreed in filecoin-project/solstice#7; the reference term is glossed as a scale reference for lot sizes, not the realized average lot (FIL fees burn and never reach the auctions), and `N(Q-1) = 0` falls back to the floor.
Band seeding (first-print anchoring): no FIL volume is priced until five fee-auction claims have met every qualifying condition except the band check.
Print ordering and "strictly after".
Writes down the four bit-exactness rules agreed in filecoin-project/solstice#7.
Parameter and admitted-list writes now apply from the next quarter boundary (topic 4). `SetPricingParams` stores nothing: the write only emits the parameter event (eng sync, Aug 18).
One wording choice to check in review: the definition of `N(Q-1)`, the number of auction claims that actually cleared in Q-1.

#### 4. SRA governance redesign
*Specification, Sections 2.2, 2.3, 2.4.4, 3.2, 4.1 to 4.4, Security Considerations*
- No governance hold on SRA methods. Every registry change is a both-multisigs write that binds at once, emits an event, and requires a published rationale. 
- The protection is transparency plus the remedy: rogue signers are replaced from the level above (FIP plus SWA re-point). The one exception is the SRA code upgrade, which keeps its hold. CancelPending and the pending queue are gone.
- `Freeze` is deleted; removal is the single sanction. `RemoveOrchestrator` burns the slice immediately (map entry repointed to f099); income accrued before it stays claimable; bindings are released; survivors gain only at the next SubmitShares.
- Wallet and multisig rotation. `ReplaceWallet` needs a public request co-signed by the Orchestrator (old wallet for rotation; new wallet as liveness proof when the old key is lost) and is prospective, so accrued balances never move.
- `ReplaceOwner` is the cooperative multisig rotation, now at both tiers: the SWA gains the mirror method (Section 3.1), binding at once, the one SWA power needing no FIP.
- Admission timing: registry effect at once; an Orchestrator admitted in Q first posts volume for Q+1 and is first paid from Q+2, anchored on the admission itself: no binding applies before the admission does (topic 5).
- Security Considerations state the two costs accepted by design: a removal is not clawback (accrued income stays claimable, so Orchestrators should claim often), and a hostile replace by two rogue Registry multisigs is rate-limited by the remedy, not prevented (a hold would only have delayed it, since the cancel rights would belong to the same captured multisigs).

#### 5. Binding timing and the removal guard 
*Specification, Sections 2.2, 2.3, 3.2, Security Considerations, Test Cases*
- `RemoveOrchestrator` reverts while an ended quarter awaits its share map. Closes a corner where an Orchestrator removed between the posting close and `SubmitShares` re-entered the fresh map and kept earning up to a quarter. The zero-FPV no-op now marks the quarter submitted.
- Binding timing, redesigned after the Aug 25 eng call. This supersedes the Aug 20 quarter rule (which had superseded the settlement-epoch rule): under the quarter rule every client appearing mid-quarter went unmeasured for its whole first quarter. Now attribution is pointwise again: a settlement counts for the pair's holder at the settlement epoch, rebuilt from registry events. A registration also covers the pair's unbound past inside its own quarter, so registration lag loses no volume; a registration in the last `REGISTRATION_CUTOFF` of a quarter (proposed 7 days, a governance-repo value) applies from the boundary, so governance always has the cutoff to contest a claim before any quarter it covers is measured. `ReassignBinding(pair, orch, inherit)`: prospective from the execution epoch for a client migration; with `inherit = true`, for a claim adjudicated wrongful, it adopts the displaced registration's call epoch, so the rightful holder gets exactly the span it would have had and the claimant keeps nothing; a binding that already applied in an earlier quarter only ever moves prospectively. Admitted lists and pricing parameters keep the next-quarter rule; determinism sits at the quarter close (every FPV input is an event at or before Q's last epoch). Quarter 1 needs no special case: only the initial governance writes need the pre-activation window, and client pairs registered during Q1 count from activation. Security Considerations gain a Claim racing bullet.
- Both are SRA behavior changes, not clarifications; flagged for @rvagg @wjmelements @tanlang @tmaciej-neti. History and rationale in filecoin-project/solstice#7.

#### 6. Miner and market alignment with the implementation
PR [#1280](https://github.com/filecoin-project/FIPs/pull/1280) by @rvagg, merged into this PR 2026-08-24. 

*Specification, Sections 1.1 to 1.5, Product/Incentive Considerations*

- ES3 renamed UpgradeSectorQuality; 
- FULL_QA_POWER flag on SectorOnChainInfo specified as the QAP mechanism; SIMPLE_QA_POWER kept on-chain but no longer read.
- Supersedes the QAP parts of FIP-0045 and FIP-0084: verified_allocation_key stays on the wire but is ignored.
- Explicit reject lists for every state-mutating DataCap and Verified Registry method (USR_FORBIDDEN); reads keep serving the frozen state; no VerReg events after the upgrade.
- Market: pending_deal_allocation_ids removed in migration; BatchActivateDeals return type simplified (miner-only caller, no external use).
- SnapDeals data-presence question resolved: deal weights still record actual data, so the existing check stays valid and no new signal is needed.

#### 7. "Safe" was shorthand: governance multisig terminology, and Question 6
*Specification, Sections 2.4.7, 3.1, 3.2, 4.1 to 4.4; Design Rationale; Security Considerations*
- As confirmed on #1278: "Safe" was (mis-)used as shorthand for "governance multisig", not a Safe{Wallet} requirement. Section 4.2 now defines **governance multisigs** (the two SWA multisigs, the two Registry multisigs), states that the contracts identify each one by its address, and that no specific multisig implementation (e.g., Safe{Wallet}) is required. Every governance use of "Safe" is renamed accordingly; the two remaining Safe{Wallet} mentions are deliberate examples.
- Adds Question 6 to the Design Rationale (why hold-plus-veto in the governed contracts instead of a multisig of multisigs), adapted from @BigLep's #1278, which this supersedes. Amended per his review here: the hold is what makes the unilateral veto realistic (the pending write is public for its full duration), the choice of host for that logic is secondary, and f02 enforcement preserves the delay, not the veto, under a compromised contract.
- Also carries one small 4.4-table clarification: volume corrections bind at window close (Section 4.3). The Implementation section now points at the Solstice project board.

#### 8. In-flight messages at the upgrade boundary
*Specification, Section 1.5; Backwards Compatibility*
Writes down the boundary behavior agreed on the Aug 25 eng call: an in-flight message that directly calls a disabled DataCap or Verified Registry write method rejects after the upgrade, acceptable because these are standalone messages; miner- and market-actor datacap pathways become no-ops, so nothing in flight to them breaks, it just no longer has a datacap effect. @rvagg to confirm against filecoin-project/builtin-actors#1760.

### Open questions
1) ~Should the ES3 method be renamed, e.g. `ExtendUpgradeSectors`, now that extension is optional and the main effect is the upgrade? If yes, I will add the rename to this PR.~ Resolved.


irenegia:
This is a behavior change in the SRA, not a text clarification: it needs a callability check against the last-submitted-quarter state, and the zero-FPV no-op must mark the quarter submitted. @rvagg @wjmelements @tanlang please confirm this is okay

BigLep:
Meta note on the timeline above: this PR was briefly and accidentally force-pushed from `64e77f8` to `2e257ba` (wiping the branch back to master's tip) and closed, as a side effect of tooling used while preparing #1278 (using [github PR stacks](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/) PR adding a Design Rationale entry on top of this one). That event will stay visible in this PR's history since GitHub's timeline doesn't retroactively remove force-push records, but the branch has since been fully restored to `64e77f8` and the PR reopened — all 10 commits and the full diff are intact, verified directly via git and the GitHub API. No content from this PR was lost. Backup refs of the pre-incident commit were also kept as an extra safety margin.

Apologies for the noise, @irenegia.

FIPS/fip-0118.md
```diff
@@ -328,8 +328,31 @@ complexity without utility.
...
+    already share: the sector's recorded fee is scaled in proportion
+    to the power change, and sectors from before FIP-0100 with no
+    recorded fee are assigned a fresh fee at upgrade. For those legacy
+    sectors the daily fee starts at the full 10× level from the
+    upgrade onward, so an upgraded sector is treated exactly like a
+    post-FIP-0100 sector of the same power.
+
+  - `ExtendSectorExpiration3` emits one FIP-0083 `sector-updated`
+    event per successfully upgraded sector. Extensions emit no event,
+    matching `ExtendSectorExpiration2`.
```
rvagg: There's a contradiction in here - "extensions emit not event" yet it's asking for an event. Also, `sector-updated` is for PRU3 and it carries piece information, which we don't have for extensions, so if you want an event for it then we'd need a new one. We could introduce an entirely new `sector-extended` event and apply it across all of them. I'm not sure how high value it'll be though.

FIPS/fip-0118.md
```diff
@@ -328,8 +328,31 @@ complexity without utility.
...
-    to be extended and gain 9× QAP under the same pledge rule.
+  - Introduce `ExtendSectorExpiration3`: a batch method that upgrades
+    existing sectors to 10× QAP under the same pledge rule. Any sector
+    below 10× is eligible, including partially verified sectors; the
+    upgrade tops the sector up to 10× QAP and its pledge up to the
+    matching level. Each declaration may carry an optional new
+    expiration, so a sector can be extended in the same call; without
+    it, only the upgrade is applied.
+
+  - The pledge is raised at most once, at the first upgrade. A sector
```
rvagg: Being slightly pedantic for the sake of clarity: "at most once" is at odds when snaps are in the picture too, but I _think_ the intention here is that this point is entirely scoped to ESE3, so we should just be clear.


```suggestion
  - The pledge is raised at most once when using `ExtendSectorExpiration3`, at the first upgrade. A sector
```

Since snaps recompute pledge using network conditions at the time of the snap, it can be raised if you use that path, so if you do ESE3 then snap, "at most once" doesn't apply.

irenegia: mmmh, this made me thinking: Why does snap recompute the pledge at all when the sector is already at 10x?

The recompute is motivated by the QAP increase. In the future or now in the ES3-then-snap flow the increase doesn't happen, so the recompute becomes a ratchet to current network conditions with no power change behind it. It also makes the order matter: snap-then-ES3 raises the pledge once, ES3-then-snap may raise it twice.

I think the right design here should be: make the snap pledge recompute conditional on a QAP increase. Then "raised at most once" holds across the whole upgrade path, in any order.

cc @akaladarshi https://github.com/filecoin-project/builtin-actors/issues/1763#issuecomment-5349523640

FIPS/fip-0118.md
irenegia: You are right... (again i think the name of this method is creating confusion, I'll change it! )

"Extensions" meant the optional expiration change inside an ESE3 call, not the method, so the sentence needs rewriting either way. The event we need to add to mark is the upgrade, not the extension. 

`sector-extended` across ES/ES2/ES3 tracks the wrong thing: for example, ES3 call with no new expiration would emit an extension event for no extension.

So we need a new sector-upgraded event, ES3 only, payload $type + sector (like sector-terminated), emitted per sector whose recorded QAP the call actually raised.

irenegia: cc cc @akaladarshi

FIPS/fip-0118.md
```diff
@@ -677,22 +752,79 @@ Once per quarter, off the per-epoch path:
...
+        bounding and merging pricing periods only mattered when each
+        period's print data had to be submitted and iterated
         on-chain; an off-chain indexer simply prices every FIL
         settlement at its own qualifying print.
 
+      - *Minimum lot.* `MIN_LOT` is not a constant: a threshold that
+        prevents manipulation at scale would reject every print in
+        today's small auctions, so it scales with realized volume:
+
+        `MIN_LOT(Q) = max(MIN_LOT_FLOOR, (MIN_LOT_ALPHA / N(Q-1)) * BoundVolume(Q-1) * FEE_RATE)`
```
rvagg: "floating-point arithmetic is forbidden anywhere in the rule" runs into problems with the division in parens I believe, we can truncate/floor to zero

FIPS/fip-0118.md
rvagg: Agreed, I'm going to make that change in my actors branch. Snaps will only recalc pledge on QAP increase, otherwise pledge at full 10x remains entirely unchanged through either method. @irenegia leave the miner actor stuff alone in your current branch, focusing instead on the SWA/SRA stuff in here, I'm going to get another PR on top of this one with a bunch of items I've listed in ~https://github.com/filecoin-project/builtin-actors/pull/1774~ plus clarifications around this point too and you can review that.

rvagg: sorry, I meant https://github.com/filecoin-project/builtin-actors/pull/1778, where I have a list of FIP things

FIPS/fip-0118.md
rvagg:
```suggestion
    require Σ new_map shares == DENOM, every share positive
    reject a repeated recipient, except f099, which may appear more
        than once
```

FIPS/fip-0118.md
```diff
@@ -1809,33 +2046,97 @@ divided among Orchestrators. In particular,
...
+        verification keeps its standing map entry, and so its income,
+        for exactly that interval (see Security Considerations).
+        Applies the removal semantics above through an immediate map
+        push: the Orchestrator's entry is repointed to `f099` and
+        every other share is left untouched. If an `f099` entry
+        already exists from an earlier removal in the same period,
+        the weight is added to that entry, because f02 rejects a map
+        with duplicate wallets (Section 2.4.4). Redistribution to
+        the survivors is triggered only at the next `SubmitShares`,
+        so nobody gains at the moment a removal binds.
```
rvagg:
```suggestion
        every other share is left untouched. Removals within one
        period need no merging, because f099 may appear more than
        once in a map (Section 2.4.4). Redistribution to the
        survivors is triggered only at the next `SubmitShares`, so
        nobody gains at the moment a removal binds.
```

FIPS/fip-0118.md
irenegia: #1280 landed without this one, so I added it in eec731c to not lose it

FIPS/fip-0118.md
irenegia: Fixed in eec731c.
Also added the `N(Q-1) = 0` case: a quarter after zero cleared claims falls back to `MIN_LOT_FLOOR`.
cc @tmaciej-neti

FIPS/fip-0118.md
```diff
@@ -2095,6 +2448,69 @@ the contracts at runtime, for two reasons.
...
+    two parties, that matters: if a bad write gets both signatures,
+    because one signer is compromised, coerced, or simply wrong,
+    undoing it requires assembling the same two-party consent that
+    produced it in the first place. A corrective write and the
+    mistake it corrects need identical authority, so "just get both
+    to sign again" is no safeguard against the case where one of the
+    two is the problem. What is needed instead is a hold after
+    approval during which either party, alone, can cancel, with no
+    need to persuade the other. That is a distinct capability from
+    co-signing, so it does not matter how the owner's signature
```
BigLep: The veto is only as good as the odds that the honest party discovers the problem inside the window. Worth saying explicitly that the hold makes discovery realistic because the pending write is public, so it does not depend on the deceived signer noticing its own error:

```suggestion
    two is the problem. What is needed instead is a hold after
    approval during which either party, alone, can cancel, with no
    need to persuade the other. The hold also makes that veto
    realistic rather than lucky: the pending write is public for its
    full duration, so discovery of a bad write does not depend on the
    deceived signer noticing its own error; anyone watching the queue
    can raise the alarm inside the window. That is a distinct
    capability from
```

FIPS/fip-0118.md
```diff
@@ -2095,6 +2448,69 @@ the contracts at runtime, for two reasons.
...
+    (weekly-to-quarterly) writes for the life of the protocol.
+    Wherever the hold-and-veto logic lives, if it lives entirely
+    inside upgradeable FEVM code, whether the governed contract or a
+    module wrapping it, it is only as trustworthy as that code's
+    current version: a captured or maliciously upgraded SWA or SRA
+    could rewrite or bypass its own safeguard. This is why
+    `SWA_TIMELOCK` is additionally enforced at f02 (Section 4.3): the
+    built-in actor is not upgradeable via a contract-level code
+    change, so the objection window survives even a fully compromised
+    governance contract, not just a compromised signer.
```
BigLep: "the objection window survives" invites a stronger reading than what is true. f02 accepts `CancelPending` only from the SWA address and the honest party's cancel is relayed through the SWA contract, so a maliciously upgraded SWA can refuse to relay it. What survives full contract compromise is the delay (guaranteed detection and response time for the network-upgrade remedy), not the unilateral veto. Security Considerations already states this correctly; this aligns Q6 with it:

```suggestion
    could rewrite or bypass its own safeguard. This is why
    `SWA_TIMELOCK` is additionally enforced at f02 (Section 4.3): the
    built-in actor is not upgradeable via a contract-level code
    change, so the 7-day delay survives even a fully compromised
    governance contract, not just a compromised signer. What survives
    is the delay, not the veto: f02 accepts `CancelPending` only from
    the SWA itself, so against a compromised contract the window buys
    detection and network-upgrade response time (Security
    Considerations), not a surviving unilateral cancel.
```

FIPS/fip-0118.md
```diff
@@ -2095,6 +2448,69 @@ the contracts at runtime, for two reasons.
...
+    two is the problem. What is needed instead is a hold after
+    approval during which either party, alone, can cancel, with no
+    need to persuade the other. That is a distinct capability from
+    co-signing, so it does not matter how the owner's signature
+    threshold is composed; the hold and the unilateral veto have to
+    be written as their own logic somewhere, whether a module of the
+    owner multisig, a wrapper contract, or (what this FIP does) the
+    governed contracts tracking the two multisig addresses directly
+    and granting either one a standing cancel right on any pending
+    discretionary write (Sections 4.2 and 4.3).
```
BigLep: Q6 lists "a module of the owner multisig" as a possible host for the hold+veto logic (e.g., timelock/veto modules from the Safe{Wallet} ecosystem) but never says why the FIP did not use one. Worth closing that gap, since it is the obvious "why not off-the-shelf" follow-up:

```suggestion
    threshold is composed; the hold and the unilateral veto have to
    be written as their own logic somewhere, whether a module of the
    owner multisig (multisig products offer timelock and veto
    modules), a wrapper contract, or (what this FIP does) the
    governed contracts tracking the two multisig addresses directly
    and granting either one a standing cancel right on any pending
    discretionary write (Sections 4.2 and 4.3). The choice among
    those hosts is secondary: each is FEVM code the same two parties
    can change, so none supplies the survivability that only f02
    enforcement provides (below), and each still leaves that
    enforcement to be built. The governed contracts themselves are
    the smallest audit surface and the only host fully specified by
    this FIP rather than by a third-party module ecosystem with
    limited Filecoin deployment.
```

FIPS/fip-0118.md
BigLep: The in-flight PR links will go stale as soon as they merge (#1775 is already slated to disappear into #1760). The project board and the repositories are the durable references, and the Solstice repository is the lasting home of the SWA and SRA contracts. Proposed replacement for the whole section body:

```suggestion
Implementation is tracked on the
[Solstice project board](https://github.com/orgs/filecoin-project/projects/143),
which spans all affected repositories:

  - **FEVM contracts and tools:** the
    [Solstice repository](https://github.com/filecoin-project/solstice)
    is the home of the SWA and SRA contracts and their shared
    two-party governance library.
  - **Built-in actors:** miner-actor changes (DataCap/verified-registry
    disconnect, 10× QAP, `UpgradeSectorQuality`) and reward-actor (f02)
    changes in
    [builtin-actors](https://github.com/filecoin-project/builtin-actors).
  - **Node implementations and shared types:** updates to
    go-state-types, Lotus, Forest, Venus and SP tooling.
```

BigLep (commented):
@irenegia : I left a couple of suggestions when reviewing again the multisig plan and implementation references.

FIPS/fip-0118.md
```diff
@@ -410,28 +438,50 @@ complexity without utility.
...
+
+  - DataCap: `Mint`, `Destroy`, `Transfer`, `TransferFrom`, `Burn`,
+    `BurnFrom`, `IncreaseAllowance`, `DecreaseAllowance`, and
+    `RevokeAllowance` reject. `Name`, `Symbol`, `Granularity`,
+    `TotalSupply`, `Balance`, and `Allowance` continue to serve.
+
+  - Verified Registry: `AddVerifier`, `RemoveVerifier`,
+    `AddVerifiedClient`, `RemoveVerifiedClientDataCap`,
+    `RemoveExpiredAllocations`, `ClaimAllocations`, `ExtendClaimTerms`,
+    `RemoveExpiredClaims`, and the universal receiver hook reject.
```
irenegia, line 457: @rvagg todo, review if getclaims from contract is blocked

irenegia: cc @akaladarshi

irenegia: solved (universal receiver hook is a write call in all cases so it’s right to disable it)

irenegia:
@rvagg in-flight language from today's call is in (1952fd9): two bullets in Section 1.5 (after the no-event bullet) plus one clause in Backwards Compatibility. Direct f06/f07 writes reject like post-upgrade calls; miner and market calls do not fail, references are accepted and ignored.

FIPS/fip-0118.md
```diff
@@ -321,15 +329,57 @@ complexity without utility.
...
+    matching level. Each declaration may carry an optional new
+    expiration, so a sector can be extended in the same call; without
+    it, only the upgrade is applied.
+
+  - The pledge is raised at most once, at the first upgrade, in either
+    order of `UpgradeSectorQuality` and SnapDeals. A sector already at
+    10× is not eligible and is skipped (not a batch failure), so
+    repeated calls lock nothing, even if the pledge requirement has
+    risen since. A 10× sector that only needs an extension uses
+    `ExtendSectorExpiration2`.
```
rvagg:
```suggestion
    10× is not upgraded and its pledge is not raised; this is not a
    batch failure, so repeated calls lock nothing even if the pledge
    requirement has risen since. Such a sector is still extended if its
    declaration carries a new expiration, exactly as
    `ExtendSectorExpiration2` would.
```

Picked up in https://github.com/filecoin-project/builtin-actors/pull/1775, if you want to extend and a sector is already at 10x then it bails before extending, because this is what the FIP literally says. So let's be clear.

rvagg:
I pushed 717cb6b0f5dea72ee3923af16a000fcfde727000 with some technical clarifications regarding f099 handling in shares, the UpgradeSectorQuality suggestion I made above (resolved) and a missing noop re `DealProposal.verified_deal`.

rvagg (approved):

FIPS/fip-0118.md
```diff
@@ -387,28 +442,69 @@ complexity without utility.
...
+
+  - Storage market `BatchActivateDeals` remains for the
+    published-to-active deal transition. The `compute_cid` parameter,
+    returned per-sector unsealed CID, and
+    `ActivatedDeal` and `SectorDealActivation` return types are removed.
+    Each sector group is now a `Vec<PieceInfo>`, and the result returns
+    `Vec<Vec<PieceInfo>>` directly. `ActivatedDeal` encodes as
+    `[cid, size]`, while `PieceInfo` encodes as `[size, cid]`, so the
+    returned bytes change. This method is only
+    callable by the miner actor so has no external use.
```
rvagg, line 499: TODO: consider removing this since it's only on the preseal path and an internal detail that nobody should care about

FIPS/fip-0118.md
```diff
@@ -677,22 +853,91 @@ Once per quarter, off the per-epoch path:
...
+        bounding and merging pricing periods only mattered when each
+        period's print data had to be submitted and iterated
         on-chain; an off-chain indexer simply prices every FIL
         settlement at its own qualifying print.
 
+      - *Minimum lot.* `MIN_LOT` is not a constant: a threshold that
+        prevents manipulation at scale would reject every print in
+        today's small auctions, so it scales with realized volume:
+
+        `MIN_LOT(Q) = max(MIN_LOT_FLOOR, floor(MIN_LOT_ALPHA * BoundVolume(Q-1) / N(Q-1)))`
```
tmaciej-neti, line 889: Let's make sure that BoundVolume(Q) (for each Q for which aggregated FPV was bound) is also either part of SRA state or an event (preferably). Or both. This is to ensure that tooling like the indexer can properly calculate MIN_LOT(Q). Don’t know if that was highlighted in previous discussion or if it needs to be a part of the spec.

FIPS/fip-0118.md
```diff
@@ -531,14 +627,37 @@ Once per quarter, off the per-epoch path:
...
+        Every input is an event at or before quarter Q's last epoch,
+        and posting opens only after that epoch, so nothing done
+        after a quarter closes changes its measurement.
+        **Application rules.** A registration applies from its call
+        epoch and also covers the pair's unbound past inside the
+        same quarter, back to the later of the quarter's first epoch
+        and the epoch the pair last became unbound, so a payer's
+        settlements count from the quarter's start however long its
+        Orchestrator takes to register the pair. A registration made
+        less than `REGISTRATION_CUTOFF` before a quarter boundary
```
tmaciej-neti, line 648: Let's make sure that `REGISTRATION_CUTOFF` is defined and treated as other SRA state or parameters and is indexable. Unless it's supposed to be a constant changeable only by another FIP.

rvagg:
I pushed a couple of commits here for things that came up in code review:

1. Removed some internal cruft related to presealing functionality that wasn't worth mentioning. In short: preseals used to allow deals included in them and this was used to set up devnet conditions with easy 10x power. But now we can just give 10x power automatically, which means we don't need to touch the market actor for onboarding preseals. This is all system-actor callable only so it's not a user-facing detail so mention of it is removed from the FIP.
2. Explicitly disabled `deal_ids` field populating in precommits, and this is now also done in teh code @ https://github.com/filecoin-project/builtin-actors/pull/1782 -- this hasn't been allowed since DDO and the removal of PCS and PCA because there's no path by which such a precommit can be activated, it would be accepted initially and then rejected at activation (wasting a deposit) so nobody uses this. Now providing `deal_ids` will get you an immediate error so you can't land such a precommit. This has the benefit of letting us remove a lot more code from the market actor and the interaction between market and miner.

FIPS/fip-0118.md
irenegia: Good catch, as written it was neither. Fixed in [2489c25](https://github.com/filecoin-project/FIPs/pull/1277/changes/2489c25fe55dd35a61a4e41c98b70780e2bc29a5): REGISTRATION_CUTOFF now goes through the same parameter write as the pricing parameters, so indexers read it from the parameter events. Initial 7 days installed pre-activation; row added to the 2.3 recap table.

FIPS/fip-0118.md
irenegia: MIN_LOT(Q) needs three inputs:

`BoundVolume(Q-1)` is SRA state: the read view exposes AggregatedFPV(Q) for every quarter (Section 2.2 step 3; recap table row in 2.3). One state read. Note per-orchestrator bound values are kept only for the active and previous quarter; the aggregate is the only historical read, and it is the one MIN_LOT needs.
`MIN_LOT_FLOOR` and `MIN_LOT_ALPHA` come from the parameter events.
`N(Q-1)` comes from the fee-auction claim events.

On events: the FIP specifies no event for `PostVolume` / `CorrectVolume`, and the merged VolumePosted / VolumeCorrected events carry no amount. This is already decided in solstice issue https://github.com/filecoin-project/solstice/issues/34: both events will carry the value. Once that lands, an events-only pipeline can recover the bound values too; until then they are readable from state only.

Does your indexer read state, or events only?

tmaciej-neti: @irenegia it can potentially read state but i would like it to be an event

FIPS/fip-0118.md
```diff
@@ -649,88 +806,182 @@ Once per quarter, off the per-epoch path:
...
+| `BoundVolume(Q-1)` | Input to `MIN_LOT(Q)` | Measured (atto-USD) | SRA state: sum of Q-1's bound FPV values | Measured | No: immutable once bound | Missing for Q = 1 |
+| `N(Q-1)` | Input to `MIN_LOT(Q)` | Measured (count) | Fee-auction claims cleared on admitted Filecoin Pay contracts in Q-1 | Measured | No: immutable history | May be zero |
+| Admitted lists | Scope prints and FPV: a print is a claim on an admitted contract, of an admitted stablecoin | Governance parameter, event-only | `SetAdmittedLists` events | Installed pre-activation (Section 2.2) | Next quarter boundary | None |
+| Band reference | The rate the band is measured around | Derived per print | The last qualifying print's `(lotUsd, claimFil)` pair | Seeded | At each qualifying print | Seeded by the five-print median (*Band seeding* above) |
+| `REGISTRATION_CUTOFF` | Late-claim guard: a registration in the last cutoff of a quarter applies from the boundary (Section 2.2) | Governance parameter | Parameter events | 7 days | Next quarter boundary | None |
```
tmaciej-neti: Might want to use epochs as units here to avoid ambiguity. Although i see multiple usage of arbitrary time units in the doc so maybe it's implied to be 20160?

irenegia: @tmaciej-neti does this look good https://github.com/filecoin-project/FIPs/pull/1277/changes/d1a30e100a809200b8aada0566eb8aa1e3dc769a ?

momack2 (approved):
Everything here aligns with the previous FIP editor review/pass

PR #1278, opened 2026-08-19
FIP-0118: add design rationale for two-"SAFE" governance

BigLep:
Stacked on #1277 (base branch: `irenegia/0118-last-call-clarifications`).

## What this adds

A new **Question 6** in FIP-0118's "Design Rationale / Alternative Approaches Considered" section (after Question 5, before "Parameter Choices and Calibration"), answering: why do SWA and SRA use the custom `UnanimousGovernance` hold-plus-veto contract instead of a simpler single-Safe-owned contract (a "multisig of multisigs")?

This relates to https://github.com/filecoin-project/solstice/issues/2.

Covers two reasons:
- Co-signing and post-approval objection are different problems: with only two governance parties, a corrective write needs the same two-party consent as the bad write it corrects, so a bare nested multisig doesn't provide a way for one party to unilaterally halt a bad write the other also signed. `UnanimousGovernance` (Section 4.3) closes this with a hold period plus a single-owner veto.
- SWA/SRA are long-lived and upgradeable, so the objection window is additionally enforced at f02 (the non-upgradeable built-in actor) rather than relying solely on FEVM contract logic, which could itself be maliciously upgraded to bypass its own safeguard.

Also explains why this FIP doesn't reuse [FRC-0099](../FRCs/frc-0099.md)'s `F3Parameters` pattern (single Safe-owned `Ownable` contract used for one-time F3 activation params): that pattern is 3-of-3 across independent implementations and the contract is immutable/self-disabling, so neither of the two properties that motivate `UnanimousGovernance` applies there.

This came out of me trying to understand the governance design. I wanted to make sure the "why not just reuse Safe/F3's approach" reasoning is written down rather than living only in Slack/verbal discussion.

## Note

The base branch (`irenegia/0118-last-call-clarifications`) was briefly and accidentally force-pushed/closed while I was setting this PR up, then fully restored — see [this comment on #1277](https://github.com/filecoin-project/FIPs/pull/1277#issuecomment-5348372052) for details. No content was lost; flagging here for anyone tracing history on this branch.


jennijuju:
More generic: about the choice of wording — is "Safe" meant to require Safe{Wallet} specifically, or is it shorthand for "multisig"?

fip-0118.md (§4.2) describes governance as "two multisig accounts held by independent entities (e.g., Safes)" and says explicitly: "How each Safe is implemented internally (threshold, keys, custody, rotation) is that entity's affair." That reads as implementation-agnostic — "Safe" as generic terminology for "the entity's multisig account," not a mandate for the Safe{Wallet}/Gnosis Safe product specifically.

However, I think this is causing real confusion, and it's already made its way into both spec and implementation:
- [`src/lib/IsASafe.sol`](https://github.com/filecoin-project/solstice/blob/main/src/lib/IsASafe.sol)
- `owner1.isProbablyASafe()` / `owner2.isProbablyASafe()` in both the SWA constructor and the SRA constructor enforce it as a hard on-chain requirement. Any owner address that isn't a Gnosis Safe proxy in the observed v1.3.0–v1.5.0 size range reverts (`NotSafeProxy` / `UnusualSafeMasterCopy`) on registration.

**If "Safe" is meant generically, the naming is a bug — please fix:**
- "Safe" collides with the dominant EVM multisig product; "put it in a Safe" already means something specific in this industry.
- Capitalized usage (60+ times) reads as a defined term, but no definition sentence exists anywhere, unlike "Orchestrator."
- An implementer had to resolve the ambiguity to ship code, and resolved it toward the literal product, unreviewed, because the spec never said "generic" precisely enough.
- No one in the FIP text wrote "governance accounts SHALL be Safe{Wallet} instances," but that's the de facto requirement now.
- "Governance multisig" or "tier M-of-N multisig" would have been better choices here.

**If requiring Safe{Wallet} specifically is intentional, I'd push back:**
- Then the FIP text needs to change too — "e.g., Safes" and "that entity's affair" both need to become an actual requirement, stated as one, not implied by an example.
- General DAO/treasury governance practice (Optimism, Arbitrum, Balancer, etc.) keeps the multisig implementation and custody tooling out of the spec/protocol layer — M-of-N is the charter-level primitive; how it's implemented and what tooling supports it is a documented operational choice, not protocol-enforced.
- The check doesn't verify what it seems to promise. `UnanimousGovernance` authorizes by `msg.sender == owner` — a caller-address check, not a signature/threshold check. Neither this implementation nor any alternative can see or verify an owner's internal M-of-N structure; it's opaque by design once an address is registered. So `isProbablyASafe()` isn't protecting "the threshold was met" — it can't. The only real property it's guarding is "not a single EOA" — so the enforcement of Safe also feels like a hallucination.
- Worse, it doesn't even reliably confirm that. The check is a codesize range plus a bytecode-length check on `masterCopy()` — not a hash, not a bytecode diff, not provenance. Any contract padded to land in the observed byte ranges passes, whether or not it's actually a Safe, and whether or not it actually enforces anything resembling M-of-N internally. If the goal was "only trust audited, battle-tested implementations," this check can't deliver that guarantee either — it's checking shape, not identity or behavior.
- The current check is also fragile in a way that bites operationally, not just theoretically: it only runs at construction and in `replaceOwner`, checked against the new address — so a routine key rotation could be blocked and break this. The byte ranges (v1.3.0–v1.5.0) are hardcoded compile-time constants with no config path. If Safe ships a new official release with a different proxy/mastercopy size to say, fix serious security bugs; or simply  an entity's own routine rotation to their own newer, legitimate wallet reverts — through no fault of theirs — and the only recovery is a full SWA/SRA code upgrade, which needs a published FIP. Is that really the dependency we want sitting under something as routine as key rotation?

Which was intended? Happy to help either direction — swap the name for something  (and change the implementation) that actually matches its stated purpose if the goal is "not a bare key / M-of-N multisig" or state the Safe{Wallet}-specific requirement explicitly in the FIP, with the actual reasoning, if it's really required.

BigLep:
Good feedback and points @jennijuju. In this PR I focused on the question of why at the protocol level are we requiring two "SAFEs", which I was understanding as a "multisig", but you are right that we should be clear about whether we actually mean _Safe{Wallet}_ or _ governance multisig_. I shouldn't have perpetuated the ambiguity with my FIP addition. I agree with you that I would expect the fip to be about a "governance multisig", but I haven't been in on the details to know.

I am happy to expand this PR to clarify the ambiguity, but I just need to know whether the word "Safe" was intentional (referencing _Safe{Wallet}_) or whether it was shorthand for "governance multisig". @irenegia : do you know the answer here or who we can followup with?

irenegia:
@BigLep @jennijuju 

I apologize for the confusion, andI do confirm that the word "Safe" was (mis-) used as shorthand for "governance multisig. Adding now a commit to my PR #1277 to fix this ambiguity and add the rationale as suggested my by @BigLep in this PR

irenegia:
Fixed in 13ac2ea on #1277:
- Section 4.2 defines "governance multisigs" and says no specific multisig implementation (e.g., Safe{Wallet}) is required; the contracts identify each multisig by its address. All governance uses of "Safe" are renamed.
- Question 6 from this PR is in the same commit, adapted to the FIP's terms: no `UnanimousGovernance` (the FIP never defines it), plain section references, closing paragraphs dedented (both flagged by Copilot here). The F3Parameters link points to master; the main one 404s.
@BigLep I think this supersedes this PR. @jennijuju the implementation check (`IsASafe.sol`) is now out of spec; tracked in the solstice issue https://github.com/filecoin-project/solstice/issues/30

BigLep:
Thanks @irenegia. I will close this PR.

PR #1279, opened 2026-08-21
FIP-0118: add Orchestrator transparency rules

jennijuju:
## Summary

Adds a reporting requirement to §4.2 that's currently missing: each Service Orchestrator SHALL share a quarterly Community Report with the community, and — since the first Orchestrator's wallet is fixed by the activation migration and never goes through `Admit` — the Registry Safes SHALL disclose who operates it, why, and what's expected of it, by activation. 

## What this adds

- **Orchestrator Community Report** — quarterly, public, binary filing (filed or not; failure is a freeze trigger). The template and its cadence live in and iterate through the [governance repository](https://github.com/FilecoinFoundationWeb/Solstice-Governance/tree/main), not in this FIP — so it can change by PR without a FIP amendment. No genesis exemption: applies to the Orchestrator seated at activation too.
- **Genesis Orchestrator disclosure** — the first Orchestrator is seated by the migration, not `Admit`, so it never gets the onchain scrutiny every later admission does. This closes that gap.
- Given the group's strong preference to keep the reporting and Orchestrator-governance process out of the FIP for quick iteration, that responsibility implicitly shifts to be heavier on [Filecoin Foundation's governance team](https://fil.org/governance), who maintain the governance repo, hold the power to set the runbook, expectations, and report template, and share in admit/freeze decisions as one of the SRA's Registry Safes — accountable to the community for keeping the process credible and neutral.

## Why

The goal is transparency. According to the overall governance goal, "our goal is to facilitate a decentralized, innovative, and transparent system to support community decision-making and collective ownership of Filecoin" (fips.filecoin.io). As raised in the FIP discussion, this subnet is GTM/PMF-focused which needs strong strategy and robust execution — companies raising capital don't get funded on revenue data alone, they also share their plans and high level operations. Paying Service Orchestrators out of the block reward with no requirement that they report to the community beyond onchain ARR, and no requirement that anyone even knows who's behind the first one, doesn't meet that bar IMO.

## Disclosure   

I'm not affiliated with Filecoin Foundation.

### DIFF
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -1934,12 +1934,11 @@ The template, required disclosures, and filing cadence are defined in the
 Failure to file is a removal trigger and forfeits re-admission for the
 following quarter. 
 
-**Genesis Orchestrator disclosure.** The Orchestrator seated at
-activation is fixed by the activation migration (Section 2.4.10), not
-by `Admit`. To carry the same public scrutiny an `Admit` call would,
-the Registry Safes SHALL share, by activation, who operates that
-wallet, why it was selected, and what is expected of it — distinct
-from and in addition to that Orchestrator's own report above.
+**Initial Disclosure of Orchestrator Seated at Activation.** The Orchestrator
+seated at activation is fixed by the activation migration (Section 2.4.10), not
+by `Admit`. To carry the same public scrutiny an `Admit` call would, the
+Orchestrator shall publish a disclosure on activation, following the
+requirements in the governance repository set for any other Orchestrator.
 
 **Service Orchestrators** are not considered a governance tier. An
 Orchestrator's only protocol interaction is receiving rewards and
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
 escalation-to-Remove policy, the dispute procedure, the audit playbook,
 and the quarterly verification duty live in the governance repository.
 
+**Orchestrator Community Report.** Each Service Orchestrator SHALL
+share a quarterly Community Report with the community, using the
+Orchestrator Community Report Template; the goal is transparency. The
+template and its filing cadence are defined and iterate in the
+[governance repository](https://github.com/FilecoinFoundationWeb/Solstice-Governance/tree/main),
+maintained by [Filecoin Foundation's governance team](https://fil.org/governance).
+Every change to the report lands as a public pull request, on the
+public record. This applies to every Orchestrator, including the one
+seated at activation: there is no genesis exemption. Filing is
+binary; failure to file is a freeze trigger and forfeits
+re-admission for the following quarter. Report quality is surfaced
+through public scrutiny; enforcement runs through the Registry
+Safes' and Orchestrators' operating guidelines in the governance
+repository, at Filecoin Foundation's governance discretion.
+
+**Genesis Orchestrator disclosure.** The Orchestrator seated at
+activation is fixed by the activation migration (Section 2.4.10), not
+by `Admit`. To carry the same public scrutiny an `Admit` call would,
+the Registry Safes SHALL share, by activation, who operates that
+wallet, why it was selected, and what is expected of it — distinct
+from and in addition to that Orchestrator's own report above.
+
 **Service Orchestrators** are not considered a governance tier. An
 Orchestrator's only protocol interaction is receiving rewards and
 conducting measurement: it registers its (payer, operator) pairs
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -1936,7 +1936,7 @@ following quarter.
 
 **Initial Disclosure of Orchestrator Seated at Activation.** The Orchestrator
 seated at activation is fixed by the activation migration (Section 2.4.10), not
-by `Admit`. To carry the same public scrutiny an `Admit` call would, the
+by `AddOrchestrator(orch, wallet)`. To carry the same public scrutiny an `AddOrchestrator(orch, wallet)` call would, the
 Orchestrator shall publish a disclosure on activation, following the
 requirements in the governance repository set for any other Orchestrator.
 
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -1927,20 +1927,12 @@ Admission criteria, the freeze criteria and evidence standard, the
 escalation-to-Remove policy, the dispute procedure, the audit playbook,
 and the quarterly verification duty live in the governance repository.
 
-**Orchestrator Community Report.** Each Service Orchestrator SHALL
-share a quarterly Community Report with the community, using the
-Orchestrator Community Report Template; the goal is transparency. The
-template and its filing cadence are defined and iterate in the
-[governance repository](https://github.com/FilecoinFoundationWeb/Solstice-Governance/tree/main),
-maintained by [Filecoin Foundation's governance team](https://fil.org/governance).
-Every change to the report lands as a public pull request, on the
-public record. This applies to every Orchestrator, including the one
-seated at activation: there is no genesis exemption. Filing is
-binary; failure to file is a freeze trigger and forfeits
-re-admission for the following quarter. Report quality is surfaced
-through public scrutiny; enforcement runs through the Registry
-Safes' and Orchestrators' operating guidelines in the governance
-repository, at Filecoin Foundation's governance discretion.
+**Orchestrator Community Report.** Each Service Orchestrator shall share a
+quarterly Community Report using the Orchestrator Community Report Template.
+The template, required disclosures, and filing cadence are defined in the
+[governance repository](https://github.com/filecoin-project/solstice-governance/).
+Failure to file is a removal trigger and forfeits re-admission for the
+following quarter. 
 
 **Genesis Orchestrator disclosure.** The Orchestrator seated at
 activation is fixed by the activation migration (Section 2.4.10), not


FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
 escalation-to-Remove policy, the dispute procedure, the audit playbook,
 and the quarterly verification duty live in the governance repository.
 
+**Orchestrator Community Report.** Each Service Orchestrator SHALL
```
jsoares: Given the rest of the FIP does not use RFC 2119 terminology, I'd suggest not using it here either, and just reverting to lowercase `shall`.

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
...
+share a quarterly Community Report with the community, using the
+Orchestrator Community Report Template; the goal is transparency. The
+template and its filing cadence are defined and iterate in the
+[governance repository](https://github.com/FilecoinFoundationWeb/Solstice-Governance/tree/main),
+maintained by [Filecoin Foundation's governance team](https://fil.org/governance).
+Every change to the report lands as a public pull request, on the
+public record. This applies to every Orchestrator, including the one
+seated at activation: there is no genesis exemption. Filing is
+binary; failure to file is a freeze trigger and forfeits
```
jsoares: https://github.com/filecoin-project/FIPs/pull/1277 removes `freeze` in favour of removal as the only sanction, so this is now inconsistent.

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
...
+re-admission for the following quarter. Report quality is surfaced
+through public scrutiny; enforcement runs through the Registry
+Safes' and Orchestrators' operating guidelines in the governance
+repository, at Filecoin Foundation's governance discretion.
+
+**Genesis Orchestrator disclosure.** The Orchestrator seated at
+activation is fixed by the activation migration (Section 2.4.10), not
+by `Admit`. To carry the same public scrutiny an `Admit` call would,
+the Registry Safes SHALL share, by activation, who operates that
```
jsoares: Same comment re `shall`.

jsoares (approved):
Great to see more transparency, see minor comments on form.

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
 escalation-to-Remove policy, the dispute procedure, the audit playbook,
 and the quarterly verification duty live in the governance repository.
 
+**Orchestrator Community Report.** Each Service Orchestrator SHALL
+share a quarterly Community Report with the community, using the
+Orchestrator Community Report Template; the goal is transparency. The
+template and its filing cadence are defined and iterate in the
+[governance repository](https://github.com/FilecoinFoundationWeb/Solstice-Governance/tree/main),
```
decentramike: Note that this repository is intended to be moved to the main Filecoin repository, so I would not recommend hard coding this link.

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
...
+maintained by [Filecoin Foundation's governance team](https://fil.org/governance).
```
decentramike: In the interest of decentralization, I don't think we should hard code any one team, organization, or link in the FIP. I suggest, "maintained through the change control process in the governance repository."

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
...
+Every change to the report lands as a public pull request, on the
+public record. This applies to every Orchestrator, including the one
+seated at activation: there is no genesis exemption. Filing is
+binary; failure to file is a freeze trigger and forfeits
+re-admission for the following quarter. Report quality is surfaced
+through public scrutiny; enforcement runs through the Registry
+Safes' and Orchestrators' operating guidelines in the governance
+repository, at Filecoin Foundation's governance discretion.
+
+**Genesis Orchestrator disclosure.** The Orchestrator seated at
```
decentramike: I don't think any one Orchestrator should be enshrined as the "Genesis" Orchestrator, which implies some special powers or treatment. Suggest removing the word "Genesis".

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,28 @@ Admission criteria, the freeze criteria and evidence standard, the
...
+activation is fixed by the activation migration (Section 2.4.10), not
+by `Admit`. To carry the same public scrutiny an `Admit` call would,
+the Registry Safes SHALL share, by activation, who operates that
+wallet, why it was selected, and what is expected of it — distinct
```
decentramike: Setting separate rules for the first orchestrator undermines the principle that orchestrators operate on equal footing, and it cuts against the eventual goal of permissionless competition in Phase 2. The intent here is parity with an Admit, but Admit's disclosure content isn't fixed in the FIP; admission criteria and the evidence standard live in the governance repository. 

Suggest the following change: 
**Disclosure of the Orchestrator seated at activation.** The Orchestrator seated at activation is fixed by the activation migration (Section 2.4.10), not by `Admit`. To carry the same public scrutiny an `Admit` call would, the Orchestrator shall publish a disclosure by activation, following the requirements in the governance repository like any other Orchestrator.
```

**Review by decentramike — CHANGES_REQUESTED** [review]

File: FIPS/fip-0118.md (hunk 4 above)
Comment (dannyob) [inline FIPS/fip-0118.md]: I don't love that we have dependencies to github in FIPs, but we've done that before; we definitely should point to a filecoin-project repo (historically, FilecoinFoundationWeb is for the FF websites, and a bit of an anomaly). Can we link to and push to https://github.com/filecoin-project/solstice-governance/ which looks like it exists?

**Comment by dannyob** [comment]
I integrated @decentramike and @jsoares comments on the Community Report in the commit above, and tightened the language a little. I didn't integrate the comments on the Genesis Orchestrator, as I'm not sure I got @decentramike 's point there.

File: FIPS/fip-0118.md (hunk 6 above)
Comment (dannyob) [reply FIPS/fip-0118.md]: Isn't the idea here that when an orchestrator starts (or joins) it should announce its source and structure some how? (So future orchestrators would have the same obligation). If that's the case, I do think "Genesis" might be misleading, as it implies it only applies to the first orchestrator. Perhaps we should say "Initial Orchestrator disclosure", where "Initial" means at the start of orchestration, rather than the name of the orchestrator.

Comment (jennijuju) [reply FIPS/fip-0118.md]: The first orchestrator is hard coded to the protocol without SRA signers sending a message to add them — which is “special”  (and I was surprised by that on a recent call), that’s why I thought the disclosure is necessary.

> Isn't the idea here that when an orchestrator starts (or joins) it should announce its source and structure some how? 

yes — I believe the first orchestrator shall do the same too that’s why I explicitly added it! I like the rewording suggestion proposed by @dannyob .

File: FIPS/fip-0118.md (hunk 4 above)
Comment (jennijuju) [reply FIPS/fip-0118.md]: I am fine with that link  — I was provided with that link;  MPG from FIDL has also suggested to host them on solstice.filecoin.io or similar

File: FIPS/fip-0118.md (hunk 5 above)
Comment (jennijuju) [reply FIPS/fip-0118.md]: I agree with the principle, and I'd like to hold us to it in substance rather than in wording.

Two things make me hesitant to take the suggestion as written:

1. **"The change control process in the governance repository" doesn't exist in documented form today.** If the FIP defers to a process, that process has to be specifiable: how a community member proposes a change to the report template or a Solstice runbook, what the bar for acceptance is, and who merges. Right now the answer to the last one is the FF governance team, by repo permissions. I'm glad to point at a process instead of a team — as soon as there's a process to point at. 
  - To be clear, I’m not asking for process for its own sake. There are two workable answers here — name an accountable body, or build a real change control process that anyone can use. Naming a body is the cheap one imho. The one that doesn’t work is removing the name and pointing at neither.
2. **Removing the team name while keeping the link doesn't decentralize anything.** The repo sits under `FilecoinFoundationWeb`, an org controlled by a single organization. Dropping the name and keeping the pointer leaves the same concentration in place while making it harder for a reader to see who holds authority — which cuts against the transparency goal this section exists to serve. We can point at a different repo under filecoin-project instead — but that only moves the question, since CODEOWNERS for reviewing and merging the runbooks and template still has to be defined either way.

So I'd propose the version of your suggestion that actually gets us there: movethe Solstice governance artifacts under `filecoin-project`, where the rest of the governance process lives, with CODEOWNERS covering the runbooks and template, and document the change control process there. With those in place I'll drop both the team reference and the org-specific link and use exactly the language you suggested.

I think that's the outcome we both want; I'd just rather the FIP not assert a neutrality the process doesn't have yet.

File: FIPS/fip-0118.md (hunk 4 above)
Comment (jennijuju) [reply FIPS/fip-0118.md]: We need a link before fip’s acceptance.

Comment (decentramike) [reply FIPS/fip-0118.md]: https://github.com/filecoin-project/solstice-governance/

File: FIPS/fip-0118.md [hunk 8]
```diff
@@ -1927,6 +1927,20 @@ Admission criteria, the freeze criteria and evidence standard, the
...
 and the quarterly verification duty live in the governance repository.
 
+**Orchestrator Community Report.** Each Service Orchestrator shall share a
+quarterly Community Report using the Orchestrator Community Report Template.
+The template, required disclosures, and filing cadence are defined in the
+[governance repository](https://github.com/filecoin-project/solstice-governance/).
+Failure to file is a removal trigger and forfeits re-admission for the
+following quarter. 
+
+**Genesis Orchestrator disclosure.** The Orchestrator seated at
```
decentramike: '''suggestion
**Initial Orchestrator disclosure.** The Orchestrator seated at
'''

dannyob:
Incorporated @decentramike's language to establish that the initial orchestrator has the same disclosure requirements as any other orchestrator, just at activation, not at `Admit`.

decentramike (approved):

decentramike (approved):

FIPS/fip-0118.md
```diff
@@ -1927,6 +1927,19 @@ Admission criteria, the freeze criteria and evidence standard, the
...
+**Orchestrator Community Report.** Each Service Orchestrator shall share a
+quarterly Community Report using the Orchestrator Community Report Template.
+The template, required disclosures, and filing cadence are defined in the
+[governance repository](https://github.com/filecoin-project/solstice-governance/).
+Failure to file is a removal trigger and forfeits re-admission for the
+following quarter. 
+
+**Initial Disclosure of Orchestrator Seated at Activation.** The Orchestrator
+seated at activation is fixed by the activation migration (Section 2.4.10), not
+by `Admit`. To carry the same public scrutiny an `Admit` call would, the
```
irenegia: The new paragraph mentions `Admit` twice, but #1277 renames the registry methods: `Admit(orch)` is now `AddOrchestrator(orch, wallet)`, `Remove(orch)` is `RemoveOrchestrator(orch, reason)`, and freeze is gone. Same root as the freeze/removal point @jsoares raised. Once both PRs land, `Admit` no longer exists in the FIP text. Suggest `AddOrchestrator` here, or wording it without the method name.

cc @decentramike

FIPS/fip-0118.md
jennijuju: please adds "who is the owner/maintainer" of https://github.com/filecoin-project/solstice-governance/ either in repo README, or I will specify here

FIPS/fip-0118.md
irenegia:
```suggestion
by `AddOrchestrator(orch, wallet)`. To carry the same public scrutiny an `AddOrchestrator(orch, wallet)` call would, the
```

decentramike (approved):

PR #1286, opened 2026-09-04
FIP-0118: spec clarifications

irenegia:
Follow-up edits to FIP-0118 after #1277. Two groups.

**Text aligned with the implementation**
- Section 2.4 (Rod, #1287 and 7f80715): reward-actor mechanics, the `ReplaceAddress` f02 method, payment channels rejected as recipients, f02 state and accounting alignment.
- Section 3.1: the SWA method list matches the contract (four named discretionary relays, `CancelPending` / `CancelPendingWeight`); the state bullet becomes a table; `QuarterlyGateCheck` emits a check-result event.
- Section 3.1.1: `QuarterlyGateCheck` reverts during the `SetWeightRecords` and `SetGateParams` holds (filecoin-project/solstice#67); consecutive late passes land one hold apart.
- Section 3.2: `ReplaceWallet(old, new)` drops `extradata`; `PostVolume` takes the quarter argument; `CancelBinding` is called by the bound Orchestrator; the one-quarter admission lag is dropped.
- Section 2.2: quarter Q is `[Start(Q), Start(Q+1))` with half-open windows; posting period plus verification window must be shorter than a quarter; same-epoch ordering of registry writes and settlements; recap table of the application rules.
- Section 1.2: only active sectors are eligible for `UpgradeSectorQuality`. Section 1.5: the RKH signer-removal migration bullet is removed.
- `EPOCHS_PER_QUARTER` is 262,974 epochs (floor(EPOCHS_IN_YEAR / 4), 91.31 days).

**Clarifications, no behavior change**
- Section 2.2: payment period vs measured quarter (Figure 5).
- Section 3.1.1: a w2 record change is paired with `SetGateParams` as a governance sequencing rule; the gate-write rejection cases name a new stream.
- The weight-sum guardrail is spelled `Σ_{i>=1} w_i <= 1` throughout.
- Security Considerations: sole-Orchestrator removal, `CorrectVolume` as the injection path, the single-multisig cases, consistency fixes. Section 4.2: which holds f02 enforces.
- Implementation links.

PR #1287, opened 2026-09-04
FIP-0118: tweak reward actor mechanics based on impl & changes

rvagg:
Ref: https://github.com/filecoin-project/builtin-actors/pull/1785

On top of #1286 @irenegia

### DIFF
--- a/FIPS/fip-0118.md
+++ b/FIPS/fip-0118.md
@@ -1265,25 +1265,22 @@ AwardBlockReward(miner, penalty, gas_reward, win_count):
     load the streams block            // absent block or store error aborts
     computed_BR = this_epoch_reward * win_count / 5
     if the block is undecodable, or its stream, tombstone or queue
-       structure is invalid:
+       structure or its accounting is invalid:
         no_award()
-    if current explicit liability is uncomputable:
-        award_without_service()
-    projection = project valid due writes and cancellation-stranded drops
-    fold_dust = projection's fold dust
+    apply the due writes, dropping the cancellation-stranded ones
+    fold_dust = the dust those folds left
     liability =
-        Σ projected live (accrued - Σ claimed_period + Σ payable)
-        + Σ projected tombstone payable
-    if liability is uncomputable:
-        award_without_service()
+        Σ live (accrued - Σ claimed_period + Σ payable)
+        + Σ tombstone payable
     if balance <= gas_reward + liability + fold_dust:
         no_award()
     BR = min(computed_BR, balance - gas_reward - liability - fold_dust)
+    if BR < 0:                        // only a negative this_epoch_reward
+        no_award()
     evaluated = ComputeWeight for every active stream
     if any record violates 0 <= floor <= v_start <= cap <= DENOM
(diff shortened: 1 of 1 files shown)


TippyFlitsUK (approved):

FIPS/fip-0118.md
irenegia:
```suggestion
Once the award commits the due folds, their fold dust is no longer owed to recipients but has not yet been burned, so it is excluded from BR and sent to f099 with the residual. Burning the exact residual conserves `BR`
```

irenegia: The PR deletes the award projection everywhere else in section 2.4.3 but missed this sentence

FIPS/fip-0118.md
```diff
@@ -1265,25 +1265,22 @@ AwardBlockReward(miner, penalty, gas_reward, win_count):
     load the streams block            // absent block or store error aborts
     computed_BR = this_epoch_reward * win_count / 5
     if the block is undecodable, or its stream, tombstone or queue
-       structure is invalid:
+       structure or its accounting is invalid:
```
irenegia, line 1268: Nitpicking: "accounting is invalid" is not defined anywhere in the FIP. The implementation defines it in one sentence: "one accrual row per live explicit stream and no others, accruals non-negative, recipient rows positive, and no wallet claimed more than it has earned." Maybe lift that into 2.4.3 or 2.4.8?

FIPS/fip-0118.md
```diff
@@ -1574,10 +1544,10 @@ repaired in band at all, because every explicit method aborts at load
 before it can propose anything; only a network upgrade carrying a
 migration recovers it.
 
-Inconsistent accounting aborts `Claim`, `SetShares`, and governance
-mutations; `AwardBlockReward` instead degrades as Section 2.4.3
-specifies. These are fail-closed responses to inconsistent or corrupted
-state, not alternative allocation modes.
+Inconsistent accounting aborts every explicit method; `AwardBlockReward`
```
irenegia: same comments as before about "inconsistent accounting"

FIPS/fip-0118.md
irenegia: This PR removes the degraded award mode, so broken accounting now
stops all block rewards until a network upgrade ships (2.4.8: it
aborts every explicit method, so no in-band write can repair it).
This bullet covers only the weight-record case; adding the
accounting case:

```suggestion
    allocation. An otherwise well-formed persisted violation mints no
    reward at all and permits a validating timelocked repair, so no
    party gains by inducing one. Inconsistent accounting is stricter:
    it aborts every explicit method, so no in-band write can repair it;
    awards pay only `gas_reward`, with penalties as today, and
    consensus mining runs on gas rewards alone until a network upgrade
    carrying a migration restores the state. Otherwise each allocation
    is floored and the exact residual burns, conserving `BR`
    atto-exactly.
```

irenegia (changes requested):

BigLep:
Per 2026-09-08 engineering sync: Rod agreed to the comments. @irenegia  will apply the suggestions and then merge the PR.

PR #1289, opened 2026-09-11
Proposal to Add Hannah Howard as a FIP Editor

ThatGuyLLC:
Given recent changes in participation and activity among the current FIP editors, we are reviewing the editor group to ensure there is sufficient active capacity to support the FIP process and the broader community. 

As part of that review, we would like to propose Hannah Howard, Director of Engineering at the Filecoin Foundation, to serve as a FIP Editor. 

Hannah brings strong engineering experience, familiarity with the Filecoin ecosystem, and the technical judgment needed to help support authors as proposals move through the FIP lifecycle. Adding another active editor would also help distribute the responsibilities currently carried by the existing editor group, improve responsiveness to authors, and provide additional capacity as FIP activity continues to grow. 

The intent is not to materially change the role or responsibilities of FIP editors, but rather to strengthen the existing group and ensure the process remains well-supported, timely, and sustainable. 

We would like to open this proposal for community discussion and feedback regarding Hannah’s addition to the FIP editor group.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-*       @jennijuju @anorth @jsoares @TippyFlitsUK @momack2 @rvagg @lanzafame @dannyob
+*       @jennijuju @anorth @jsoares @TippyFlitsUK @momack2 @rvagg @lanzafame @dannyob @hannahhoward


jsoares (approved):
Very much in favour. 

Note that the list of editors is duplicated in https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0001.md.

I was under the impression Alex had stepped down, so might be worth updating that too.

rvagg (approved):

